PR #772, opened 2014-07-14
Add exceptional websocket commands

tolysz:
Just added some other commands from the websocket library.
I am not sure however if this is the correct way of handling exceptions. For my things I needed that there will be some IO action executed, which basically would shut down some other process running.


tolysz:
nice :)

snoyberg:
Sorry for the delay on it, it slipped under the radar. I'm working on cleaning up the issue tracker right now to avoid that from happening again.

tolysz:
I have two extra functions let me add them in a separate PR

PR #801, opened 2014-08-07
Proposal: Switch from using ENTER to CTRL-D for exiting

burz:
Using CTRL-D as an exit strategy as opposed to ENTER is simple to implement. Furthermore, it allows users to enter new lines into the console to separate request blocks for easier debugging.


snoyberg:
It looks like this was already resolved by the -c option. I've also just modified things with 46fbb1a to require explicitly typing in `quit`.

Issue #767, opened 2014-07-07
Yesod eats 10-30% CPU when running idle in development


darthdeus:
I've asked around on IRC and it seems I'm not the only one experiencing this. I just noticed today when I left my Yesod app running and left the room for about 5 minutes, only to come back and hear my cpu fan running, so I opened htop and there was yesod, eating half of my CPU :)

I've been also using [DevelMain.hs](https://github.com/chrisdone/ghci-reload-demo) from @chrisdone, which seems to eat even more CPU and hangs at around 25-50%, while `yesod devel` itself runs only at about 10-30%. I'm on OS X 10.9 and GHC 7.8.2 if that's of any relevance.

Is there some way that I can inspect what is using the CPU?

jezen:
This issue is now ancient, and the PR that may have been an approach to addressing this was closed half a decade ago, unmerged. I think we can close this.

PR #815, opened 2014-08-31
fix of issue #767 (great CPU usage in devel)

dunric:
Hi,

I've played a while with an annoying issue #767 when yesod binary run in a development mode consumes inappropriate amount of CPU cycles.
There is a recent partial fix which increases polling time interval of file list check (https://github.com/paul-rouse/yesod/commit/1d3b60abefb9833e9f3d3a382ae3404313638652) however find is not quite satisfactory.

Suggested fix removes polling interval entirely as you can completely rely on `filesModified` semaphore controlled by `System.FSNotify.watchTree` . Unless there is no alteration in source tree, there is no point to perform detailed check of changes.
After the fix applied CPU usage becomes minimal when idle (drops from 15% to 1% on my computer).

I can say the fix does work for me reliably after a several days of testing without any quirks (on Linux machine). Would welcome if users of other systems can confirm whether it also works for them.

### DIFF
--- a/yesod-bin/Devel.hs
+++ b/yesod-bin/Devel.hs
@@ -35,6 +35,9 @@ import qualified Data.List                             as L
 import qualified Data.Map                              as Map
 import           Data.Maybe                            (fromMaybe)
 import qualified Data.Set                              as Set
+import qualified Data.Text                             as T
+
+import qualified Filesystem.Path.CurrentOS             as FP (toText)
 
 import           System.Directory
 import           System.Environment                    (getEnvironment)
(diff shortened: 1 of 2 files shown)


snoyberg:
I'm not sure what the current status of this is, or whether it should be merged or closed. @gregwebs @paul-rouse @dunric any input?

paul-rouse:
As it stands I think it should just be closed: the real solution is, I believe, something more radical involving `determineDeps`.  When polling, this patch has worse asymptotic behaviour and may be broken in extreme cases.

snoyberg:
Thanks for the feedback, closing

PR #904, opened 2015-01-07
Updated french translation of Yesod.Auth.Message.

legrostdg:
partly revert c9dff10bd19bc83a42a22b13306c5e2dcbd25b6f (in particular courriel != adresse électronique)


snoyberg:
This puts me in a difficult position as maintainer right now... I have no knowledge of French at all, and I essentially have two people telling me two completely conflicting things about the language. I'm not sure how to proceed. Is there anyone out there to break the tie?

gregwebs:
Ultimately we may just need to have alternative translations

snoyberg:
Not that I generally trust Yahoo! Answers, but this implies that we *should* imply the current PR:

https://uk.answers.yahoo.com/question/index?qid=20080117064253AAwsLo8

tl;dr: courriel is local dialect/more informal. If I don't hear an objection, I'll merge this in the next few days.

paul-rouse:
@snoyberg  My wife is a professional translator - she is going to check these with some of her contacts who are French native speakers.  I'll let you know...

snoyberg:
Hah, that's awesome, thank you!

legrostdg:
native speaker here :-).

courriel =~ courrier électronique <=> email
mél =~ adresse électronique <=> email address

"courriel" is ok, but "mél" is not very common.

We can keep "courriel" were it applies, but not as a replacement for "email address".

paul-rouse:
The replies my wife received all basically agree with @legrostdg , so I suggest accepting this PR.

One response did acknowledge that "courriel" is sometimes used informally to mean the address, but still felt that "adresse électronique" is better in this context, as it is more precise and acceptable to French speakers everywhere.  All were happy with the one remaining use of "courriel" to mean the *message*, although one pointed out that it is not common in the francophone parts of Africa.

PR #909, opened 2015-01-15
Upper bound on hspec for known breakage.

dlthomas:
yesod-test 1.2.6 uses a type that doesn't exist in hspec >=2.0


snoyberg:
Can you provide the error message you got? I've used this combination many times.

dlthomas:
Huh, it seems to have built just fine that time.  Maybe I hadn't cleared something out... :-\  Anyway, will investigate further if it comes up again.  For now go ahead and bin this one.

PR #911, opened 2015-01-18
Fix pluralization in template loading error message

MaxGabriel:
I got this error message using `widgetFile` and noticed the pluralization was incorrect:

```
Handler/PostDetails.hs:8:21:
    Exception when trying to run compile-time code:
      Called widgetFileReload on "postDetails/post.hamlet", but no template were found.
    Code: widgetFile "postDetails/post.hamlet"
    In the splice: $(widgetFile "postDetails/post.hamlet")
```

It should be "but no templates were found".

### DIFF
--- a/yesod/Yesod/Default/Util.hs
+++ b/yesod/Yesod/Default/Util.hs
@@ -102,7 +102,7 @@ combine func file isReload tls = do
             , func
             , " on "
             , show file
-            , ", but no template were found."
+            , ", but no templates were found."
             ]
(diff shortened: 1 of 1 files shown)


paul-rouse:
Wouldn't "no template was found" be better, since it is only looking for one of them?  (Call me a pedant?)

MaxGabriel:
I think it's looking for Hamlet/Julius/Lucius/Cassius in this case.

The wording is a little awkward, admittedly.

gregwebs:
thanks!

PR #915, opened 2015-01-25
Demonstrate failing CSRF token lookup (don't merge)

MaxGabriel:
This isn't quite a PR yet, this just demonstrates a bug where `addNonce_` fails to lookup a form by ID.

I believe this is because [this code](https://github.com/yesodweb/yesod/blob/2fb7cccf15c04cef78158f792cf1a569bd87ceec/yesod-test/Yesod/Test.hs#L493) concatenates the selector with `"input[name=_token][type=hidden][value]"`, causing the CSS query to be `"#formIDinput[name=_token][type=hidden][value]"`. If I add a space after `"#formID"` (`"#formID "`), then the query works correctly.

``` haskell
matches <- htmlQuery' rbdResponse $ scope <> "input[name=_token][type=hidden][value]"
```

I initially guessed I could easily fix it by adding a space before `input`, but that lead to a 500 error on another test, so I'm still investigating.

### DIFF
--- a/yesod-test/test/main.hs
+++ b/yesod-test/test/main.hs
@@ -156,6 +156,13 @@ main = hspec $ do
                     setMethod "POST"
                     setUrl ("/labels" :: Text)
                     byLabel "Foo Bar" "yes"
+
+        ydescribe "tokens" $ do
+            yit "can find a CSRF token field" $ do
+                get ("/tokens" :: Text)
+                request $ do
+                    addNonce_ "#formId"
+
     describe "cookies" $ yesodSpec cookieApp $ do
         yit "should send the cookie #730" $ do
             get ("/" :: Text)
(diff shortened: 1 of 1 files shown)


MaxGabriel:
Ah, so I think a CSS query starting with a space returns `Right [[]]` as the `SelectorGroup`, which matches the whole document. Then `Yesod.Test` gets the "value" back from that HTML, gets the empty list, and calls `head` on it causing the 500.

```
*Yesod.Test> parseQuery "#foo"
Right [[DeepChildren [ById "foo"]]]
*Yesod.Test> parseQuery " #foo"
Right [[]]
*Yesod.Test> parseQuery " input[name=_token][type=hidden][value]"
Right [[]]
*Yesod.Test> parseQuery "input[name=_token][type=hidden][value]"
Right [[DeepChildren [ByTagName "input",ByAttrEquals "name" "_token",ByAttrEquals "type" "hidden",ByAttrExists "value"]]]
```

Investigating if that's an issue with the TransversingCSS module or the Yesod.Test code should just not send queries starting with a space to TransversingCSS.

snoyberg:
Is this safe to close?

MaxGabriel:
@snoyberg I confirmed the bug is still present. I can look into fixing it.

MaxGabriel:
Looking back, my comments on this PR are a little unintelligible; the [files changed](https://github.com/yesodweb/yesod/pull/915/files) gives a better idea of what's going on.

StevenXL:
This seems to have been address in #1209.

Issue #914, opened 2015-01-25
yesod-test functions/documentation still use the term "nonce"


MaxGabriel:
After discussion in #214, yesod-form started using the term "token" instead of "nonce" (implemented [in this commit](https://github.com/yesodweb/yesod/commit/5762272f8b6ab49e37e9cae73378047d35f563ea)). 

However, yesod-test still has [functions](https://github.com/yesodweb/yesod/blob/2fb7cccf15c04cef78158f792cf1a569bd87ceec/yesod-test/Yesod/Test.hs#L489-L501) and [documentation](https://github.com/yesodweb/yesod/blob/2fb7cccf15c04cef78158f792cf1a569bd87ceec/yesod-test/Yesod/Test.hs#L21) using that word. 

Proposal: 
* Deprecate `addNonce` and `addNonce_`; add new functions called something like `addToken` or `addCsrfToken`. Implemented in #917
* Change documentation/comments in yesod-test to use the term "token". Implemented in #917
* Update the scaffolding to use the new function names (they [currently use the `addNonce` function](https://github.com/yesodweb/yesod-scaffold/blob/e4b2fb01eec21c60ed07e3e2a84a0f823341d742/test/Handler/HomeSpec.hs)).
* Make a PR to remove the deprecated functions in a future major release of Yesod.

Full search for "nonce" in this repo: https://github.com/yesodweb/yesod/search?utf8=%E2%9C%93&q=nonce

gregwebs:
sounds good to me. The scaffolding will need to require the newer version of yesod-test.

PR #917, opened 2015-01-26
Move yesod-test to using the term "token" instead of nonce.

MaxGabriel:
I also considered `addCsrfToken` but I found it super hard to type, and it seems like the rest of Yesod just uses token as well (with the exception of Yesod.Auth.GoogleEmail2).

Addresses part 1 of #914

Edit: would you like me to update the changelog when I make a PR? I know some projects just go through commits when they make a release and others put new changes into a "master" section of the changelog that later gets moved to a specific version number.


snoyberg:
@MaxGabriel Would you be able to rebase these changes to be compatible with the other PR?

MaxGabriel:
@snoyberg Yes I would. I'll also update the documentation based on these changes

MaxGabriel:
@snoyberg Ok, I've rebased against master and updated the documentation to talk about tokens.

snoyberg:
Beautiful, thank you!

PR #922, opened 2015-02-04
Fix add-handler putting two routes on one line

### LINKED ISSUE #921: Add handler should check the line it is writing to is empty
With a routes file with no trailing newline at the end of the file, running add-handler appends the new route to the old one, resulting in invalid code.  Should be simple to insert the newline if required.

MaxGabriel:
* If there is a new line at the end of the file, add the route as normal
* If there isn't, add a newline character before the route
* Closes #921

I tested this on OS X. I'm not super familiar with what type of newlines Windows programmers use; I'm not sure if that's an issue here.


snoyberg:
Thanks!

PR #923, opened 2015-02-04
Perform `onLogout` before session cleaning

s9gf4ult:
There is setuations where this hook must get some data from session, so it must be performed before session cleaning.


gregwebs:
Thanks, this looks like a good fix. I will wait for Travis to build.

s9gf4ult:
Wow so fast.

s9gf4ult:
Would be cool to see it on hackage.

Issue #920, opened 2015-01-29
yesod-test README code doesn't compile


MaxGabriel:
The [README example](https://github.com/yesodweb/yesod/blob/5d431fbdb542084d1a77f9c4df786d32cfe79721/yesod-test/README.md) in yesod-test, afaict, doesn't typecheck.

Based on testing it, the `post url $ do` style that's used several times doesn't work, perhaps it's meant to be `request $ do`?

```
post "url/to/post/to" $ do
    addToken -- Add the CSRF _token field with the currently shown value
```

Additionally, no `byName` function exists, that's probably meant to be `addPostParam`:

```
-- And of course you can add any field if you know it's name
byName "answer" "42"
```

snoyberg:
Looks like that would be an improvement. Would you be able to send a PR?

MaxGabriel:
Yeah I can work on that

PR #927, opened 2015-02-08
Update the example in the yesod-test README

MaxGabriel:
This is slightly larger of an update described in #920.

This PR fixes usage of `request` and `addPostParam`.

I felt like the `main` function in the example was out of date, too. Initially I tried to replace it to get a runnable example, but I realized that noone would be able to run the example code anyway without a corresponding Yesod app. So I just went with a spec implemented like you'd see in the yesod-scaffolding + pointed to the Yesod scaffolding as an example using a real app. 

This also fixes a handful of typos/formatting issues in the top section of the markdown.


snoyberg:
Thanks!

PR #929, opened 2015-02-09
Add more semigroups instances

MaxGabriel:
@gregwebs I thought there might as well be Semigroup instances for the rest of the monoidal types as well. Does this look ok?


gregwebs:
I am not sure to what extent these others end up getting used by end users, but it makes sense to add instances to anything that is publicly exposed.

PR #936, opened 2015-02-16
Added means to fetch user's Google profile

zudov:
The existing GoogleEmail2 auth did that:
  - Acquire user access token with offline access
  - Use token to acquire user's profile
  - Build `Creds` with user's email as `ident` and the other profile
    details as stringy key-value pairs in `credsExtra`

This wasn't enough for me, for several reasons:
  - Access token was not saved after authentication. If we request 'offline'
    token why not to have a way of using it later.
  - Stringy key-value profile is not nice and `credsExtra` can be accessed
    only from `getAuthId`
  - I might want to request the profile after authentication process

So I've added the needed features.
  - The access token is saved in a session
  - There is a `Person` type with `FromJSON` instance and `getPerson`
    can be used to acquire it from `HandlerT`

The changes are not breaking.

### DIFF
--- a/yesod-auth/Yesod/Auth/GoogleEmail2.hs
+++ b/yesod-auth/Yesod/Auth/GoogleEmail2.hs
@@ -22,8 +22,9 @@
 --
 -- Since 1.3.1
 module Yesod.Auth.GoogleEmail2
-    ( -- * Authentication handler
+    ( -- * Authentication handlers
       authGoogleEmail
+    , authGoogleEmailSaveToken
     , forwardUrl
     -- * User authentication token
     , Token(..)
@@ -48,7 +49,7 @@ module Yesod.Auth.GoogleEmail2
 import           Blaze.ByteString.Builder (fromByteString, toByteString)
 import           Control.Applicative      ((<$>), (<*>))
 import           Control.Arrow            (second)
-import           Control.Monad            (liftM, unless)
+import           Control.Monad            (liftM, unless, when)
 import           Data.Aeson               ((.:?))
 import qualified Data.Aeson               as A
 import qualified Data.Aeson.Encode        as A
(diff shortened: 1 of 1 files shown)


snoyberg:
This looks great, and a quick review looks like the code is solid. I'll need a bit more time to dive into this deeply.

paul-rouse:
Would it be worth having an option to prevent adding the access token to the session?  I prefer to avoid extra baggage in the session when I don't need it.  Alternatively, but more messy for the caller, perhaps export something to allow the caller to delete the extra session variable: either export ```accessTokenKey``` or provide a function to do the deletion.

zudov:
@paul-rouse That's definetely a good idea. Features should be optional.

Rephrasing your message there are two ways:

1. Add `Bool` parameter to `authGoogleEmail` which would change the API, though the users' code would need very little change.
2. Add `removeToken` function and a note in documentation.

No doubt, the first way is much cleaner and more convinent for the user. But does it worth to break the API? @snoyberg What do you think?

snoyberg:
Still haven't had a chance to dig into this, but gur reaction is to avoid
the API breakage.

zudov:
I've just figured the might-be-best solution.

My current code changes the behaviour of `authGoogleEmail` auth-plugin. Instead I could leave it as it was before and add another auth-plugin, `authGoogleEmailSaveToken` (need a better name) which would be saving token, allowing to query the profile.

paul-rouse:
I was about to say something very similar!

zudov:
I would imlement it now. Any ideas on better name?

paul-rouse:
I am struggling to do better.  Maybe it could describe what it gives access to, instead of just the idea of saving the token, but I can't think of a nice, brief way of saying it in a function name!

zudov:
I've added the changes so that @snoyberg can review it. The function's name is under objection.

snoyberg:
This looks like solid code, very impressive and well written. Thank you! I'm OK with that naming for the function; if no one has an objection in the next day, I'll merge this in.

snoyberg:
OK, no objections, so I've merged and released to Hackage as version 1.4.3. Thanks again!

zudov:
Great

Issue #476, opened 2013-01-08
Yesod devel complains "Yesod has been compiled with a different GHC version"


MasseR:
I installed yesod within virthualenv. Yesod scaffolding commands basically work, but when trying to run devel, I get the following:

```
Yesod devel server. Press ENTER to quit
Configuring introitu-0.0.2.0...
ERROR: Yesod has been compiled with a different GHC version, please reinstall
```

Yesod has just been reinstalled and I only have one GHC version installed on my laptop.

JulianBirch:
Hi, I just got bitten by this after changing from 7.8.2 to 7.8.4. The message could really do with saying something to the effect of "runghc reported 7.8.2 but devel.hs is expecting 7.8.4".

simonmichael:
Me too. In my case ghc and runghc were both reporting 7.8.4. I needed to cabal install yesod-bin again to update the yesod executable. yesod-bin is easy to overlook, since it isn't listed by ghc-pkg. Ideally yesod would specifically recommend reinstalling it in this case.

snoyberg:
That's a good idea, anyone fell like sending a PR to update the text?

PR #940, opened 2015-02-22
Improve error message when "yesod devel" complains about the GHC version...

haroldl:
....

I hit this problem and the fix was to reinstall yesod-bin (because I'd
upgraded my version of ghc):

    cabal install --reinstall yesod-bin

See issue #476 .


creichert:
Thanks! I'm not sure what's wrong with the build but it seems unrelated.

snoyberg:
Thanks!

PR #945, opened 2015-02-28
Disregard proxy environment variables in yesod devel

paul-rouse:
After the change to proxy settings in http-client (https://github.com/snoyberg/http-client/issues/94) the ```yesod devel``` reverse proxy now attempts to connect to the application via any further proxy specified by the environment variables.  I can't see any reason why the new behaviour might be useful in this specific case, so it seems better to me to force no proxy - am I missing something?

### DIFF
--- a/yesod-bin/Devel.hs
+++ b/yesod-bin/Devel.hs
@@ -68,7 +68,9 @@ import           Data.Streaming.Network                (bindPortTCP)
 import           Network                               (withSocketsDo)
 import           Network.HTTP.Conduit                  (conduitManagerSettings, newManager)
 import           Data.Default.Class                    (def)
+#if MIN_VERSION_http_client(0,4,7)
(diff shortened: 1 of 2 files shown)


snoyberg:
Looks good, but we need a lower bound on http-client for that new function.

paul-rouse:
Good point (blush) - lower bound now added.  Following on, though, do you think it is OK to break compatibility with earlier http-client versions?

snoyberg:
It's a "nice to have." Probably all that would be necessary is a quick MIN_VERSION_http_client(0,4,7) that simply doesn't alter this setting for older versions. I'm fine with or without that compatibility. I'll hold off on merging till I hear whether you want to add the compatibility back or not.

paul-rouse:
Compatibility done with CPP conditionals.  I'll try to remember this needs tidying up after a reasonable period...

snoyberg:
Perfect!

PR #947, opened 2015-03-04
Add and process Keter option 'extraFiles'

chreekat:
Submitting for review and feedback. If this seems useful, please point me to any docs that would need updating, and I'll add them to the PR.

This allows keter.yml files like such:

    exec: ../dist/build/Snowdrift/Snowdrift
    args:
        - production
    host: localhost
    extraFiles:
        - ../migrations

Like exec, the location is relative to the configuration file itself.

### DIFF
--- a/yesod-bin/Keter.hs
+++ b/yesod-bin/Keter.hs
@@ -9,7 +9,7 @@ import qualified Data.Text as T
 import System.Exit
 import System.Process
 import Control.Monad
-import System.Directory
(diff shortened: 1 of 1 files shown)


chreekat:
Hm, I saw a potential refactor, and then ghc started sprouting warnings that I don't remember seeing. Let me work on those first.

chreekat:
Okidokey fixed.

snoyberg:
Thanks!

PR #958, opened 2015-03-20
loosen upper bound on monad-control in generated scaffolding

tdimiduk:
The scaffolding appears to create a cabal file with an overly restrictive upper bound on monad-control. The current scaffolding limits monad-control to < 0.4, causing my build to fail when I have monad-control-1.0. However, if I lift the upper bound to say < 1.1, it compiles fine. Is this a fine change to make? Is this the correct way to do this fix?


paul-rouse:
Actually a change to do this is already merged into the scaffold (https://github.com/yesodweb/yesod-scaffold/pull/68), but I don't think it has got into a release yet.

tdimiduk:
Ok, cool. As long as it is getting fixed I am happy :).

PR #959, opened 2015-03-20
Add AuthenticationResult and authenticate function

### LINKED ISSUE #956: getAuthId to return a more useful result
When I'm unable to authenticate a user based on some `Creds`, the only information I really have is that `getAuthId` returned `Nothing` and the page shows "Invalid login". I frequently have to debug `getAuthId`, when a more informative return value could have told me directly what was going on.

The simplest thing would be to upgrade the `Maybe` to `Either`, but then we'd have to decide if that `Left` message is intended for a developer or the end-user. We could return something like:

```haskell
data AuthResult
    = Success UserId
    | UserError Message -- User's fault, show them message
    | ServerError Message -- Server's fault, log message, show user "Sorry" or something
```

Obviously this is a breaking change, but I wanted to put it out there and start the discussion. I could handle my own use case in a few other ways (exceptions, logging, etc), but this felt like a Good Idea in general.

pbrisbin:
- getAuthId returns Maybe AuthId with no useful information in the
 Nothing case.
- AuthenticationResult includes whether it was a User or Server error
 (with an accompanying message) in the failure case.
- User errors are displayed back to the user and have a 401 status in
 JSON responses. Server errors are logged and a generic error message
 is presented to the user, with a 500 status in JSON responses.

Resolves #956

I really wanted to call these `AuthResult` and `getAuthResult` but `Yesod.Core`
has already taken `AuthResult` for "Auth**orization** result". I'm very much
open to naming ideas because I'm not happy with what I have, but can't think of
anything better.

I'm also nervous that none of this is tested. Is there a higher-level test that
hits these paths? Should I try to start a test suite here in yesod-auth? I
think it'd be valuable, but I don't know how much effort it would be, i.e. if
it's worth it.

### DIFF
--- a/yesod-auth/Yesod/Auth.hs
+++ b/yesod-auth/Yesod/Auth.hs
@@ -121,6 +121,7 @@ class (Yesod master, PathPiece (AuthId master), RenderMessage master FormMessage
     --
     -- Default implementation is in terms of @'getAuthId'@
     --
+    -- Since: 1.4.4
     authenticate :: Creds master -> HandlerT master IO (AuthenticationResult master)
     authenticate creds = do
         muid <- getAuthId creds
--- a/yesod-auth/Yesod/Auth.hs
+++ b/yesod-auth/Yesod/Auth.hs
@@ -28,6 +28,7 @@ module Yesod.Auth
     , loginErrorMessage
     , loginErrorMessageI
       -- * User functions
+    , AuthenticationResult (..)
     , defaultMaybeAuthId
     , maybeAuthPair
     , maybeAuth
@@ -67,7 +68,7 @@ import qualified Yesod.Auth.Message as Msg
 import Yesod.Form (FormMessage)
 import Data.Typeable (Typeable)
 import Control.Exception (Exception)
-import Network.HTTP.Types          (unauthorized401)
+import Network.HTTP.Types (Status, internalServerError500, unauthorized401)
 import Control.Monad.Trans.Resource (MonadResourceBase)
 import qualified Control.Monad.Trans.Writer    as Writer
 import Control.Monad (void)
@@ -78,6 +79,12 @@ type AuthHandler master a = YesodAuth master => HandlerT Auth (HandlerT master I
 
 type Method = Text
 type Piece = Text
+
+-- | The result of an authentication based on credentials
+data AuthenticationResult master
+    = Authenticated (AuthId master) -- ^ Authenticated successfully
+    | UserError AuthMessage         -- ^ Invalid credentials provided by user
+    | ServerError Text              -- ^ Some other error
 
 data AuthPlugin master = AuthPlugin
     { apName :: Text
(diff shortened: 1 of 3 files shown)


yesod-auth/Yesod/Auth.hs
```diff
@@ -110,8 +117,27 @@ class (Yesod master, PathPiece (AuthId master), RenderMessage master FormMessage
     -- destination exists.
     logoutDest :: master -> Route master
 
+    -- | Perform authentication based on the given credentials.
+    --
+    -- Default implementation is in terms of @getAuthId@
```
MaxGabriel: Slightly preferable to have single quotes around `getAuthId` so that haddock generates a hyperlink to the function name.

yesod-auth/Yesod/Auth/Message.hs
```diff
@@ -592,6 +604,7 @@ russianMessage PasswordResetPrompt = "Введите адрес эл.почты
 russianMessage InvalidUsernamePass = "Неверное сочетание имени пользователя и пароля"
 russianMessage (IdentifierNotFound ident) = "Логин не найден: " `mappend` ident
 russianMessage Logout = "Выйти"
+russianMessage AuthError = "Ошибка аутентификации" -- FIXME by Google Translate
```
geraldus: looks correct, no need to fix (:

yesod-auth/Yesod/Auth.hs
```diff
@@ -110,8 +117,27 @@ class (Yesod master, PathPiece (AuthId master), RenderMessage master FormMessage
     -- destination exists.
     logoutDest :: master -> Route master
 
+    -- | Perform authentication based on the given credentials.
+    --
+    -- Default implementation is in terms of @'getAuthId'@
+    --
+    authenticate :: Creds master -> HandlerT master IO (AuthenticationResult master)
```
MaxGabriel: Not sure what other people think of this, but since `authenticate` gives more information than `getAuthId`, it could be good to recommend that plugin creators implement `authenticate`. (Especially if there's a long-term plan to deprecate `getAuthId`; wasn't sure if that was going to be a thing).

pbrisbin: > Especially if there's a long-term plan to deprecate getAuthId; wasn't sure if that was going to be a thing

I'm not sure this needs to be a long-term plan. Can I deprecate now as part of this change?

snoyberg: That sounds reasonable to me.

snoyberg:
Looks like the only things that need to be done are:

1. Add the deprecation on getAuthId
2. Add `since` comments to the Haddocks to indicate the version that these new data types and functions were added

pbrisbin:
I should have time to add the deprecation later tonight. I expected to leave the `Since` comment to be added whenever this is released -- is that correct, or is there something I should add in the PR?

@snoyberg: to confirm, you're good with the naming here?

snoyberg:
@pbrisbin I'm good with the naming. I'd feel comfortable if you want to do the version bump yourself and add the since comments.

pbrisbin:
Updated and rebased (I left the version commit distinct). Should be ready to go (pending Travis).

snoyberg:
Tested locally and merged, thanks!

PR #962, opened 2015-03-24
Add simple authentication helpers

s9gf4ult:
I think it must be in `yesod-core`

### DIFF
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -52,6 +52,9 @@ module Yesod.Core.Handler
     , lookupCookie
     , lookupFile
     , lookupHeader
+      -- **** Lookup authentication data
+    , lookupBasicAuth
+    , lookupBearerAuth
       -- **** Multi-lookup
     , lookupGetParams
     , lookupPostParams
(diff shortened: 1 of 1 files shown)


snoyberg:
Can you clarify why this must be in `yesod-core`? This looks like it would be a great piece of code to start in a separate library to allow the API to evolve and, once stabilized, consider moving into a more core package.

s9gf4ult:
This is trivial implementation of common standardized methods of authentication. Many users will need this and it looks like core functionality for any web framework. But if you think it must be in separate package I will create some.

snoyberg:
I'm not necessarily opposed, I just wanted to understand the "must" comment, which I think I get now. Here's my recommendation:

* The non-Yesod stuff could likely go into wai-extra instead to be more generally available
* I'd avoid `decodeUtf8`, since it's a partial function. I'd instead use `decodeUtf8With lenientDecode`
* Instead of putting this in a dedicated module, let's just add it to Yesod.Core.Handler

Does that sound reasonable?

s9gf4ult:
Ok, so: 

* move `extractXAuth` to `wai-extra` 
* use `decodeUtf8With lenientDecode`
* move `lookupXAuth` to `Yesod.Core.Handler` 

looks logical for me.

snoyberg:
Cool. Would you be able to send a PR?

s9gf4ult:
Yes, I will prepare PR towards evening.

s9gf4ult:
Ok, ive patched `wai-extra` here https://github.com/yesodweb/wai/pull/352

snoyberg:
Looks good, I'm releasing both to Hackage now. Thanks!

PR #964, opened 2015-03-30
Add support for https on yesod devel

DanBurton:


DanBurton:
Tweaked it to use `race_` as suggested.

snoyberg:
Thanks :)

PR #969, opened 2015-04-15
Updated yesod-bin/Devel.hs

urbanslug:
Replaced deprecated function `configCompiler` with `configCompilerEx`
  Removed unused argument "_opts" from function `lockFile`
  Added newline before guards which I think makes it more beginner friendly.

  Module has no warnings or errors.


snoyberg:
Thanks!

PR #970, opened 2015-04-16
Add `copy-to-args` section to keter.yaml which allows all kinds of options.

wiz:
`copy-to-port` left in place and should be deprecated first.

I've tried to implement combined string/array option but scp can't understand single-string args. List of args works fine though even w/o quoting (but that looks silly):

```yaml
copy-to: user@localhost:/opt/keter/incoming
copy-to-args:
    - "-l 1024"
    - -P 2222
```


snoyberg:
Thanks!

PR #973, opened 2015-04-19
update warp-tls's version

srayuws:
yesod-bin uses ``` Network.Wai.Handler.WarpTLS.tlsSettingsMemory``` which is introduced in wrap-tls-3.0.1


snoyberg:
Thanks!

snoyberg:
I've patched the cabal file on Hackage as well.

PR #980, opened 2015-04-26
Adds export of email verify route

rzetterberg:
This allows users of the library to easily render VerUrl to use in verification emails to new users, if the user was created outside of the regular register functionality.

This would be useful when creating a user management system where the admin creates users instead of them registering.


rzetterberg:
Weird, seems that one of the builders failed (dependency problem) but the other one was successful.

snoyberg:
Thanks! I think I need to report that build failure upstream.

snoyberg:
OK, upstream build failure reported at https://github.com/TomMD/cipher-aes128/issues/9

PR #986, opened 2015-05-11
Yesod init new args name and database

### LINKED ISSUE #974: Supplement "yesod init" interactive mode with non-interactive mode
I was wondering if it might be possible in the future that "yesod init" could accept the app name and database backend as inline arguments to the yesod init command itself. This would allow me to wrap yesod init inside a fish shell function that also did other repetitive things (like initializing a git repo, or generating a dash-haskell docset), by just passing the arguments to the fish function, which in turn would then be able to pass them to yesod init.  It would be nice to retain the --bare flag for both modes of yesod init too.

Thank you for any consideration of this enhancement request :)

triplepointfive:
This resolves #974

### DIFF
--- a/yesod-bin/Scaffolding/Scaffolder.hs
+++ b/yesod-bin/Scaffolding/Scaffolder.hs
@@ -66,16 +66,14 @@ backendBS Simple = $(embedFile "hsfiles/simple.hsfiles")
 backendBS Minimal = $(embedFile "hsfiles/minimal.hsfiles")
 
 validPackageName :: String -> Bool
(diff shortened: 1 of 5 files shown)


snoyberg:
Note that #987 is intended for the same thing. I'm looking at both of these, I'll have to figure out which one to merge.

triplepointfive:
Fair enough

snoyberg:
Since the Travis tests are failing here, I've focused my review on #987 instead.

triplepointfive:
Build is fixed

snoyberg:
Thank you!

PR #987, opened 2015-05-12
Add support for non-interactive mode to yesod init

mbeidler:
This resolves issue #974. I noticed after I had already finished the changes on a clone that someone else submitted a pull request. Nevertheless, I decided to submit it anyway, since the changes were complete.

### DIFF
--- a/yesod-bin/Scaffolding/Scaffolder.hs
+++ b/yesod-bin/Scaffolding/Scaffolder.hs
@@ -3,6 +3,7 @@
 module Scaffolding.Scaffolder (scaffold) where
 
 import           Control.Arrow         ((&&&))
+import           Control.Monad         (mfilter)
 import qualified Data.ByteString.Char8 as S
 import           Data.Conduit          (yield, ($$), ($$+-))
 import Control.Monad.Trans.Resource (runResourceT)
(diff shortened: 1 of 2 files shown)


yesod-bin/Scaffolding/Scaffolder.hs
```diff
@@ -19,15 +20,18 @@ import           Data.Maybe            (isJust)
 import           Distribution.Text     (simpleParse)
 import           Distribution.Package  (PackageName)
 
-prompt :: (String -> Maybe a) -> IO a
-prompt f = do
-    s <- getLine
-    case f s of
+prompt :: (String -> Maybe a) -> Maybe String -> IO a
+prompt f (Just s) = accept f s
```
snoyberg: It seems strange that, if the command line input is invalid, we'd switch to interactive mode. It would seem more logical to just error out in that case.

yesod-bin/Scaffolding/Scaffolder.hs
```diff
@@ -66,26 +70,33 @@ backendBS Simple = $(embedFile "hsfiles/simple.hsfiles")
...
 
-    ebackend' <- prompt $ \s -> if s == "url" then Just (Left ()) else fmap Right $ readBackend s
+    ebackend' <- prompt parseBackend projectBackend
 
     ebackend <-
         case ebackend' of
             Left () -> do
                 puts "Please enter the URL:  "
-                fmap Left $ prompt parseUrl
+                fmap Left $ prompt parseUrl Nothing
```
snoyberg: Again, prompting in this case seems like the wrong choice.

snoyberg:
Sorry, I didn't see that you'd added commits to this to address my comments, it's a good idea to add a comment after doing so as Github sends no notifications. I ended up merging in the other PR in the interim. Thank you for sending the commits. This is actually the first time in the project we've had two people send a PR for the same thing at the same time.

PR #993, opened 2015-05-19
remove shakespeare dependency from Yesod.Core

gregwebs:
This is an experimental PR that I am looking for feedback on.

## Motivation

The motivation is to have a package yesod-core that has no dependency on the shakespeare package. This is beneficial for API servers that have no need for templates.

## Overview

Anything from the shakespeare package is moved into the module Yesod.Widget. None of the other module import Yesod.Widget. This can be confirmed by removing Yesod.Widget and shakespeare from the cabal file and seeing that things still compile. Note that I have not attempted to fix the test suite compilation.

This means that the Yesod.Widget module can now be cleanly moved into a separate yesod-widget package.

## Remaining issues

I ended up taking a few shortcuts to push out this PR tonight: those have a FIXME, but I think they are all fixable. I just realized some of these things can probably be made a lot easier by just copying this type into Yesod.Core

    type HtmlUrl url = Render url -> Html

I will try that tomorrow. Then the main blocker now is to figure out is how to deal with the Yesod typeclass having `defaultLayout :: WidgetT ...`. The main option that springs to mind is to have 2 typeclasses, the 2nd inheriting from the base and adding widget functionality. This could mean adding `Yesod site => YesodWidget site` or changing the current `Yesod` typeclass to `YesodCore` and having `YesodCore site => Yesod site`, where `Yesod site` resides in `yesod-widget`.


meteficha:
Having had already a lot of troubling experience trying to separate code into different modules and not being able to (without hs-boot), I think that having an answer to the typeclass dependency is of utmost importance. I'm not too keen on a YesodWidget typeclass if it ends up creeping everywhere that isn't an API server.

snoyberg:
I really dislike playing the dependency game. It's a stupid circlejerk the Haskell community has gotten into because the cabal dependency solver is so finicky\*. This looks like it's going to make the lives of people wanting to use templates a bit worse. What's the concrete goal for API servers? Avoiding dependency problems? Faster compiles? Smaller executables? I want to be sure we understand what we're trying to accomplish so that we can measure appropriately. For example, if the goal is avoiding a tricky dependency, it might make more sense to put the necessary Css/Javascript datatypes in a different package. If the goal is smaller executables, we need to measure that this really matters.

\* Sorry, did I come off too strong there? ;)

gregwebs:
I do think you are coming off a little strong before doing the right thing and asking to clarify the motivation :)


Goals

* easier to install
* less bloat
* easier inter-operation with other template systems
* perception of both modularity and suitability as an API server


# easier to install

Lets start with this one since it is what you focused on. I don't think shakespeare has been much of a source of installation issues since it was merged into a single package. However, this approach makes the contribution to the bloat issue greater since you get everything from shakespeare even if you only want one thing.


# less bloat

Unecessary dependencies add bloat even after the installation is successful. It slows down the linking phase of every single build unless you build with ghci. It also bloats the final binary. A medium-sized Haskell project ends up depending on over 100 packages, and once lens is added you can easily get closer to 200. One starts looking for ways to drop unnecessary dependencies.


# inter-op with other template systems

Yesod is a web framework. Just as yesod-core does not depend on persistent, yesod-core should not depend on shakespeare. A Yesod user can use any database library  they want. We just provide easy integrations with persistent with yesod-persistent and the scaffolding. The situation with template languages should be the same. We should not have types specific to a package for templates in yesod-core. One should be able to use any template language system that satisfies some contract. 

Unfortunately just moving widget code to a separate module does not fully meet this requirement, and it may indicate more changes are needed. That is, the Widget system itself should easily allow for any template system.


# perception of both modularity and suitability as an API server

Still, JSON servers should have no need for a Widget system designed for server-side templates.
 
The #1 reason people didn't use Yesod in the past is because they are not generating HTML. The documentation has always been biased towards the HTML use case. I showed this problem to Michael, and some changes to the book and docs were made. Still, users that have no requirement for HTML are very hesitant to choose Yesod and instead reach for the Scotty clones or now Servant. 

Personally I would like to see a web framework that has modular components that can be dropped when they are not being used, even though a default scaffolding would still present the same great Yesod experience we have today.

gregwebs:
The goal of opening this PR is to get the help from others to figure out how to make this kind of change while making the lives of existing template users a bit better, not a bit worse.

Right now I am thinking that the best approach is moving most of the current `Yesod` typeclass to a `YesodCore` typeclass: that typeclass has nothing to do with templates. The current `Yesod` typeclass then inherits from `YesodCore` and adds things that have to do with widgets. It would need to be placed in the `yesod` package instead of the `yesod-core` package. I don't believe this makes the lives of new or existing yesod users more difficult. Existing users can use everything the same way. Users will have the benefit of seeing the `YesodCore` typeclass and knowing that the functionality does not need widgets.

The main pain point would be for library authors that depend on yesod-core: the package shuffling would be disruptive.

snoyberg:
I'm not going to disagree with this being (at least in theory) the right way to design things from the outset. I am still hesitant to agree to any kind of breaking changes, as Yesod is at a stable point right now. On the other hand, I've been thinking about the "Yesod for API servers" quite a bit also, and see the allure of it.

I think what I'm getting at is: the question of whether we can absorb breaking changes like this at our current stage of development is the real issue, and I'm not going to be able to make that decision myself. Consider me at about -0.5 on it. If I heard back from the broader community that they're completely behind this, that -0.5 would change to +1 immediately. But I have a feeling that's not what we're going to hear.

gregwebs:
Most current Yesod users are OK with how things work now. The natural reaction for most people to change is to be opposed. Those not using Yesod usually don't realize what they are missing and they certainly aren't going to switch all their projects to Yesod. So the biggest target of this change is those who are not yet in the Haskell community so it isn't really possible to poll them for their opinion.

At this point though we haven't really figured out what exactly the breakage would be. I think we can get the breakage for existing users close to zero or at least provide a very simple migration guide and the only real problem would be breaking library authors.

gregwebs:
I just added another commit so that the dependency chain is now

Yesod.Core -> Yesod.Widget -> Yesod.Shakespeare

So a Widget itself has no dependency on Shakespeare anymore. This gives us 2 options for how to split things up:

* yesod-widget package: contains Yesod.Widget and Yesod.Shakespeare
* yesod-shakespeare package: contains Yesod.Shakespeare, so yesod-core still contains Yesod.Widget

The second form could help a little to reduce breakage since more things can still come from the yesod-core package. On the other hand, this commit introduces a new breakage:
the removal of the `Javascript` newtype in `GWData`. I will definitely need to compile the tests and otherwise dig deeper into the repercussions of that.

Note that I did not attempt to remove the dependency on blaze for the Html type.

paul-rouse:
Just a quick opinion: I'd be fine with a breakage at the library-author level, but admittedly I only look after the rather small hashdb auth plugin.  Unless I've missed something, there is only likely to be the odd type constraint and import to change.  However, I do agree that migration for ordinary users needs to be simple and minimal.

meteficha:
This may be Yesod "2.0" territory due to the breakage it will cause.

If we're going to bite this bullet, how about considering making the Widget system Yesod-agnostic?  By this, I mean moving most of the code into a framework agnostic package (e.g., shakespeare package doesn't care about Yesod), and making `yesod-widget` just connect the pieces at the right places.  I'm not sure how feasible this idea is and what advantages/disadvantages it would bring, I'm just throwing it out there.

Also, we have to remember that if one used a single package that supports both JSON and HTML, then it would bring `yesod-widget` and everything else with it.  For example, if you used `yesod-auth` for your authentication needs, you'd need `yesod-widget` as well.  To avoid this we'd have to either:

  * Split all the auth packages in half, one with just JSON, the other with HTML as well.

  * Make some JSON-only auth packages, even if duplicating code.

  * Tell people with API servers to implement auth from scratch everytime.

meteficha:
Personally I think that Yesod could be better positioned for API servers in the following ways:

  1. Better documentation of how to do it.

  1. Better examples of people doing it.

  1. Scaffold support with sane defaults (e.g., code without templates, server that never responds with HTML).

  1. Examples of strictly REST APIs, including links etc. 

  1. Examples of auth that doesn't use cookies.

These points are ordered from least to most difficult to do by oneself today, IMHO.

Regarding @gregwebs points:

  * I don't think that "easier to install" and "less bloat" are relevant, even though they're nice.

  * "easier inter-operation with other template systems" could be done with the current hard dependencies if the widget system was made more modular.  While it's an interesting goal, I don't think it's a good motivation for the split.

  * "perception of both modularity and suitability as an API server" is what I think is most important, but I think it's better adressed by the points above than by changing dependencies.

gregwebs:
> If we're going to bite this bullet, how about considering making the Widget system Yesod-agnostic?

With the new commits that is now the case on this PR (I will also add another that actually gets Yesod.Shakespeare to compile). I am now favoring keeping Yesod.Widget in yesod-core but having a separate yesod-shakespeare package. This means I could even attempt to maintain the same single Yesod typeclass interface if it helped to minimize breakage.

You also bring up a good point about yesod-auth needing to be split up for one to have an API-server only approac. I think it is important to keep in mind that yesod-auth was designed from the beginning for all the complexities of an HTML-based login where there are anonymous and authorized routes and login forms and user flows. The support for JSON is really for sites that need all that complication but just want to do some parts with AJAX instead of refreshing the page. For an API server where every request must authenticate there may be a very simple amount of code for authentication that can exist entirely in yesod-core.

meteficha:
Expanding a bit on my point above about auth without cookies, I think that it would be best if we had native support for a sane, minimal subset of OAuth 2.0 on the server-side.  Sane, because OAuth 2.0 is a mess, unfortunately.  Subset, because we want something that just works like clientsession + yesod-auth today.  Minimal subset because people require different parts of OAuth, and it's difficult to provide a composable library to support them all---we'd have to be opinionated. 

In fact, throwing again random things at the air, our current auth system has conflated two similar but different concepts:

  * What kind of authentication is used.  Examples:
    * E-mail and password (Yesod.Auth.Email).
    * Username, e-mail and password (yesod-auth-account).
    * Google account (Yesod.Auth.GoogleEmail2).

  * How that authentication is performed.  Examples:
    * Setting a session variable on the existing session vs handling a token.
    * An HTML form vs JSON submission (this has been getting better with support for both, but still JSON is a 2nd-class citizen).

Ideally, on my API server I'd like to be able to choose from the kind of authentications that already exist, but with an interface that doesn't exist.  I think we can hack it today by having a session backend that hijacks the "_ID" key and turns it into an access token, but this is subpar.  Perhaps with some insight we could break each plugin in two parts (even if on the same package):

  * Data types and functions defining inputs, outputs and authentication methods.

  * Interfaces for these data types using HTML forms and JSON APIs.

Since I'm already in berserk mode, some people care about the security advantages of having server-side sessions, especially since they're the only revocable ones [1].  We don't have that currently.

[1] Well, you could include a token on a client session and check it on your database, but this kind of defeats the purpose of having a client session in the first place.

meteficha:
> For an API server where every request must authenticate there may be a very simple amount of code for authentication that can exist entirely in yesod-core.

I agree, and part of my comment above is me saying that this simple amount of code should be shared and used by the complicated HTML code, at least ideally.

gregwebs:
> I agree, and part of my comment above is me saying that this simple amount of code should be shared and used by the complicated HTML code, at least ideally.

yes, exactly. This change can be made independent of this PR and possibly in a backwards-compatible way.

snoyberg:
I'm not going to be able to focus on this for the next week or so, I'm going to have all my attention devoted to some other projects. Don't take my silence as objection, just not having had a chance to devote the necessary cycles to this (important) topic. Please don't let my absence stop progress on plans here though, and I will get back to this in a week or so (and if I forget, feel free to ping me).

gregwebs:
This PR is basically finished now. It ended up just focusing on making Yesod's Widget not depend on Shakespeare and creating a separate Yesod.Shakespeare module that could be placed in a yesod-shakespeare package. More things are moved into the Yesod.Widget module now. However, to maintain better backwards compatibility we are likely going to want to re-export them from Yesod.Core instead of forcing the import of Yesod.Widget.

Note that yesod-core still depends directly on aeson and blaze. To get rid of the shakespeare dependency but minimize breakage I just translated some hamlet down to blaze (with -ddump-splices).

The splitting of Yesod.Shakespeare into yesod-shakespeare creates a breakage: some things that were previously imported from Yesod.Core will now need to be imported from either Yesod.Shakespeare or Yesod

BlakeRain:
Some thoughts:

I maintain a number of API servers written using Yesod. I'm not entirely sure why the inclusion of shakespeare into Yesod would make someone think Yesod is unsuitable. Having said that, it's also quite useful to have shakespeare there when you need it. For example (I am sorry to admit) I do often use xhamlet for generating XML responses. Ha.

I cannot recall ever having had a problem using a different template system to shakespeare when I've needed to. From what I can recall, the ToWidget type class was perhaps the only issue that I had. That and the fact that Yesod is quite complex, so knowing how to go about it requires a good understanding of Yesod and how it works.

Whilst I know this PR is not about stopping me from doing the things I do, and a more modular system can have it's benefits, I'm concerned that perhaps too much that is fundamental to Yesod is being changed.

In terms of building APIs, one small issue that does arise is that the JSON responses from the default `errorHandler` are not from a ToJSON instance of `ErrorResponse`. As we have some systems that talk to our APIs, not having a `FromJSON` instance for `ErrorResponse` can be a little troublesome.

Authentication is not particularly difficult with Yesod. I've so far avoided OAuth 2.0 thankfully, although we do use OAuth 1.0a internally. Unfortunately all external consumers of our APIs choose to use HTTP Basic.

Perhaps @meteficha is right in that there is not enough documentation/examples of Yesod as an API server.

rehno-lindeque:
> I think what I'm getting at is: the question of whether we can absorb breaking changes like this at our current stage of development is the real issue, and I'm not going to be able to make that decision myself. Consider me at about -0.5 on it. If I heard back from the broader community that they're completely behind this, that -0.5 would change to +1 immediately. But I have a feeling that's not what we're going to hear.

I just thought I'd pitch in and mention that we're already using a cut down version of Yesod for [CircuitHub](https://circuithub.com). It does make updating Yesod a little bit tougher on us, so I'd certainly appreciate a change like this - especially if it would help to bump down Yesod's [sandbox build times](https://halcyon.sh/shootout/#results) a little bit.

The Yesod server is battle-tested, it's a silly thing to lose developer mind-share simply because they're not interested in templating for whatever reason.

meteficha:
@rehno-lindeque, why don't you use vanilla Yesod?

rehno-lindeque:
@meteficha It was originally an experiment to see if I can make Yesod more lean and cut down on sandbox builds for deployment purposes. I haven't had any reason to revisit the issue again.

Never-the-less,
```bash
$ cabal sandbox init
$ cabal install shakespeare shakespeare-css shakespeare-i18n shakespeare-js shakespeare-text hamlet --only-dependencies
...
$ time cabal install shakespeare shakespeare-css shakespeare-i18n shakespeare-js shakespeare-text hamlet
...
real	0m44.342s
user	0m44.941s
sys	0m4.076s
```
If the patch managed to save my team 60 seconds of [disruption](https://xkcd.com/303/) per day over the past 200 days, then I'd call it a win. It probably only took me about an hour or two to rip out in the first place.

Besides, all else being equal - and for my own mental well-being - I prefer spending an hour getting acquainted with the internals of the [yesod](http://hackage.haskell.org/package/yesod) package to spending an hour spread out over the next year waiting on a computer to do its thing.

I am sure that there are more people like me that spend an hour or two fiddling with Yesod and wondering why they have to bundle in this html/css stuff when all they really want is a JSON service. For better or worse people are accustomed to super modular frameworks like [koa](http://koajs.com/). Plugging in a little bit of templating doesn't seem like it should be a big deal in this day and age.

meteficha:
Thanks for the detailed explanation. 

But I can't help but think it's probably saving less than one minute since you should not be installing the deprecated packages and you should be using parallel builds. And, in the end, if I had to make a sandbox rebuild daily and that blocked my work, it would be terrible with or without that half minute of CPU time.

OTOH, I agree that as a matter of principle it's better for yesod-core not to depend on shakespeare, even if I think that the practical reasons are overrated :).

rehno-lindeque:
> But I can't help but think it's probably saving less than one minute since you should not be installing the deprecated packages and you should be using parallel builds.

Oh good point on the deprecated packages, only 28 seconds for `shakespeare` it looks like. I also agree that there are ways to pay less through caching etc. I think that what you gain in optimizations you lose again in hidden costs like larger sandbox downloads (on deployments, CI etc), link times as well as the collective cost spread out over a team of people.

I think everyone generally agrees that this isn't a big technical issue (aside from perception, which I do think is a problem from a marketing point of view), but small victories win the war.

rehno-lindeque:
P.S. Sorry for beating a dead horse. I'm not actually arguing with anyone in particular (I actually think there are some very valid concerns in the comments above). I just happened to notice the top two stories on reddit today are:

* [Should I bring Haskell into a corporate environment and if so, how?](http://www.reddit.com/r/haskell/comments/371xz9/should_i_bring_haskell_into_a_corporate/) - has no mention of Yesod so far:


> I would use scotty or spock. aeson for the JSON (remember to use Generic to derive ToJSON/FromJSON instances automatically for your records). lucid if you want to generate HTML and don't want to deal with separate template files.

* [What do Haskellers want? FP Complete survey of over a thousand users](http://www.reddit.com/r/haskell/comments/36z17n/what_do_haskellers_want_fp_complete_survey_of/)


> I work on Haskell projects in the neighborhood of 20k - 50k lines (50k - 120k lines including non-hackage dependencies) and compile times have become a pretty big development slowdown. This issue is very high on my list of pain points and getting higher all the time.


> It still takes > 10 minutes to install scotty in a sandbox, and a lot more for a complex stack like yesod. Sandboxes are a band aid on more fundamental problems that needs to be fixed - binary downloads, and multiple concurrently installed package versions.


> Anyone brand new to the language. I struggled mightily with cabal hell trying to install yesod until I realized I had two different GHCs on my Mac and the older version's global packages (alex, happy) were symlinked into my PATH. Once I fixed that issue, then of course it took two hours to compile yesod.
> I got through it and now I'm happily building a web app with yesod (which is a freaking amazingly well thought-out framework) but man, those first couple of days were rough. I almost gave up and went crawling back to ruby on rails... Glad I didn't though.

I'm not saying there aren't workarounds for these problems, but newbies (and even experienced devs as evidenced by the 20k-150k quote) never-the-less feel that they are running into them. I'm not saying that this PR is *the* solution either as it clearly isn't, but it might not hurt to tackle a couple of small things to create a more lean Yesod experience where possible.

snoyberg:
As I'm reading the current state of things:

1. Everyone seems to agree that in an ideal world, writing things from scratch, it would be the right breakdown to have a lower-level, HTML-free layer, and add an HTML-friendly one on top.
2. We may be able to get away with moving over to the HTML-free layer with minimal user-facing breakage. However:
    * There will certainly be library level breakage
    * It's unclear how things like yesod-auth will play out in this new world, since JSON services would like authentication, but yesod-auth should be able to generate HTML responses for those use cases
3. There's broad agreement among Yesoders that Yesod is definitely capable as acting as a web service framework, but there's a misperception by non-Yesoders that it doesn't fit that role.
4. The extra HTML-based code does add some CPU time on compiles, and some size to compiled binaries, but it's unclear if that's significant.

Are there any disagreements on these points?

To me, it seems like (3) is the right one to focus on first, since making Yesod more obviously JSON-friendly is something that will have immediate benefit, will help in both the current library setup and a theoretical split library setup, and that work may in fact guide how we decide to break up the libraries. It's also work that can take place without upsetting the current libraries at all.

My concrete proposal is: we\* start a new project to create a Yesod-based library for creating JSON services (we could resurrect the now unused yesod-json name). We design something wonderful, make beautiful documentation, and when it's ready, really fold it in with the rest of the Yesod ecosystem. Then, we can assess how big a burden the current combined lower layer is providing.

I'll admit to some bias here, because (1) I've wanted a yesod-json library myself like this, and (2) I don't feel like having the book and other documentation I've written suddenly be invalidated yet again.

\* For some value of "we"

gregwebs:
I disagree on point 2 about yesod-auth: Felipe and I already came up with a better approach.

I don't know what problem the creation of a yesod-json library would actually be solving.

I don't think it is possible to change the perception of yesod without at least dropping the shakespeare dependency. So I definitely want to move forward with this patch (i should be getting back to it soon) starting with an initial completely backwards-compatible release but then splitting off a yesod-shakespeare package.

These changes are not removing the html dependency on blaze-html (or on aeson), which would be much more disruptive.

gregwebs:
moving this over to #1004

gregwebs:
I think this is a good example of how the community thinks. Here someone creates a list of possible REST API servers and nobody mentions yesod

http://www.reddit.com/r/haskell/comments/3a4qrl/libraryframework_suggestions_for_a_rest_api_in/

meteficha:
@gregwebs To be fair, the OP on that link said he was using Yesod and looking for alternatives, so it makes sense nobody is suggesting Yesod :).  But, yes, no one disputed OP's claim that Yesod was heavy.

cdepillabout:
@gregwebs, @meteficha I am the OP from that reddit post.

In the Slack http://fpchat.com #haskell channel, I was asked what I didn't like about Haskell.  This was my response:

> I actually really like Yesod.  Like I said in that reddit post, I will definitely continue using
> Yesod in the future.  But for a (simple?) REST api, it just feels a little a little "heavy".  I just
> don't fee like I'm that fast with it.
>
> Now, I realize that this is kind of a disservice to the community.  It's not possible for you to
> argue against my "feeling".  I guess I just want to play with some other libraries and
> frameworks to see how they work.  Maybe then I'll get a better feeling for where Yesod
> could be improved, and then I could give some helpful feedback.

So I don't actually have any helpful constructive critism of Yesod :-\  But I do really like Yesod, and I will be watching this thread and #1004 intently.

gregwebs:
@cdepillabout it would be helpful if you could identify what creates the perception of "heavy". The shakespeare dependency is one thing which I have identified.

PR #1000, opened 2015-05-27
add a Dockerfile for haskell development

gregwebs:
Mostly just the base haskell image plus a non-root user and the stackage-tooling.
I may be looking to find a different home for this Dockerfile in the future since it isn't really yesod specific.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,9 @@ cabal.sandbox.config
 /vendor/
 .shelly/
 tarballs/
+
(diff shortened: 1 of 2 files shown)


geraldus:
Very helpful, thank you @gregwebs

gregwebs:
@geraldus you are welcome!

I am merging this PR now because it won't break anything. I am open to documenting its usage more or finding a different home for the Dockerfile.

PR #1001, opened 2015-06-03
Export log formatting

ygale:
Make the advice in the `makeLogger` haddock practical:

For logging during app startup, it recommends creating a `Logger`, using during startup, and then supplying it to `makeLogger` when creating the `Yesod` instance. But for that to be practical, you need access to the default implementations of the various logging-related methods of `Yesod` and/or the `formatLogMessage` function. Unfortunately, none of those are available before the `Yesod` instance is actually created, leading to a chicken-and-egg problem.

This PR moves the default implementations of the logging-related methods to separate functions that don't depend on the `Yesod` instance, and exports all of those plus `formatLogMessage` from `Yesod.Core`.

### DIFF
--- a/yesod-core/Yesod/Core.hs
+++ b/yesod-core/Yesod/Core.hs
@@ -30,6 +30,11 @@ module Yesod.Core
     , AuthResult (..)
     , unauthorizedI
       -- * Logging
+    , defaultMakeLogger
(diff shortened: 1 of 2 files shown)


gregwebs:
Great!

snoyberg:
Awesome, thanks!

Issue #1002, opened 2015-06-03
Exposing two internal functions


andrewthad:
Could the functions `subHelper` and `stripHandlerT` be exposed? I occasionally build subsites without using yesod's routing TH (to get non-monomorphic routes), and I have copied these functions out into my project because they're so dang helpful for this. My current solution works fine, but I think that it would be nice to not have duplicated code in my project.

snoyberg:
Yes, I'm good with that. Can you send a PR?

andrewthad:
https://github.com/yesodweb/yesod/pull/1003

andrewthad:
The PR is ready

PR #1003, opened 2015-06-05
Make stripHandlerT and subHelper available for public use

andrewthad:
I haven't compiled this to check it, but I'm pretty sure it works. I assume that Travis should catch any compile errors if it's broken.

### DIFF
--- a/yesod-core/Yesod/Core/Dispatch.hs
+++ b/yesod-core/Yesod/Core/Dispatch.hs
@@ -30,7 +30,7 @@ module Yesod.Core.Dispatch
     , defaultMiddlewaresNoLogging
       -- * WAI subsites
     , WaiSubsite (..)
-    , module YCCD
+    , subHelper
     ) where
 
 import Prelude hiding (exp)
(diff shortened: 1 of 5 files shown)


andrewthad:
https://github.com/yesodweb/yesod/issues/1002

MaxGabriel:
Since these functions are being made public, it'd be nice if `stripHandlerT` was haddock-documented, and potentially the `subHelper` function documentation improved (the current documentation kind of indicates its just for internal use).

andrewthad:
@MaxGabriel That's a very good point. I will add some documentation to them.

andrewthad:
@MaxGabriel I've improved the documentation (although it's still pretty lackluster). Let me know if you have any suggestions.

andrewthad:
This should be ready to go now

yesod-core/Yesod/Core/Dispatch.hs
```diff
@@ -30,6 +30,7 @@ module Yesod.Core.Dispatch
     , defaultMiddlewaresNoLogging
       -- * WAI subsites
     , WaiSubsite (..)
+    , module YCCD
```
snoyberg: Why not just export `subHelper` directly?

andrewthad: Good point. I've now changed it to do that instead.

snoyberg:
Thanks for fixing that, I'll just wait for Travis to confirm and then this is good to go.

Issue #933, opened 2015-02-13
Per-environment configuration in new scaffolding


joelteon:
It is now non-obvious how to have different configuration settings for different environments.

PR #1004, opened 2015-06-07
Core without shakespeare

gregwebs:
This is a rebased and updated version of PR #933
## Goal
-  yesod-core has no dependency on the shakespeare package
-  yesod's widgets can work with other templatig packages
## Motivation
- increase the modularity of the framework
- satisfy the user desire to not have a dependency on shakespeare
### Overview

Anything from or relying on the shakespeare package is moved into the module `Yesod.Shakespeare` which is be broken out into a separate `yesod-shakespeare` package.
### User perspective
- Users requiring just yesod-core (and using shakespeare functionality) will need to also require yesod-shakespeare.
- Where shakespeare functionality is required and there is only an import of `Yesod.Core`, there are 2 options
  - add `import Yesod.Shakespeare`
  - change they `Yesod.Core` import to just `Yesod` (no `.Core). This options is only available for those that depend on`yesod`.

In addition, some instances have been moved to `Yesod.Core.Widget` (which is exported by `Yesod.Core`). So it is possible that a user may been to add `import Yesod.Core ()`, although it is unlikely that that import is not already in scope.
### Html and Value are still part of yesod-core

As much as possible, functionality relating to Html has been moved to `Yesod.Widget`, although blaze-html and aeson are still required by the core yesod typeclass.

Some shakespeare code was translated to blaze-html (with the help of -ddump-splices).
### Type changes

widgets (`gwdJavascript` of `GWData`) no longer use the `Javascript` newtype so that they are not tied to `shakespeare`. All of the functions that interact with `gwdJavascript` have been changed according. so I don't expect this to affect anyone.
## TODO
- [ ] version bump (all?) yesod packages to 1.5


meteficha:
Some general comments without having read the code yet:

  * It would be nice if you rebased everything.

  * From your overview it looks great.  Things you didn't say (and I didn't look at myself):

    * Does the `yesod` package pull & re-export `yesod-shakespeare`?

    * Is it implied that users will, at most, have to import `Yesod.Shakespeare`?

gregwebs:
Great questions, I will update the description accordingly

* I did some more rebasing
* I added a commit to re-export `Yesod.Shakespeare` from `Yesod` (of the `yesod` package).
* users requiring just `yesod-core` will need to also require `yesod-shakespeare`. Where shakespeare is required and there is only an imports of `Yesod.Core`, an additional `Yesod.Shakespeare` needs to be added, or it should be changed to import `Yesod` (no `.Core).

gregwebs:
The Travis build passes now and this is ready to merge. Note the TODO for version bumping in the description: I don't want to bother with that yet because this PR may take some time to review and it can create merge conflicts.

Because yesod-shakespeare is used by other packages and yesod-core has many test cases, I have good confidence in these changes even though there are a lot of them
I would like to keep this PR limited to technical feedback. For a discussion of whether or not the idea of the change is good, I think it is better to keep using #933 or the yesodweb mail ist.

snoyberg:
I wrote up a bunch of questions while reading through it, but I think I understand what you're doing better now and therefore the questions don't apply. It seems like you're going beyond simply "drop the shakespeare dependency," but also trying to put as much HTML-centered functionality in Yesod.Shakespeare. Example: `redirectToPost` could reasonably be kept in yesod-core, but you moved it out anyway. Is this an accurate assessment?

I think I'm on board with this, and I also think it's intended to pave the way towards a Yesod 2.0 that completely gets rid of HTML stuff from yesod-core. Am I reading that correctly?

The one thing I'd like to avoid is unnecessary orphans, such as the orphaned `ToTypedContent Html` instance, which seems on first blush like it doesn't help anyone.

There are individual points here and there that I might disagree with, but not strongly at all. I think you've convinced me that this kind of change is worth the breakage, consider me on board :).

gregwebs:
Thanks for re-considering and taking a closer look and pointing out that I did not explain the changes well enough.

yesod-shakespeare is really just for moving the shakespeare dependency out. So I should probably actually move `redirectToPost` back to core (I just need to compile it down to blaze-html first, which is why it got moved out originally).

I will look over the orphans (which reminds me that I need to look over compilation warnings) and also at moving redirectToPost (and maybe some other things) back to yesod-core.

But the other thing I did  that is not explained in the description is to start moving anything `Html` focused over to `Yesod.Widget`. This is a source of some of the orphans. `Widget` is really a bad name that means: things needed to generate a piece of a web page. So in a pure API server, Widget is not needed, so moving anything Html over to Widget now is a step towards that direction.

I agree that a yesod-2.0 would have a `yesod-webpage` package that has the widget functionality. But it will create a whole lot of breakage to remove `Html` from yesod-core right now. This shakespeare movement combined with moving more `Html` into `Widget` is something with minimal breakage that gives a baby step in this direction.

gregwebs:
I moved `redirectToPost` back to core and fixed a bunch of warnings.

Again, all of the orphans not due to the yesod-shakespeare split out should be from moving Html related code into the Yesod.Core.Widget module. I don't believe that moving instances within a package should cause orphaning problems. They just have the potential to cause breakages to someone who is not just importing from Yesod.Core. So I will also list that in the user facing changes.
I am open to getting rid of that Html-based maneuvering, but as stated I think it is the right direction in the long term.

gregwebs:
should take care of https://github.com/snoyberg/classy-prelude/pull/92 first

snoyberg:
@gregwebs I'm officially signing off on this change, merge when you're ready. Since this is a breaking change, it's worth merging in the other breaking changes at the same time.

gregwebs:
what are the other breaking changes?

MaxGabriel:
I have 2 PRs introducing small breaking changes. They're labelled "next
major release" or something (on mobile so can't confirm the exact name).

creichert:
I haven't given much input on this one but I want to mention that I am very much in favor of this change.

seanwestfall:
Hey @gregwebs how much more work is involved in finishing this PR? I'd actually like to see this change happen as well, and it doesn't look like there's been a lot of commits in awhile.

gregwebs:
@seanwestfall the change was completed at the time it was submitted. It just needed to be reviewed and released. Now it just needs to be merged and released.

seanwestfall:
@gregwebs Did you see this PR? https://github.com/yesodweb/yesod/pull/1372

jezen:
@gregwebs is this PR still valid? It's been sitting for a number of years, and I think it might now be a struggle to reconcile it with the current master especially after the breaking changes introduced in 1.6.

jezen:
Closing due to extensive inactivity.

PR #1008, opened 2015-06-09
Is there a reason why add-handler doesn't generate a spec file?

nathankot:


MaxGabriel:
Thoughts on if this should have a command line flag to disable adding the test portion? I imagine people have different test directory structures, or have swapped out Hspec for something else, or just don't want to write tests for some handlers. On the other hand, it's easy to just delete the generated file afterwards.

nathankot:
@MaxGabriel to be honest, I run into trouble getting `pending` to work :S how is it normally done with yesod tests?

```
it "spec blabla" $ pending
```

gives me `Couldn't match type '()' with 'App'`

nathankot:
@MaxGabriel since add-handler makes assumptions about your `handler` dir and the scaffold also includes specs in the `test` dir I assumed it was 'the standard way', I imagine people with custom file structures would avoid using the scaffold tool all-together?

For those who don't write specs at all it might be good idea to give them a friendly nudge haha, then again imagine the amount of times you would forget to append the `--no-specs` flag lol

For swapping out Hspec, it's still easier to edit an existing file than it is to make a new one :stuck_out_tongue: 

Either way, it's not a lot of work so your call!

MaxGabriel:
> imagine the amount of times you would forget to append the --no-specs flag lol

Great point. I would never remember that. I withdraw the suggestion to add such a flag.

yesod-bin/AddHandler.hs
```diff
@@ -78,7 +79,11 @@ addHandlerFiles cabal (name, handlerFile) pattern methods = do
     modify cabal $ fixCabal name
```
snoyberg: I believe `fixCabal` also needs to be modified so that it adds the Spec module to the test suite.

nathankot: do you mean `exposed-modules`? My `yesod init` came stock without `HomeSpec` being added to the cabal file

snoyberg: Good point

snoyberg:
Thanks!

PR #1011, opened 2015-06-12
Use nonce package to generate verification keys and CSRF tokens

geraldus:
Also removed almost all warnings and applied **stylish-haskell**.

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -64,7 +64,6 @@ import qualified Yesod.PasswordStore as PS
 import qualified Text.Email.Validate
 import qualified Yesod.Auth.Message as Msg
 import Control.Applicative ((<$>), (<*>))
-import Control.Monad (void)
 import Yesod.Form
 import Data.Time (getCurrentTime, addUTCTime)
 import Safe (readMay)
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -49,13 +49,12 @@ module Yesod.Auth.Email
     , defaultSetPasswordHandler
     ) where
 
-import Network.Mail.Mime (randomString)
 import Yesod.Auth
-import System.Random
 import qualified Data.Text as TS
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as TE
 import qualified Crypto.Hash.MD5 as H
+import qualified Crypto.Nonce as Nonce
 import Data.ByteString.Base16 as B16
 import Data.Text.Encoding (encodeUtf8, decodeUtf8With)
 import Data.Text.Encoding.Error (lenientDecode)
@@ -69,6 +68,8 @@ import Control.Monad (void)
 import Yesod.Form
 import Data.Time (getCurrentTime, addUTCTime)
 import Safe (readMay)
+import System.IO.Unsafe (unsafePerformIO)
+
 
 loginR, registerR, forgotPasswordR, setpassR :: AuthRoute
 loginR = PluginR "email" ["login"]
@@ -163,9 +164,7 @@ class ( YesodAuth site
     --
     -- Since 1.1.0
     randomKey :: site -> IO Text
-    randomKey _ = do
-        stdgen <- newStdGen
-        return $ TS.pack $ fst $ randomString 10 stdgen
+    randomKey _ = Nonce.nonce128urlT defaultNonceGen
 
     -- | Route to send user to after password has been set correctly.
     --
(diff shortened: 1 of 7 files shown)


meteficha:
`nonce` wasn't on Stackage, which would be a problem, but I just sent a PR to fix that. https://github.com/fpco/stackage/pull/632

geraldus:
One way I can create a single generator here is to use `unsafePerormIO`, but I'm not certain is it good practice too.

Also GHC 7.10.1 build failed already, and error message looks odd for me.

> I just sent a PR
Great!

meteficha:
I'm not sure what you mean by using `unsafePerformIO`.  Do you mean creating a global generator via the non-inline `IORef` trick?

geraldus:
@meteficha yeah

```haskell
nonceGenerator = unsafePerformIO Nonce.new
```

geraldus:
@snoyberg sorry for noise, so should I go the way Felipe suggested?

snoyberg:
I'm OK with the `unsafePerformIO` approach in this case, but I don't know the internals of `nonce`. I leave the decision about which way is better to Felipe, who actually knows what he's talking about here (as opposed to me, who is essentially speaking from pure ignorance).

meteficha:
I don't like the `unsafePerformIO` trick in general, but there's no harm in using it here wrt `nonce`.  I think @snoyberg should have the final say regarding `Yesod.Auth.Email`, so I say go ahead :).

geraldus:
@snoyberg @meteficha updated

phew, finally I've found time to make updates

geraldus:
oh, I have to rebase to master first, sorry will do in a moment!

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -586,3 +586,7 @@ setLoginLinkKey :: (YesodAuthEmail site, MonadHandler m, HandlerSite m ~ site) =
 setLoginLinkKey aid = do
     now <- liftIO getCurrentTime
     setSession loginLinkKey $ TS.pack $ show (toPathPiece aid, now)
+
+
+defaultNonceGen :: Nonce.Generator
+defaultNonceGen = unsafePerformIO (Nonce.new)
```
meteficha: I think you need a `NOINLINE` pragma.

yesod-auth/Yesod/Auth/GoogleEmail2.hs
```diff
@@ -549,3 +550,7 @@ allPersonInfo (A.Object o) = map enc $ M.toList o
     where enc (key, A.String s) = (key, s)
           enc (key, v) = (key, TL.toStrict $ TL.toLazyText $ A.encodeToTextBuilder v)
 allPersonInfo _ = []
+
+
+defaultNonceGen :: Nonce.Generator
+defaultNonceGen = unsafePerformIO (Nonce.new)
```
meteficha: Same here.

meteficha:
After the `NOINLINE` pragmas are added, the `nonce` part of things will look good to me. :+1: 

BTW, @geraldus, did you test it?

geraldus:
Your mean exactly these changes?  No, I didn't.  I can test this a bit later.

geraldus:
@meteficha updated

geraldus:
OK, I've created simple project in sandbox:
```bash
$ mkdir nonce-test
$ cd nonce-test
$ cabal sandbox init
$ cabal sandbox add-source ~/forks/yesod/yesod-core
$ cabal sandbox add-source ~/forks/yesod/yesod-form
$ cabal sandbox add-source ~/forks/yesod/yesod-auth
$ cabal sandbox add-source ~/forks/yesod/yesod-persistent
$ cabal init
```
Here is [example code][ex].

I was able to login using Google API and create new user and verify registration via email.  I don't know better way to test this.

I've faced only one issue related to `nonce` package with one of my projects so far: I use `markdown` package to render verification letters sent to users, there is something like this
```md
Verify your email by clicking on [this link](#{ver_url})
Direct link: #{ver_url}
```
and if there is two underscores in verification key (e.g. YzdCYQN0wsYD_0_GhSf6ZIsY), markdown renders direct link incorrect, italicizing part of it.
[ex]:https://gist.github.com/geraldus/5db4bf42c73ea8035a6b

meteficha:
It seems to me that the presence of underscores is not a big deal.  @geraldus, can you confirm that using the following works for you?

```markdown
Verify your email by clicking on [this link](#{ver_url})
Direct link: `#{ver_url}`
```

@snoyberg, looks good to me.

geraldus:
@meteficha yes, I confirm.

snoyberg:
Awesome, thank you!

Issue #732, opened 2014-05-07
yesod-static: perform hash calculations in a separate thread


andrewthad:
I've noticed that on the site I've built that has around 30,000 files uploaded (they go in the static folder), it takes around 60 seconds to start warp. If I move these files elsewhere (so that they wouldn't be served in the static subsite), warp starts almost immidiately, just like it does locally. Does anyone know why this would be happening?

snoyberg:
Lower priority, closing.

PR #1013, opened 2015-06-15
Parallelize hashing of static files

dirk:
This is for #732. It parallelizes the hashing of static files by using `Control.Parallel.Strategies.parMap` from the [parallel](https://hackage.haskell.org/package/parallel-3.2.0.4/docs/Control-Parallel-Strategies.html#g:7) package.

### DIFF
--- a/yesod-static/Yesod/Static.hs
+++ b/yesod-static/Yesod/Static.hs
@@ -100,6 +100,7 @@ import System.Directory (createDirectoryIfMissing)
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Encoding as TLE
 import Data.Default
+import Control.Parallel.Strategies
 --import Text.Lucius (luciusRTMinified)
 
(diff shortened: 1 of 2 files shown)


gregwebs:
looks good. Did you do some timing tests?

snoyberg:
I don't think this will make any difference. This code will construct a list of actions in parallel, but will still perform that list sequentially. We'd need some kind of a thread pool or the like to handle this (which isn't too difficult to achieve with async, e.g. [Control.Concurrent.Execute](https://github.com/commercialhaskell/stack/blob/master/src/Control/Concurrent/Execute.hs)).

dirk:
@gregwebs: Still getting up to speed; will hopefully have some soon.

@snoyberg: Had a suspicion that was the case; will experiment some and verify. Thanks for pointing to the right direction!

snoyberg:
I'm going to close this for now, as it's not ready to be merged. When you're ready (or want more review), please open a new PR.

Issue #1016, opened 2015-06-18
CSRF protection when making requests from Javascript


MaxGabriel:
Yesod supports [CSRF protection when generating forms](http://www.yesodweb.com/book/forms), but as far as I'm aware, doesn't have native support for validating CSRF tokens when making requests from Javascript requests.

To improve this situation, I'd like to:

1. Add the CSRF token to the scaffolding's default layout in a way that is accessible by Javascript, potentially in a `<meta>` tag in the `<head>` of the HTML (This is what Ruby on Rails does):

    ```haskell
    mCsrfToken <- fmap reqToken getRequest -- in defaultLayout function
    ```
    ```
    $maybe token <- mCsrfToken
      <meta name="csrf-token" content=#{token}>
    ```
2. Add something to the scaffolding to add the CSRF token to a request made by Javascript (potentially using jQuery's [`ajaxPrefilter`](http://api.jquery.com/jquery.ajaxprefilter/) to add the token as a header). This would go along with the PR to add more JSON examples to the scaffolding (yesodweb/yesod-scaffold#76)
3. Add a function to Yesod to check the CSRF token in a standard way for Javascript requests. I'm not sure if there's a better solution here than requiring users to manually call a `checkCsrfToken` method. Potentially it could be combined with functions like `requireJsonBody`?

I think these changes would improve security for Yesod users, who might not be aware of CSRF attacks, and improve Yesod as an API server. Thoughts?

<hr>
Somewhat related: [SO answer](http://stackoverflow.com/a/28379873/1176156) in which I implement some bits of this functionality.

gregwebs:
I like the angular way of doing card protection better. https://docs.angularjs.org/api/ng/service/$http

MaxGabriel:
It looks like the Angular approach is to store the CSRF token in a cookie (`XSRF-TOKEN`) to send downstream, and include it as a header (`X-XSRF-TOKEN`) going upstream. This seems like it would simplify accessing the CSRF token from non-browser environments, because you wouldn't have to query the DOM for token (perhaps this would be good for tests?). 

If I understand correctly, there might be a security advantage to this approach because JS loaded from other domains wouldn't be able to read your domain's cookies (though this advantage is negated if you store CSRF tokens elsewhere in the page, like in server-generated forms).

A slight disadvantage is that it would increase request size for requests that don't need a CSRF token, like requests for images, but that's probably negligible.

I'm cool with that approach.

MaxGabriel:
[...]

PR #1017, opened 2015-06-23
Draft of adding CSRF protection for AJAX requests

MaxGabriel:
cf #1016

~~The code is missing tests and high-level documentation, but should be ready for review otherwise.~~

* Thoughts on what the default name should be for the CSRF cookie and header? I'm not super familiar with what other frameworks use (I see Angular uses "XSRF" and Rails uses "csrf"). Yesod uses "CSRF" in its functions instead of "XSRF", which is a strike against using "XSRF" for the naming scheme, even if it ends up being more consistent with tools like Angular.
* Is yesod-core the right home for these functions?

cc @gregwebs

### DIFF
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -153,6 +153,24 @@ module Yesod.Core.Handler
     , cached
     , cachedBy
     , stripHandlerT
+      -- * AJAX CSRF protection
+
+      -- $ajaxCSRFOverview
+
+      -- ** Setting CSRF Cookies
+    , setCsrfCookie
+    , setCsrfCookieWithCookie
+    , defaultCsrfCookieName
+      -- ** Looking up CSRF Headers
+    , checkCsrfHeaderNamed
+    , hasValidCsrfHeaderNamed
+    , defaultCsrfHeaderName
+      -- ** Looking up CSRF POST Parameters
+    , hasValidCsrfParamNamed
+    , checkCsrfParamNamed
+    , defaultCsrfParamName
+      -- ** Checking CSRF Headers or POST Parameters
+    , checkCsrfHeaderOrParam
     ) where
 
 import           Data.Time                     (UTCTime, addUTCTime,
@@ -185,6 +203,8 @@ import           Text.Hamlet                   (Html, HtmlUrl, hamlet)
 import qualified Data.ByteString               as S
 import qualified Data.ByteString.Lazy          as L
 import qualified Data.Map                      as Map
+
+import           Data.Byteable                 (constEqBytes)
 
 import           Control.Arrow                 ((***))
 import qualified Data.ByteString.Char8         as S8
@@ -219,6 +239,8 @@ import Data.Conduit (Source, transPipe, Flush (Flush), yield, Producer
                    )
 import qualified Yesod.Core.TypeCache as Cache
 import qualified Data.Word8 as W8
+import qualified Data.Foldable as Fold
+import Data.Default
 
 get :: MonadHandler m => m GHState
 get = liftHandlerT $ HandlerT $ I.readIORef . handlerState
@@ -479,10 +501,10 @@ setUltDestReferer = do
 redirectUltDest :: (RedirectUrl (HandlerSite m) url, MonadHandler m)
                 => url -- ^ default destination if nothing in session
                 -> m a
-redirectUltDest def = do
+redirectUltDest defaultDestination = do
     mdest <- lookupSession ultDestKey
     deleteSession ultDestKey
-    maybe (redirect def) redirect mdest
+    maybe (redirect defaultDestination) redirect mdest
 
 -- | Remove a previously set ultimate destination. See 'setUltDest'.
 clearUltDest :: MonadHandler m => m ()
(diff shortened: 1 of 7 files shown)


MaxGabriel:
Added a draft of the CSRF overview section. Need to verify the haddocks look sensible but am otherwise happy with that part.

gregwebs:
For CSRF vs XSRF, you are doing it right by picking one and making it configurable

MaxGabriel:
This current PR also requires manually calling `checkCsrfHeader` on requests that require it. Ideas on making that more automatic/seamless when accepting AJAX requests, the way that `runFormPost` handles CSRF tokens automatically? I was thinking of some sort of variant of `requireJsonBody` that also checked the CSRF header.

gregwebs:
Hmm, I would like to turn this into a middleware (at least a yesod middleware). That way enabling it would just mean dropping in a middleware.

MaxGabriel:
A Yesod middleware seems right for this, sounds good (Wai middleware is doable, but would it need a function passed in to extract the CSRF token from the request; not sure if that's worthwhile).

gregwebs:
You are welcome to use a Yesod middleware. If you went the wai route, you would probably take advantage of the vault parameter for request-local storage rather than passing in a function.

MaxGabriel:
Getting pretty close. Added Yesod Middleware and tests. The Yesod Middleware isn't well organized or documented yet, so working on cleaning up that.

MaxGabriel:
This should be ready for review. I'm happy with the explanation of CSRF attacks and the prevention mechanism. The middleware makes CSRF protection very simple to add. 

I'm not sure how this fits into yesod-form's CSRF protection yet; this middleware could handle that case if you used Javascript to add the CSRF header when a request was made from a form. If you didn't have Javascript enabled, that wouldn't work, though. Maybe on version of the middleware could skip CSRF protection for forms somehow? Anyway, I'm investigating this.

gregwebs:
forms should be sent as form-data. This CSRF protection would then only be on for JSON data (content type of application/json).

MaxGabriel:
Yeah, it looks like I can't just prevent CSRF protection from applying to `application/x-www-form-urlencoded ` and `multipart/form-data` because functions like `requireJsonBody` don't require an `application/json` content type, so those functions would be vulnerable to CSRF by sending a JSON body with a form content type. I could potentially change those functions to require that content type (though that's a breaking change).

I can have it only run CSRF protection on `application/json`, but it might end up biting people when they use javascript to submit a different content type and silently don't have CSRF protection. 

I can also just require users to add the CSRF header to their forms using e.g. jQuery. This makes it take extra work to support users with Javascript disabled, though. Maybe that's ok.

gregwebs:
> This makes it take extra work to support users with Javascript disabled, though. Maybe that's ok. 

Doesn't one have to use the existing Yesod forms that inject a hidden attribute to support disabled javascript?

MaxGabriel:
@gregwebs Yeah the current yesod-forms library works with disabled Javascript, but the middleware will reject the request before the handler can run the form to check the CSRF token.

snoyberg:
@gregwebs @MaxGabriel For the record, this is the prime example of a PR you guys should feel free to move ahead with without my input :)

MaxGabriel:
I'm not sure what I was thinking before, to handle the case where Javascript was disabled I just needed to check the form parameters for CSRF tokens as well, which I thought would be hard for some reason but was quite easy.

I made a few more modifications like using constant time comparison, as yesod-form suggests.

Anyway, I think this PR is ready for final review. I know you said to go ahead without your input @snoyberg but since it's a security thing I wouldn't mind more review; for a long time this PR wasn't doing things like constant time comparison.

snoyberg:
LGTM :+1:

yesod-core/Yesod/Core/Handler.hs
```diff
@@ -1258,3 +1280,113 @@ stripHandlerT (HandlerT f) getSub toMaster newRoute = HandlerT $ \hd -> do
...
+--
+-- Since 1.4.14
+checkCsrfHeaderOrParam :: MonadHandler m
+                       => CI S8.ByteString -- ^ The header name to lookup the CSRF token
+                       -> Text -- ^ The POST parameter name to lookup the CSRF token
+                       -> m ()
+checkCsrfHeaderOrParam headerName paramName = do
+  validHeader <- hasValidCsrfHeaderNamed headerName
+  validParam <- hasValidCsrfParamNamed paramName
+  unless (validHeader || validParam) (permissionDenied "Invalid CSRF token in header.")
```
gregwebs: The error message here only mentions one case.

yesod-core/test/YesodCoreTest/Csrf.hs
```diff
@@ -0,0 +1,78 @@
...
+        let body = "_token=" <> csrfValue
+        postRes <- srequest $ SRequest (defaultRequest { requestMethod = "POST", requestHeaders = [("Content-Type","application/x-www-form-urlencoded")] }) (fromStrict body)
+        assertStatus 200 postRes
+
+
+    it "403s write requests without the CSRF header" $ runner $ do
+        res <- request (defaultRequest { requestMethod = "POST" })
+        assertStatus 403 res
+    it "403s write requests with the wrong CSRF header" $ runner $ do
+        res <- request (defaultRequest { requestMethod = "POST", requestHeaders = [(defaultCsrfHeaderName, "foo")] })
```
gregwebs: it would be good to see this parallel the above test more closely (run `setCookieValue` and append "foo" to the value)

gregwebs:
looks fantastic! I only had 2 comments for making minor improvements. Looks like this also need to be rebased to the lastest master.

MaxGabriel:
Cool, I made changes based on @gregwebs's comments and rebased against master. I'll rebase into one commit and merge later tonight.

gregwebs:
After merging

- [ ] add to scaffold

MaxGabriel:
Cool, I'm adding it to this PR to the yesod-scaffolding https://github.com/yesodweb/yesod-scaffold/pull/76

Working on adding support to yesod-test for using CSRF tokens this way; mostly done with that.

MaxGabriel:
So based on some testing, it seems like `lookupPostParam` will consume the response body when JSON is sent to the server, so the response body is empty when the handler tries to parse JSON from it.

Haven't investigated much; initial ideas are to only use `lookupPostParam` if the header doesn't have the CSRF token, or to only use `lookupPostParam` on the content types forms use.

PR #1022, opened 2015-06-30
yesod-bin needs directory >= 1.2.1 due to System.Directory.findFiles

bergmark:
GHC 7.4 is no longer supported because of this (but also because of missing dep on ghc-prim and possibly other things)


snoyberg:
Thanks!

Issue #1026, opened 2015-06-30
Even better logging of SQL messages


andrewthad:
I have, on a previous occassion, opened a thread about [better logging of SQL](https://github.com/yesodweb/yesod/issues/939) in yesod. But, I just thought of a way to make it even better. Right now, the messages look like this:

    [Debug#SQL] SELECT "id", "vacation", "office365_user", "completed" FROM "office365_calendar_queue_job" WHERE ("completed"=?); [PersistBool False] @(<unknown>:<unknown> <unknown>:0:0)

Cool, sort of. In a suggestion I made to [ditch the source location](https://github.com/yesodweb/yesod/issues/939#issuecomment-75793415), I failed to realize that `logDebugN` uses the awful-looking `<unknown>` string and `0` line, which is annoying to see appended to the end of every line. What I want to see instead is this:

    [Debug#SQL] SELECT "id", "vacation", "office365_user", "completed" FROM "office365_calendar_queue_job" WHERE ("completed"=?); [PersistBool False]

Notice how there isn't any trailing garbage. My thought is that I can just change `defaultMessageLoggerSource` to target messages that are from source `"SQL"` and then **not** display the extra stuff at the end for those messages. I will end up needing to change `formatLogMessage` as well (but that should be fine because it isn't exported). Basically, what I want to ask is: Does this seem agreeable? If so, I can do it and put up a PR.

gregwebs:
That seems like an obvious improvement to me. But implementation-wise it would be better to not look at the log level but instead to look at whether or not there is actually location information.

andrewthad:
Unfortunately, `monadLoggerLog` expects a `Loc`, not a `Maybe Loc`, and `Loc` looks like this:

    data Loc
      = Loc { loc_filename :: String
      , loc_package  :: String
      , loc_module   :: String
      , loc_start    :: CharPos
      , loc_end      :: CharPos }

So, there is no where to concretely say "This information does not exist". I wish that this situation were different, but changing anything there would definitely not be worth it. So, I'll interpret your suggestion to mean "check to see if `loc_filename` is `"<unknown>"`, which could be a better approach. Seems a little hacky though.

andrewthad:
@gregwebs Also, just to clarify, my suggestion is to examine `LogSource` to make the decision, not `LogLevel`:

    messageLoggerSource :: site
                        -> Logger
                        -> Loc -- ^ position in source code
                        -> LogSource
                        -> LogLevel
[...]

PR #1027, opened 2015-06-30
Don't show source location for logs that don't have that information

andrewthad:
https://github.com/yesodweb/yesod/issues/1026

Don't merge this yet. I haven't tested it. Two questions:

  - There's a line in a haddock comment I added where I try to escape some special characters. Did I do this right?
  - Is there any reason we can't use `<>` instead of `mappend`? I'd much rather switch to that.

### DIFF
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -633,10 +633,6 @@ formatLogMessage :: IO ZonedDate
                  -> IO LogStr
 formatLogMessage getdate loc src level msg = do
     now <- getdate
-    let sourceSuffix = if loc_package loc == "<unknown>" then "" else mempty
-          `mappend` " @(" 
-          `mappend` toLogStr (fileLocationToString loc) 
-          `mappend` ")"
     return $ mempty
         `mappend` toLogStr now 
         `mappend` " [" 
@@ -650,6 +646,11 @@ formatLogMessage getdate loc src level msg = do
         `mappend` msg 
         `mappend` sourceSuffix
         `mappend` "\n"
+    where 
+    sourceSuffix = if loc_package loc == "<unknown>" then "" else mempty
+        `mappend` " @(" 
+        `mappend` toLogStr (fileLocationToString loc) 
+        `mappend` ")"
 
 -- | Customize the cookies used by the session backend.  You may
 -- use this function on your definition of 'makeSessionBackend'.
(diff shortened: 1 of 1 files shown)


gregwebs:
* I am horrible at haddocks :)
* `<>` was not introduced until GHC 7.4. Not sure if we know whether < 7.4 is broken yet.

Also, I would use a `where` binding instead of a `let` for `sourceSuffix`. That way it is clear it is only dependent on the function inputs.

creichert:
- You can `cabal haddock` (or stack haddock?) locally and open that specific file if you want to see it. It looks good to me, though.

- It would be nice if the CI coverage could definitively tell us whether something like `<>` will cause breakage on supported GHC versions. I can't remember what the conclusion was the last time this was discussed.

andrewthad:
Looks like `stack haddock` builds docs for all the dependencies too. Super cool. It looks like it ended up formatted correctly.

I'll leave `mappend`, just in case some poor soul is building yesod on GHC 7.2.

@gregwebs I like your suggestion about using a `where` binding. I've changed it to use that instead.

andrewthad:
So, I'm trying to test this, and it builds correctly with stack, but I want to make sure that it has the expected runtime behavior. I created a minimal yesod+sqlite single-file app named `example.hs`, but I can't figure out how to run it:

    # Neither of these work
    stack ghc example.hs
    stack exec runhaskell example.hs

Everything I try gives:

    example.hs:11:8:
        Could not find module ‘Yesod’
        Use -v to see a list of the files searched for.

Does anyone have any suggestions for using a library you build with stack?

snoyberg:
You probably are using a different snapshot. Try running stack build yesod first and see if that changes the result.

snoyberg:
And FYI, there's also stack runghc.

andrewthad:
Thanks Michael. I had forgotten that `yesod-core` doesn't export `Yesod` (the module), so I basically needed to build more.

Something weird is happening. The changes that I've introduced strip off the piece at the end for anything that uses `logDebugN`, etc. **However**, this does **not** fix the SQL logs. I put some additional debugging logic around it and discovered that yesod's `messageLoggerSource` isn't even called for SQL logs. So, does anyone know where they are getting formatted or why they are immune from the normal logging process?

andrewthad:
Wait a moment. I think I know why.

andrewthad:
Nevermind, everything is good, and this should be ready to be merged in. I was building one of the example apps from the yesod book. It uses [runStderrLoggingT](https://hackage.haskell.org/package/monad-logger-0.3.13.1/docs/Control-Monad-Logger.html#v:runStderrLoggingT) for the SQL logs, which cuts yesod out of the equation and uses monad-logger's `defaultLogStr` instead, which produces very similar (maybe even the same) default logs as yesod does. Maybe that one should be changed too, but I'll leave that for another day.

andrewthad:
This is ready to be merged in. It has the expected behavior.

geraldus:
Cool stuff! That `<unknown>` things are very confusing to me.

Issue #686, opened 2014-03-11
Add i18n placeholder support to yesod-form


meteficha:
Currently you may add a `placeholder` attribute using the `fsAttrs` field of `FieldSettings`.  However, that isn't a good solution since you either have to forego i18n for your placeholder or manually use `getMessageRender`.

An option is to make `placeholder` a first-class citizen by creating a `fsPlaceholder` attribute.  I'm not sure how much code would be broken by this change.

snoyberg:
I think we'll want to rethink `FieldSettings` a bit for the next breaking change, may as well include this at that time.

meteficha:
I agree.  And I think that promoting it to a separate field is the way to go.

geraldus:
This will be very useful!

StevenXL:
Please see #1597 and #1593. Both attempt to add i18n placeholder support, but were rejected. We are closing down this issue, but as always, we can re-open if need be.

geraldus:
> I think we'll want to rethink `FieldSettings` a bit for the next breaking change, may as well include this at that time.

Sorry for off-topic, very interested in Yesod's future, is there any place where you discuss / inform about pending changes and etc.?  I'm interested in participation

snoyberg:
The Yesod mailing list would be the best place for that. I've been avoiding
breaking changes for a few years now, which is why such discussions haven't
happened much. But please do feel free to bring up such discussions there.

PR #1029, opened 2015-07-07
Hidden fields in bootstrap3 layouts and autocomplete with completion values stored in hidden field

myShoggoth:
I wanted to make an autocomplete form input that would potentially match against existing db entries, and for those cases where the entry already existed would fill in a hidden input with the ID number. I've done this in RoR and GWT before. JQuery has support for it, so I gave it a whirl and...
1. Hidden inputs were not handled well with bootstrap3 layout, it tried to put them in the form groups and that made the layout go screwy
2. The existing jquery autocomplete didn't have support for this case

Here's some code snippets from the project that uses this:

```
investmentForm :: Maybe Investment -> Maybe Investor -> Form (Text, Maybe InvestorId, Maybe Double, Maybe Int)
investmentForm investment investor = do
    let ident = "investor"
    renderBootstrap3 (BootstrapHorizontalForm (ColMd 2) (ColMd 3) (ColMd 0) (ColMd 4)) $ (,,,)
      <$> areq ((jqueryAutocompleteField'' 2 True) InvestorAutocompleteR) (withPlaceholder "Investor's Name" (withAutofocus (jqfs MsgInvestorName ident False True))) (investorName <$> investor)
      <*> aopt hiddenField (jqfs MsgInvestorName ident True True) (investmentInvestorId <$> investment)
      <*> aopt doubleField (bfs MsgAmount) (Just (investmentAmount <$> investment))
      <*> aopt intField (bfs MsgShares) (Just (investmentShares <$> investment))
```

and the InvestorAutocompleteR:

```
getInvestorAutocompleteR :: Handler Value
getInvestorAutocompleteR = runDB $ do
    text <- do
    ((res, widget), enctype) <- runFormGet $ renderDivs $ areq textField "Text"  Nothing
        case res of
            FormSuccess (rawtext) -> return rawtext
            _                     -> return ""
    let like field val = Filter field (Left $ T.concat ["%", val, "%"]) (BackendSpecificFilter "like")
    emails <- selectList [EmailAddress `like` text] [LimitTo 10]
    investors <- selectList [InvestorName `like` text] [LimitTo 10]
    investorsEmails <- selectList [InvestorId <-. (map (fromJust . emailUserId) (map entityVal emails))] []
    return $ Array $ fromList $ jsonInvestorAutocompletes (investorsEmails ++ investors)
```


andrewthad:
This makes a backwards-incompatible change with `FieldSettings`, so we would definitely need a major version bump for this. An alternative approach would be doing a hack similar to the way we currently target the submit button for the bootstrap layout.

Just to make a note of a thought I've been having, at some point in the future, I feel like it would be nice to have a bootstrap-specific form library has options more focused on bootstrap and could handle these kinds of cases better.

meteficha:
Note that you can use `renderBootstrap3` to get an `MForm` then add the hidden input yourself.  It sure is painful, though.

myShoggoth:
Yeah, and I'm using a jquery control in a bootstrap form, which happens all the time in the real world but is not handled in the current design for forms well.

I think a new model for forms in Yesod is due, but I haven't yet figured out what it might be.

Instead of calling the new FieldSetting field "isHidden" it could more generally called something like "no custom processing", because that's all that is really happening here.

creichert:
What is the consensus on merging this one? I think the biggest blocker are the breaking changes. We do have several other breaking changes being prepared for a new release already in other areas of yesod.

If this is a favorable change than I think it should be accompanied by an example and documentation. I am also not in favor of the `jqfs` name or the jqueryFieldSelector'' export, necessarily.

I do think this looks like an improvement overall, though.

meteficha:
It should be bundled up with #686, which is going to change the same data type. Perhaps it should be merged to a "next-major-version" branch?

PR #1030, opened 2015-07-09
yesod keter support for stack

chreekat:
Based on Snoyman's quick fix for yesod devel.

### DIFF
--- a/yesod-bin/Keter.hs
+++ b/yesod-bin/Keter.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 module Keter
     ( keter
+    , KeterOpts(..)
     ) where
 
 import Data.Yaml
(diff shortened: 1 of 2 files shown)


snoyberg:
In this case, why not simply call out to `stack build` instead?

chreekat:
Overly paranoid about the location of the binary changing on me, mostly. `dist/build/Snowdrift/Snowdrift` seems reliable, but I wasn't sure about `.stack-work/dist/x86_64-linux/Cabal-1.18.1.5/build/Snowdrift/Snowdrift`. Maybe there's a better way?

snoyberg:
`stack exec which Snowdrift` or `stack path --dist-dir`

chreekat:
That information has to get into the yaml somehow. In fact, my first idea was to bake `stack path` into yesod-bin such that I could specify apps in keter.yaml without paths at all.

```
- type: webapp
  exec: Snowdrift
```

I decided that would be more invasive. Now that I've written it out, though, it doesn't seem too crazy. Would still need to differentiate from non-built executables, e.g. a bash script I have running as a background task.

snoyberg:
Oh, that's a good point. You could use the newly added `--local-bin-path` argument though to `stack install` to get the executables placed somewhere specific. That would require waiting for the next stack release (probably on Sunday). Your call, if you want this now, I don't mind merging this as a stop-gap.

chreekat:
No one else on our team needs this, so I'm fine waiting. :)

snoyberg:
@chreekat The `--local-bin-path` option is in stack and will be released in the next day or two. Are you OK with closing this and reopening a new PR that uses that technique instead?

chreekat:
Yeah, I'll do that next week.

PR #1032, opened 2015-07-15
Fix compilation error in yesod-websockets/sample.hs.

creichert:
Add OverloadedStrings language extension.

I'm going to go ahead and merge this since it's a trivial change in the example.


snoyberg:
As usual, thanks!

PR #1035, opened 2015-07-19
Update WebSockets.hs

tolysz:
Add `send and  receive` `DataMessageE` so the whole WS API is mirrored


tolysz:
thanks, now I can nuke my `yesod-websocket-extra` package :)

Issue #1037, opened 2015-07-21
`IsString` instance for `WidgetT site IO ()`


andrewthad:
I don't know if this has been discussed before, but what do people think about having an `IsString` instance for `WidgetT site IO ()`? Every `String` can trivially be promoted to a widget via:

    toWidget . toHtml . pack

This instance would be helpful to me in certain situations (and while I'm at it, an `IsString` instance for `MarkupM ()` but that isn't part of yesod). Does anyone else think this is a good idea, or do you guys see this kind of thing as harmful/confusing?

snoyberg:
I don't see a problem

PR #1038, opened 2015-07-21
Add IsString instance for WidgetT site m ()

andrewthad:
https://github.com/yesodweb/yesod/issues/1037

### DIFF
--- a/yesod-core/Yesod/Core/Types.hs
+++ b/yesod-core/Yesod/Core/Types.hs
@@ -250,7 +250,7 @@ instance (a ~ (), Monad m) => Monoid (WidgetT site m a) where
     mempty = return ()
     mappend x y = x >> y
 instance (a ~ (), Monad m) => Semigroup (WidgetT site m a)
-instance Monad m => IsString (WidgetT site m ()) where
+instance (Monad m, a ~ ()) => IsString (WidgetT site m a) where
     fromString = toWidget . toHtml . T.pack
       where toWidget x = WidgetT $ const $ return $ ((), GWData (Body (const x)) 
                          mempty mempty mempty mempty mempty mempty)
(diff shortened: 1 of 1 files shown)


yesod-core/Yesod/Core/Types.hs
```diff
@@ -250,6 +250,10 @@ instance (a ~ (), Monad m) => Monoid (WidgetT site m a) where
     mempty = return ()
     mappend x y = x >> y
 instance (a ~ (), Monad m) => Semigroup (WidgetT site m a)
+instance Monad m => IsString (WidgetT site m ()) where
```
snoyberg: I think you'll get better type inference with:

```haskell
instance (Monad m, a ~ ()) => IsString (WidgetT site m a)
```

andrewthad:
I'll wait two days to see if there is any opposition to this, and then I'll merge the branch.

yesod-core/Yesod/Core/Types.hs
andrewthad: Ah! That dumb trick I always forget.

snoyberg: I only remembered because the previous instance uses it

creichert:
@andrewthad Do you mind if I push a doc string for the new instance?

andrewthad:
@creichert That would be wonderful.

andrewthad:
@creichert I think you might have typoed the docstring a little. It reads:

> Any type with an 'IsString' instance can be trivially promoted to a widget.

I think it was supposed to be:

> An 'String' can be trivially promoted to a widget.

Or something like that. Because there are lots of types (`ByteString` for example) that have `IsString` instances but can't be trivially promoted to a widget.

creichert:
Yea, my mistake, thanks. I pushed a fix.

PR #1041, opened 2015-07-23
Enable stack with yesod keter

chreekat:
Cabal is used as usual unless stack.yaml exists.

Executables are placed in ./dist/bin, a location of my own creation. Maybe
./dist-stack would be better, though? Or worse?


snoyberg:
LGTM, let's merge once Travis passes.

PR #1044, opened 2015-07-25
Change done.cg to refer to 'stack build'

hritchie:
Remove instruction to use cabal install.
Instead direct users to use 'stack build', as is suggested in the quickstart
guide.


gregwebs:
should it be:  `cd PROJECTNAME && stack build && stack exec -- yesod devel` ?

hritchie:
@gregwebs That's a nice improvement. I have made the change as you suggested

gregwebs:
Thanks!

PR #1050, opened 2015-08-02
Fix parameter order for uglifyjs.

mpietrzak:
UglifyJS2 requires filenames (or "-") first, then options. Otherwise it just prints "Error parsing arguments in: -". Tested with uglify-js 2.4.24.


gregwebs:
Thanks!

PR #1053, opened 2015-08-06
Test middleware

gregwebs:
I want to see a request logger of the tests. Allowing me to configure my middleware for tests accomplishes this. I didn't see how to make this change in a non-breaking way.

The type alias is questionable, but may be helpful.

We can hold off on merging for a while since it is a breaking change


MaxGabriel:
#919 also has a breaking change for yesod-test for whenever this is merged.

gregwebs:
I am going to go ahead and make the breaking release with these 2 PRs after first releasing master

Issue #1051, opened 2015-08-04
mkYesod and HandlerSite with polymorphic type parameters


Daniel-Diaz:
Currently, it seems impossible to use ``myYesod`` with a type with at least one polymorphic type parameter. Something like this:

```haskell
data Foo a = Foo a

mkYesod "Foo" $(parseRoutesFile "foo-routes-file")
```

is not allowed.

```
   The first argument of ‘Yesod.Core.Types.HandlerT’
      should have kind ‘*’,
      but ‘Foo’ has kind ‘* -> *’
```

This means we are currently restricted to kind ``*``.

Is there anyway I can use a type with polymorphic variables? If not, how feasible is to add this feature? Easy, hard, time consuming?

Thanks.

mgsloan:
Looks to me like this would be easy to add.  In `mkYesodGeneral`, [here](https://github.com/yesodweb/yesod/blob/b5077abdd988e267a1cdc7dcaff43674a64d1755/yesod-core/Yesod/Core/Internal/TH.hs#L67), you'd use `lookupTypeName` and `reify` to get info about the datatype.  (it's minorly unfortunate that the `mkYesod` functions don't take a `Name`..)

You'd then need to write an `Info -> Maybe Int` function that gets the arity of the datatype.  You'd then create new polymorphic variable names, and supply them in the`site`  type constructed  [here](https://github.com/yesodweb/yesod/blob/b5077abdd988e267a1cdc7dcaff43674a64d1755/yesod-core/Yesod/Core/Internal/TH.hs#L86).

gregwebs:
We could create a `makeYesod` version that takes a `Name` and eventually deprecate the existing form.

gregwebs:
Why do you need to parameterize `Foo`?

andrewthad:
@gregwebs That's was I was wanting to know.

MaxGabriel:
I believe this is closed by #1055

PR #1055, opened 2015-08-11
mkYesodGeneral: Argument types can now be polymorphic

Daniel-Diaz:
PR related with #1051.

## Summary

I made some changes to ``mkYesodGeneral``, so that it is possible to create sites with polymorphic type parameters.

## Type change

The type of ``mkYesodGeneral`` has changed from ``String -> [String] -> Bool -> [ResourceTree String] -> Q ([Dec],[Dec])`` to ``String -> [Either String [String]] -> Bool -> [ResourceTree String] -> Q ([Dec],[Dec])``. The ``[String]`` argument was used to pass monomorphic type arguments to the type named by the first argument. In the new function, you can either specify a monomorphic type (using ``Left``) or a polymorphic type (using ``Right``). The list of ``String``s represents a list of classes that will be added to the context of the ``YesodDispatch`` instance.

I know this might seem confusing, but I thought it was the closest way to achieve what I wanted without making substantial changes. Plus, everything that worked before, should still work without changes.

## Example

An example to explain how the new mkYesodGeneral works:

```haskell
data Foo a = Foo a

class Bar a where
  ...

mkYesodWith "Foo" [Right ["Bar"]] (...)
```

This would create Yesod types and instances for the type ``Foo``, and the ``YesodDispatch``class would look like:

```haskell
instance Bar a => YesodDispatch (Foo a) where
   ...
```

The ``Handler`` will be parametrized too:

```haskell
getSomethingR :: Handler a x
getSomethingR = ...
```

If you write a route handler like:

```haskell
getSomethingElseR :: Bar a => Handler a x
getSomethingElseR = ...
```

Since the ``Bar a`` constraint has been added to the ``YesodDispatch`` instance context, it won't fail because a missing instance.

Note that the type variable of polymorphic parameters is not given by the user, but automatically generated by ``mkYesodGeneral``.

## Other considerations

* The ``Either`` type is a little obscure, in the sense that it doesn't really tell us much about the intention of the argument. In the first place I considered defining a new type to make the intention of the argument clearer. I failed to the ``Either`` side because I wanted to make as little changes as possible.

* New ``mkYesodGeneral`` uses a very simple pattern match to deduce the arity of the site type. This has flaws. For example, it doesn't see through type synonyms, which will consider of arity 0. In any case, this means that for many cases it will just work as it has been working before.

* A new function ``mkYesodWith`` was added in the spirit of ``mkYesod``, but with an additional argument to pass type parameters (either monomorphic or polymorphic). The name was chosen arbitrarily and can be changed to any other.

* In issue #1051 was suggested to make ``mkYesodGeneral`` arguments be of type ``Name`` instead of ``String``, which probably makes sense, but I didn't want to do it just yet.

## Epilogue

Please, let me know any changes that are needed to get this functionality merged. I need it for a server I'm writing where the site type is parametrized. This shouldn't affect any yesod users as ``mkYesod`` will behave exactly the same way for traditional use cases. This change will only make possible to compile more cases.

Thanks!

### DIFF
--- a/yesod-core/Yesod/Core/Dispatch.hs
+++ b/yesod-core/Yesod/Core/Dispatch.hs
@@ -11,6 +11,7 @@ module Yesod.Core.Dispatch
     , parseRoutesFile
     , parseRoutesFileNoCheck
     , mkYesod
+    , mkYesodWith
       -- ** More fine-grained
     , mkYesodData
     , mkYesodSubData
--- a/yesod-core/Yesod/Core/Internal/TH.hs
+++ b/yesod-core/Yesod/Core/Internal/TH.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE CPP #-}
 module Yesod.Core.Internal.TH where
 
 import Prelude hiding (exp)
@@ -94,7 +95,13 @@ mkYesodGeneral name args isSub resS = do
                    case arg of
                      Left  t  -> ( ConT (mkName t):xs, n:ns, cs )
                      Right ts -> ( VarT n         :xs,   ns
-                                 , fmap (\t -> AppT (ConT $ mkName t) (VarT n)) ts ++ cs )
+                                 , fmap (\t -> 
+#if MIN_VERSION_base(4,8,0)
+                                               AppT (ConT $ mkName t) (VarT n)
+#else
+                                               ClassP (mkName t) [VarT n]
+#endif
+                                          ) ts ++ cs )
                  ) ([],vns,[]) args
         site = foldl' AppT (ConT $ mkName name) argtypes
         res = map (fmap parseType) resS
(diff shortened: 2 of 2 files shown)


creichert:
Can you outline a specific use-case for this? (I'm not doubting it's useful but I am curious). 

Assuming there are no other issues, it would also be nice to have some form of a test case for the exposed function (something similar to the WaiSubsiteTest).

creichert:
Direct link to build issue on ghc 7.8: https://travis-ci.org/yesodweb/yesod/jobs/75128065#L1367

Daniel-Diaz:
@creichert My current use-case involves a shared state (implemented as a mutable reference in the site type) with a Map, where the keys (maybe elements too) of the map are different in different instances of the server. The types are decided from a configuration file at run time.

The build failure seems to be related with the GHC version used by travis. I'll see how to make it compatible with version GHC previous to 7.10.

Daniel-Diaz:
Compilation problems solved.

gregwebs:
This is good, but it looks like a breaking change. Is it possible to export a new function (perhaps YesodPolymorphic) and maintain the old interface?

Daniel-Diaz:
@gregwebs Please, can you tell me how this is a breaking change? As far as I know, ``mkYesodGeneral`` is not exported to the user.

Thanks!

gregwebs:
ahh, ok

PR #1058, opened 2015-08-20
Add functions to yesod-test to support the new CSRF middleware

MaxGabriel:
This PR adds functions to add CSRF tokens from cookies, supporting the new CSRF middleware added in #1017.

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -77,6 +77,8 @@ module Yesod.Test
     , addToken_
     , addNonce
     , addNonce_
+    , addTokenFromCookie
+    , addTokenFromCookieNamedToHeaderNamed
 
     -- * Assertions
     , assertEqual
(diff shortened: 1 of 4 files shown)


MaxGabriel:
@snoyberg or @gregwebs Can you upload yesod-core-1.4.14 to Hackage so that yesod-test can depend on it?

yesod-test/Yesod/Test.hs
```diff
@@ -93,6 +95,7 @@ module Yesod.Test
     -- * Grab information
     , getTestYesod
     , getResponse
+    , getRequestCookies
```
MaxGabriel: I don't know the best spot for this function, either in the export list or in the source code.

snoyberg:
Uploaded

yesod-test/Yesod/Test.hs
gregwebs: looks fine to me

gregwebs:
Thanks for handling this issue from every angle :) Testing is often forgotten about

PR #1064, opened 2015-08-23
yesod-test: have `request` not default to the form-urlencoded Content-Type

### LINKED ISSUE #1063: Can't use Content-Type application/json from yesod-test
Afaict, yesod-test currently adds a form content type ("application/x-www-form-urlencoded" or "multipart/form-data") to all requests, which is undesirable if you want to use a different content-type. I wrote [a test demonstrating this issue](https://github.com/yesodweb/yesod/compare/yesodweb:master...MaxGabriel:contentTypeJson?expand=1).

I think the root cause of this in the code is that the `request` method [will always use either the `singlepart` or `multipart` functions](https://github.com/yesodweb/yesod/blob/ad7326aaaaafb83efdfa4f6ed475fba7de5c5c61/yesod-test/Yesod/Test.hs#L796-L803), and those add the [urlencoded](https://github.com/yesodweb/yesod/blob/ad7326aaaaafb83efdfa4f6ed475fba7de5c5c61/yesod-test/Yesod/Test.hs#L880) and [multipart](https://github.com/yesodweb/yesod/blob/ad7326aaaaafb83efdfa4f6ed475fba7de5c5c61/yesod-test/Yesod/Test.hs#L853) headers, respectively.

This seems to cause issues with the new CSRF middleware, because when you try to send JSON to the server with the CSRF header, the server sees the content-type as URL-encoded, and tries to parse POST parameters from that. I think this is consuming the response body, so it's empty by the time you try to parse JSON from it in a Handler.

I haven't investigated that much, but potentially a good solution would be not automatically adding a form content-type header if [`BinaryPostData` is used](https://github.com/yesodweb/yesod/blob/ad7326aaaaafb83efdfa4f6ed475fba7de5c5c61/yesod-test/Yesod/Test.hs#L801).  (Edit: This fix seems to work, except that the [default request](https://github.com/yesodweb/yesod/blob/ad7326aaaaafb83efdfa4f6ed475fba7de5c5c61/yesod-test/Yesod/Test.hs#L779) uses `MultipleItemsPostData`)

MaxGabriel:
* Only set the Content-Type to `application/x-www-form-urlencoded` if key-value pairs are added
* Previously the `application/x-www-form-urlencoded` Content-Type would be added even if you set a binary request body.
	* You could add your own Content-Type with `addRequestHeader`, but this resulted in multiple Content-Type headers.
* Closes #1063

### DIFF
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -1,3 +1,9 @@
+## 1.5.0.1
+
+* Fixed the `application/x-www-form-urlencoded` header being added to all requests, even those sending a binary POST body [#1064](https://github.com/yesodweb/yesod/pull/1064/files)
+	* The `application/x-www-form-urlencoded` Content-Type header is now only added if key-value POST parameters are added
+	* If no key-values pairs are added, or the request body is set with `setRequestBody`, no default Content-Type header is set
+
 ## 1.5
 
 * remove deprecated addNonce functions
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -876,8 +876,7 @@ request reqBuilder = do
       SRequest simpleRequest' (simpleRequestBody' rbdPostData)
       where
         simpleRequest' = (mkRequest
-                          [ ("Cookie", cookieValue)
-                          , ("Content-Type", "application/x-www-form-urlencoded")]
+                          ([ ("Cookie", cookieValue) ] ++ headersForPostData rbdPostData)
                           method extraHeaders urlPath urlQuery)
         simpleRequestBody' (MultipleItemsPostData x) =
           BSL8.fromChunks $ return $ TE.encodeUtf8 $ T.intercalate "&"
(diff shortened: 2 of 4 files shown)


gregwebs:
looks good to me. I have a PR to upgrade the scaffold to yesod-test-1.5, so we probably don't need to backport this to yesod-test-1.4.x
https://github.com/yesodweb/yesod-scaffold/pull/93

MaxGabriel:
Cool, thanks @gregwebs. I don't think this is necessary to backport either; it's been working well for everyone so far (the main reason I wanted to patch this was for the CSRF code).

This is technically a breaking change, but I think it would only affect people who are submitting forms with no parameters or CSRF tokens, which is probably a pretty weird/rare case.

MaxGabriel:
I conclude to not treat this as a breaking change: this change is unlikely to effect anyone, it came right immediately after a breaking change (so people upgrading from 1.4.x to 1.5.x already have to be prepared for breaking changes), it only applies to test suites so it wouldn't break any production code, and it's fixing a bug.

Added to the Changelog entry as 1.5.0.1. Greg's already reviewed so I will merge shortly.

Issue #1059, opened 2015-08-21
yesod-core update seems to have broken git-annex


mdorman:
Even though it's git-annex that fails to build, I'm raising this issue here because it seems unlikely to be a truly git-annex specific issue (though [I have also logged it in git-annex's bug tracker](http://git-annex.branchable.com/bugs/Build_failures_with_7.10/)) 

Anyway, the Nix build job for git-annex started failing yesterday, with the following error ([full build log](http://hydra.cryp.to/build/1099681/nixlog/1/raw)):

    Assistant/WebApp/Types.hs:39:1: ‘WebApp’ is not in scope at a reify

This failure coincided with pulling yesod-1.4.14 (41f217351e9e81a618ce4de9abb1e2efa56ae9ed being that commit) into nix.

As I said, I realize it may actually be a git-annex at fault, but it went from building to broken as a consequence of the yesod upgrade, so it seems like at least something you might want to warn your users to check regarding their apps.

creichert:
This could coincide with 9991e307e3ba3caa83d44fdfcc705e2b679676d8 - ea62a38464eb4c22fa4d100f387bd22720d72b67 which seem to have been merged right before the release. 

For reference: https://github.com/joeyh/git-annex/blob/master/Assistant/WebApp/Types.hs#L39

gregwebs:
@Daniel-Diaz can you take a look at this?

Daniel-Diaz:
@gregwebs I'm looking into it. It really seems related to my changes.

Daniel-Diaz:
If you place the ``data`` definition _above_ the ``mkYesodData`` TH call, the mentioned code compiles. In order to use a type in TH you have to define it above. Otherwise, the type is not in scope at the time of generating the splice.

gregwebs:
So did the changes require the data definition to be placed above now? Was it a GHC accident that it worked before?

Daniel-Diaz:
It's because of the use of ``reify``, which is mentioned in the error. ``mkYesod`` (and variants) didn't use ``reify`` before. This is needed for the new feature - accepting types with parameters as yesod site types. Function ``reify`` extracts information about a type that must be already in scope at splicing time.

Daniel-Diaz:
Sorry, I didn't think about this breaking change when I wrote the patch, and instinctively defined the type before using it in all my tests.

gregwebs:
[...]

PR #1065, opened 2015-08-23
mkYesodGeneral: Assume type arity 0 when type is not in scope at splicing time

Daniel-Diaz:
This avoids calling `reify` when the reified type is not in scope at splicing time (TH call). Since we still need to specify the arity of the type, we assume it to be 0, producing code equivalent to the code produced by previous versions of yesod (before 1.4.14).

I'm not sure this is a step forward, so please, before merging, read my comment https://github.com/yesodweb/yesod/issues/1059#issuecomment-133948435.


Daniel-Diaz:
By the way, I placed in one of the tests the type definition after the call to `mkYesod` to make sure this is working.

gregwebs:
Thanks! This closes #1059

PR #1067, opened 2015-08-28
yesod-websockets: accept websocket requests with a subprotocol

wuzzeb:
The websockets library allows passing an AcceptRequest when
upgrading the request to websockets, which at the moment only
contains the subprotocol to use with the client.  The list of
subprotocols that the client is prepared to speak is in the
RequestHead.


creichert:
Looks good to me, I'll merge when travis gives the OK

creichert:
Awesome, thanks!

PR #1073, opened 2015-09-10
yesod-auth: google sign-in ask for profile permission

gregwebs:
somehow we are able to read the profile
of most users without this


creichert:
Thanks! I'm assuming that's actually 2 scopes with a space in between? (I had to look it up)

gregwebs:
Yep

PR #1078, opened 2015-09-28
Create CODE_OF_CONDUCT.md

maxnordlund:
This is just a suggesting from http://contributor-covenant.org/ but I think it's a good a start as any to begin a discussion of how it should look. However I think it is to important just to leave it hanging.

Thoughts and ideas are very much welcome, whatever they may be, but try to be polite :bow:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,41 +1,8 @@
 # Contributor Code of Conduct
 
-As contributors and maintainers of this project, and in the interest of
-fostering an open and welcoming community, we pledge to respect all people who
-contribute through reporting issues, posting feature requests, updating
-documentation, submitting pull requests or patches, and other activities.
+Always be nice. When communicating online treat people the way you would if
+they were standing next to you. The project maintainers have the right and
+responsibility to maintain that standard of niceness, and you should contact
+them if it is not upheld.  Don't forget to be nice whenever representing the
+project to others outside the project.  If you are not nice, apologize. 
 
-We are committed to making participation in this project a harassment-free
-experience for everyone, regardless of level of experience, gender, gender
-identity and expression, sexual orientation, disability, personal appearance,
-body size, race, ethnicity, age, religion, or nationality.
-
-Examples of unacceptable behavior by participants include:
-
-* The use of sexualized language or imagery
-* Personal attacks
-* Trolling or insulting/derogatory comments
-* Public or private harassment
-* Publishing other's private information, such as physical or electronic
-  addresses, without explicit permission
-* Other unethical or unprofessional conduct.
-
-Project maintainers have the right and responsibility to remove, edit, or
-reject comments, commits, code, wiki edits, issues, and other contributions
-that are not aligned to this Code of Conduct. By adopting this Code of Conduct,
-project maintainers commit themselves to fairly and consistently applying these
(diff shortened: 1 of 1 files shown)


gregwebs:
hmm, I haven't seen issues with people violating this code of conduct so I am not sure if this is necessary. I usually like making the implicit explicit but I hesitate in this case because I feel that the code could be better expressed in a few sentences in a clearer way with more useful information and less negativity.

```
Always be nice. When communicating online treat people the way you would if they were standing next to you. The project maintainers have the right and responsibility to maintain that standard of niceness, and you should contact them if it is not upheld.
Don't forget to be nice whenever representing the project to others outside the project.
If you are not nice, apologize. 
```

What do you think?

erikd:
Personally I thought this was un-necessary  until I read http://sarah.thesharps.us/2015/10/05/closing-a-door/ .

Now I think we should have one and I am very much in favour of something simple and to the point
like @gregwebs suggested.

maxnordlund:
I've updated it to the suggestion from @gregwebs. But it might need some more polish. Before merge I'll squash them together, but for comparison they're separate commits.

gregwebs:
Let me know how it can be polished without talking down to the readers

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,8 @@
+# Contributor Code of Conduct
+
+Always be nice. When communicating online treat people the way you would if
+they were standing next to you. The project maintainers have the right and
+responsibility to maintain that standard of niceness, and you should contact
+them if it is not upheld.  Don't forget to be nice whenever representing the
```
maxnordlund: Who would I contact if the maintainer is the culprit?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,8 @@
...
+project to others outside the project.  If you are not nice, apologize.
```
maxnordlund: Where should the person apologize? In public, or does in private count? How would you verify that? Should it be verified? By whom?

maxnordlund:
Maybe add something about inclusion of all people, regardless of background/gender/etc. But it might be bad to keep a huge list, but it might be good as well.

I don't quite know how to tweak it, just more questions. But the more eyes the better, I hope.

On the other hand, this might be good start and leave the polishing for later.

gregwebs:
@maxnordlund I would like to point out that the original code didn't specify specific answers either. We could add a sentence to the end like the following:

```
If someone is not nice to you, tell them in a nice way or tell a project maintainer: we care about fostering a welcoming community.
```

@erikd The Linux kernel is very different from Yesod. I could be wrong, but my impression is that everything going on within github and the mail list is polite. The problems I have seen within the last year have probably only been on Reddit.

So for the most part I would rather stay "agile" and deal with problems in a specific way when they arise then try to spell out lots of details here.

erikd:
@gregwebs Yes, I agree, the Haskell comunity in general and the Yesod community in particular has very, very few problems of this kind (at least to my knowledge). I'm sure we'd all like to keep it that way.

gregwebs:
merged and made [one small change](https://github.com/yesodweb/yesod/commit/bb7f0afe3415ddee184bf6e4dc3ecb8420de932d)

Thanks!

maxnordlund:
No problem :)

PR #1079, opened 2015-09-30
added support for feed logos and item enclosures

nek0:
I hope I didn't mess things up too badly ...

### DIFF
--- a/yesod-newsfeed/Yesod/Common.hs
+++ b/yesod-newsfeed/Yesod/Common.hs
@@ -1,9 +1,0 @@
-module Yesod.Common
-  ( removeItem
-  ) where
-
-removeItem :: Eq a => a -> [a] -> [a]
-removeItem _ [] = []
-removeItem r (x:xs)
-  | r == x = removeItem r xs
-  | otherwise = x : removeItem r xs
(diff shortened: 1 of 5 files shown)


yesod-newsfeed/Yesod/Common.hs
```diff
@@ -0,0 +1,9 @@
+module Yesod.Common
+  ( removeItem
```
gregwebs: what is this used for?

gregwebs:
Thanks for the contribution! Can you add an entry to the ChangeLog.md ?

nek0:
sure.

yesod-newsfeed/Yesod/Common.hs
nek0: it isn't used anymore. I used it to clean the lists up a bit, which didn't work out.

nek0:
changelog has arrived

gregwebs:
Thanks for the great Changelog! Unfortunately this is a breaking change. Let me know if you have any other breaking changes in mind. If not we can release this now with a version bump

nek0:
I have no more changes in mind.

gregwebs:
@snoyberg I need hackage access to upload this package. yesod-newsfeed/1.5 tag is already pushed.

snoyberg:
Done

nek0:
Hackage still ships 1.4.0.1 as most recent version...

gregwebs:
uploaded now

PR #1081, opened 2015-10-02
add a runHttpRequest typeclass method

gregwebs:
I am trying to make it possible to handle HTTP failures in yesod-auth

### DIFF
--- a/yesod-auth/ChangeLog.md
+++ b/yesod-auth/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.4.7
+
+* add a runHttpRequest function for handling HTTP errors
+
 ## 1.4.6
 
 * Use nonce package to generate verification keys and CSRF tokens [#1011](https://github.com/yesodweb/yesod/pull/1011)
--- a/yesod-auth/Yesod/Auth.hs
+++ b/yesod-auth/Yesod/Auth.hs
@@ -46,6 +46,7 @@ module Yesod.Auth
     , asHtml
     ) where
 
+import           Control.Applicative ((<$>))
 import Control.Monad                 (when)
 import Control.Monad.Trans.Maybe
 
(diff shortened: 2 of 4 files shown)


gregwebs:
The idea is that one can run this in a `try`, log the error, and retry the request. The reason that `Request` is passed as a parameter is so that it can be inspected in case that helps decide how to handle the failure.

snoyberg:
It seems like a strange type signature. I'd have expected something like:

```haskell
runHttpRequest :: Request -> (Response BodyReader -> HandlerT master IO a) -> HandlerT master IO a
```

It just seems like the current signature may be slightly too limiting, e.g., what about adding in auto-retry logic?

gregwebs:
My concept is that this is just for dealing with errors. So I don't want to make a friendlier API for modifying the response. I agree that this is a strange API for dealing with errors because there is nothing explicit in the types about errors, but I think this provides all the power one needs. Adding retry logic seems simple enough.

```
= try (doReq req) >>= either (doReq req) return
```

snoyberg:
The problem is that such a signature will not allow distinguishing between rerunning the request and rerunning the user actions taken thereafter with the `Response`. That said, I don't feel strongly about this, if you want to keep this type signature, feel free to merge.

gregwebs:
I actually think it is desirable to be able to run Handler code afterwards that concludes with `sendResponse` and sends something entirely different.

If that is what you were getting at? Before you said the type signature was too limiting but now you are saying it is not limiting enough? 

I am open to changing this, I just want to have something very extensible without having to program a ton of stuff. What would your default implementation be?

snoyberg:
I guess "limiting" depends on who's side you're looking at ;)

I'm not sure that wrapping up `sendResponse` will work as expected, since it short-circuits by throwing an exception.

Anyway, my proposed implementation would be something like the following (not actually type checked):

```haskell
runHttpRequest :: Request -> (Response BodyReader -> HandlerT master IO a) -> HandlerT master IO a
runHttpRequest req inner = do
  man <- asks authHttpManager
  HandlerT $ \r -> withResponse req man $ \res -> unHandlerT (inner res) t
```

Which could be cleaned up if desired with monad-control, though explicit is kind of nice here.

gregwebs:
@snoyberg updated with a compiling form of your suggestion

snoyberg:
This introduces a breaking change by requiring HasHttpManager. While I think that's a good change, I'm not sure we want to bundle that in now. Afaict, it's possible to implement this without that superclass constraint.

gregwebs:
good point. I changed the code to use http-client in the typeclass as per your original suggestion.

PR #1083, opened 2015-10-10
Thread leak when calling neverExpires 

DaveCTurner:
My commit (42f098ff64d5a9c8f8cccac7ebc940ecdf45dfc4) causes `neverExpires` to leak threads rather badly. See https://groups.google.com/forum/#!topic/yesodweb/TJgI1LKtMgM for graphs.

I clearly misunderstood where I'd added the call to `mkAutoUpdate` - I intended it to be when the application was being set up but now I look at it again I see that it's calling it in every handler.

I can't promise I will be able to fix this for a few days, but a fix is on its way.

Deepest apologies!

### DIFF
--- a/yesod-core/Yesod/Core/Dispatch.hs
+++ b/yesod-core/Yesod/Core/Dispatch.hs
@@ -240,5 +240,5 @@ warpEnv site = do
 getGetMaxExpires :: IO (IO Text)
 getGetMaxExpires = mkAutoUpdate defaultUpdateSettings
   { updateAction = getCurrentMaxExpiresRFC1123
-  , updateFreq = 60 * 60 * 1000000 -- Update once per hour
+  , updateFreq = 24 * 60 * 60 * 1000000 -- Update once per day
   }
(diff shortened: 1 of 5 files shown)


DaveCTurner:
Hmm, I was thinking something more along these lines. Would you prefer this or something that uses `unsafePerformIO`?

gregwebs:
I know the value is cached, but I was trying to get at an approach where there is never a need to even start the update thread or do any work if you don't call `neverExpires` (I work on a large application that does not). That would be my preference.

DaveCTurner:
OK, that sounds solvable, but is it worth the effort? An auto updater forks
a thread that immediately blocks on a MVar until the value is requested,
which is pretty cheap in my understanding.

I know the value is cached, but I was trying to get at an approach where
there is never a need to even start the update thread or do any work if you
don't call neverExpires (I work on a large application that does not). That
would be my preference.

—
Reply to this email directly or view it on GitHub
<https://github.com/yesodweb/yesod/pull/1083#issuecomment-147122383>.

gregwebs:
it is really only worth it if it avoids bugs like this one. So I am fine to merge this change in.

DaveCTurner:
Ok, I've pushed the other two changes you suggested.

gregwebs:
Thanks!

PR #1089, opened 2015-10-19
Add Foldable and Traversable instances for FormResult

andrewthad:
This will let users operate on a `FormResult` with a bunch of useful combinators. In particular, I'm thinking of `forM_`, since when handling a `FormResult`, I often write:

    case res of
        FormSuccess myData -> do
          ...
        _ -> return ()

Instead, this let's people write:

    forM_ res $ \myData -> do
        ...

I'm 99% sure that the instances I've provided are law-abiding, but if anyone wants to confirm that for me, I would appreciate it.

### DIFF
--- a/yesod-form/Yesod/Form/Types.hs
+++ b/yesod-form/Yesod/Form/Types.hs
@@ -35,6 +35,8 @@ import Data.String (IsString (..))
 import Yesod.Core
 import qualified Data.Map as Map
 import Data.Semigroup (Semigroup, (<>))
+import Data.Traversable
+import Data.Foldable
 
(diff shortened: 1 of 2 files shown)


snoyberg:
This looks fine to me. Could you do a version bump in the .cabal file and add relevant `@since` Haddocks to the instances?

andrewthad:
Good call. Done.

snoyberg:
Merged! I had to modify the version bump to be a minor instead of patch bump.

andrewthad:
Ah, that's what the fourth number is for. That makes since.

Issue #1090, opened 2015-10-19
Logged in users can visit /auth/login


pharpend:
Steps to reproduce:

1. [Build Snowdrift](https://git.gnu.io/snowdrift/snowdrift/blob/master/BUILD.md), run it via `stack exec -- yesod devel`.
2. Visit <http://localhost:3000/auth/login>, log in as `admin`, password `admin`.
3. Visit <http://localhost:3000/auth/login> again. You can even log in again.

We tried to fix this, thinking it was a bug in Snowdrift, but it appears to be a bug in yesod-auth.

Snowdrift ticket: <https://github.com/snowdriftcoop/snowdrift/issues/144>

Thanks,

Peter Harpending

gregwebs:
Thanks for reporting. I think this hasn't been reported as an issue before because in most applications it is rare for a user to go straight to the login page. If you have an /auth/login page you would normally have links on your home page to enter the app, and then if they are not logged in they would get redirected to /auth/login.

I think fixing this is probably pretty easy though.

wolftune:
@gregwebs Snowdrift is the same in that the links in the navigation change when you are logged in, so it remains rare for this to happen. It's a weird edge case we came across. But it's still awkward thing that should be fixed.

paul-rouse:
I am a little concerned that there isn't a single obviously correct behaviour.  If an authenticated user goes explicitly to `/auth/login`, do you redirect the same way as after a successful login (as decided by `redirectToReferer` and `loginDest`), or do you have a special page reminding them they are logged in, or do you give an error response?  I can see reasons why any of those, or even the status quo, might be correct in a particular application, and conversely why any of them might seem odd to the user.

As an alternative, it is fairly easy to fix this in the application, and make whichever choice you want.  The application can define `loginHandler` for the `YesodAuth` instance, and start with something like:

```
ma <- lift maybeAuthId
when (isJust ma) $
    lift $ redirect HomeR
```

However, this is a little harder than it needs to be because you probably want to drop into the default implementation straight after this, unless there are other reasons for overriding it.  It would be made easier if we exposed the default implementation, say as `defaultLoginHandler`.  If that's an acceptable answer, I am quite happy to PR it.

wolftune:
My thinking would be: redirect as per a successful log-in *but* have an alert show that says "you are already logged in, click here to sign out if you want to log in as a different user" or something like that.

paul-rouse:
[...]

PR #1091, opened 2015-10-21
Provide default loginHandler as separate function

paul-rouse:
This is the first, non-breaking part of my suggestion for #1090


gregwebs:
great!

PR #1095, opened 2015-10-27
Add Functor instance for Option and OptionList

andrewthad:
I doubt that this there will be any objection to these instances. Could someone look this over?


snoyberg:
Lgtm

PR #1099, opened 2015-11-13
Have clientSessionDateCacher use auto-update

DaveCTurner:
Tracked down the problem I was having at https://mail.haskell.org/pipermail/haskell-cafe/2015-November/122172.html to the thread started within `clientSessionDateCacher`.

A PR will follow shortly.

### DIFF
--- a/yesod-core/Yesod/Core/Internal/Session.hs
+++ b/yesod-core/Yesod/Core/Internal/Session.hs
@@ -45,11 +45,14 @@ decodeClientSession key date rhost encrypted = do
 ----------------------------------------------------------------------
 
 
--- Mostly copied from Kazu's date-cache, but with modifications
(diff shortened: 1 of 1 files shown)


DaveCTurner:
Tempted to also get rid of the second component in the return type as (a) it wasn't obviously called and (b) it is now trivial. What do you think?

meteficha:
Looks good to me! :+1: 

I'm opposed to changing its type as it will increase pain without giving much back.  But I'd like a comment explaining it, e.g., `The second component of the tuple returned used to be an action that kills the thread that is used to update the cache.  Now it's a noop.`.

DaveCTurner:
Right, done.

meteficha:
Great! :shipit:

PR #1100, opened 2015-11-23
Add UserName message constructor

geraldus:


geraldus:
Travis failed with following errors:
7.8.4:
> cabal: /home/travis/.cabal/logs/transformers-base-0.4.4.log: does not exist

7.10.1
> cabal: /home/travis/.cabal/logs/network-2.6.2.1.log: does not exist

snoyberg:
Can you apply the advice from this blog post?

http://www.yesodweb.com/blog/2015/07/s3-hackage-mirror-travis

I just noticed this problem with Hackage on another repo

geraldus:
@snoyberg, surely, already in progress, meanwhile can you give some comments, for example:
* should "User Name" become "User name"?
* what about suffixes -- I can drop this commit or make convert suffixes to lowercase.

geraldus:
Ok, Travis done.

snoyberg:
I like UserName. I'm not sure what you mean be suffixes though, can you clarify? The other changes seem fine too, if that's what you mean.

geraldus:
@snoyberg, hm, maybe I should say postfixes (I'm not sure).  I'm talking about these changes:
"Login" → "Log In". So, should I convert "In"s and "Out"s to lowercase?

gregwebs:
If it is a title that would indicate to me that all the words should be upper cased. But I don't think there is a right or wrong answer here.

geraldus:
@snoyberg @gregwebs 
Also, maybe we should introduce `LogoutTitle` message which is the same as existing `Logout` for the sake of consistency (and deprecate the latter).

_Initially I wanted to introduce `Login` message for input's label, but then I've found that this may confuse users, fortunatelly I quicky found `UserName` as replacement._

gregwebs:
makes sense to me. But @snoyberg may take issue with a breaking change. Perhaps keep the existing constructor and point it to the new one?

snoyberg:
I'm OK with introducing new constructor for this without bumping the major version number.

geraldus:
Done, please review everything once again.

snoyberg:
Thanks!

geraldus:
@lubomir , I've created [another PR](https://github.com/yesodweb/yesod/pull/1102) for this.

PR #1103, opened 2015-11-24
Travis with stack

snoyberg:
We've been seeing a lot of failing builds with cabal recently with the "not found" error message, implying network connection issues. Switching to use the Hackage mirror on S3 hasn't solved the problem, but moving to Stack (which has built-in retry logic) should do the trick. Downside: we're testing against a curated package set instead of dependency solving. I think it's worth the trade-off though.


gregwebs:
I am a little wary of testing against stack because we will have worse support for anyone using cabal. On the other hand, experience has shown that yesod cannot really support cabal.

Do you think it would be a good idea to keep 1 or more of the cabal builds and have them be an allowed failure? That way we can still try to stay on top of some of the version bound tweaking.

snoyberg:
I don't think there will be a lot of value in ignored cabal failures to be
honest. It's also significantly more work to set up than just using the
standard Stack Travis formula

geraldus:
Maybe I'm wrong, but it seems now no tests are running during build.

snoyberg:
Looks like tests are running to me, e.g.
https://travis-ci.org/yesodweb/yesod/jobs/92994157#L1266. What evidence do
you see to the contrary?

geraldus:
Oh, never mind, I have missed word "test" there (:

PR #1104, opened 2015-11-24
Simple auth plugin for hardcoded auth entities

geraldus:
I'm not sure if this will be useful for someone, feel free to reject this module.
Would be happy to have any comments on this.

### DIFF
--- a/yesod-auth/Yesod/Auth/Hardcoded.hs
+++ b/yesod-auth/Yesod/Auth/Hardcoded.hs
@@ -1,0 +1,75 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE QuasiQuotes           #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeFamilies          #-}
+module Yesod.Auth.Hardcoded
+  ( YesodAuthHardcoded(..)
+  , authHardcoded
+  , loginR )
+  where
+
+import           Yesod.Auth          (Auth, AuthPlugin (..), AuthRoute,
+                                      Creds (..), Route (..), YesodAuth,
+                                      loginErrorMessageI, setCredsRedirect)
+import qualified Yesod.Auth.Message  as Msg
+import           Yesod.Core
+import           Yesod.Form          (ireq, runInputPost, textField)
+
+import           Control.Applicative ((<$>), (<*>))
+import           Data.Text           (Text)
+
+
+loginR :: AuthRoute
+loginR = PluginR "hardcoded" ["login"]
+
+class (YesodAuth site) => YesodAuthHardcoded site where
+
+  -- | Check whether given user name exists among hardcoded names.
+  doesUserNameExist :: Text -> HandlerT site IO Bool
+
+  -- | Validate given user name with given password.
+  validatePassword :: Text -> Text -> HandlerT site IO Bool
+
+
+authHardcoded :: YesodAuthHardcoded m => AuthPlugin m
+authHardcoded =
+  AuthPlugin "hardcoded" dispatch loginWidget
+  where
+    dispatch "POST" ["login"] = postLoginR >>= sendResponse
+    dispatch _ _ = notFound
+    loginWidget toMaster =
+      [whamlet|
+        $newline never
+        <form method="post" action="@{toMaster loginR}">
+          <table>
+            <tr>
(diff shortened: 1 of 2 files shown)


gregwebs:
Thanks. Can't you have a default definition for isUserExists ?

geraldus:
@gregwebs I have clear idea about this.  I think it is possible if `YesodAuthHardcoded` class will contain one extra property, let's say `hadrcodedUsers :: [AuthHardcodedLogin site]`.  What do you think?

Also, I see now that `HardcodedCreds` data type is redundant in current implementation.

geraldus:
I've tried to implement such class property _(what is the right word -- operation?)_ as a `Map`, e.g.

```haskell
class (YesodAuth site) => YesodAuthHardcoded site where
  type AuthHardcodedLogin site

  -- | A list of all hardcoded uses names.
  hardcodedUsers :: Map.Map Text (AuthHardcodedLogin site)
  hardcodedUsers = Map.empty

  isUserNameExists username =
    return
      (case Map.lookup username hardcodedUsers of
         Nothing -> False
         Just  _ -> True)

```

But this way gives ambiguity when trying to get this list in `isUserExists` function.

> Could not deduce (YesodAuthHardcoded site0)
>   arising from a use of ‘hardcodedUsers’
> from the context (YesodAuthHardcoded site)
>   bound by the class declaration for ‘YesodAuthHardcoded’
>   at …
> The type variable ‘site0’ is ambiguous
> In the second argument of ‘Map.lookup’, namely ‘hardcodedUsers’
> In the expression: Map.lookup username hardcodedUsers
> In the first argument of ‘return’, namely
>  …

I believe this will be solved if I add extra param for both functions:

```haskell
hardcodedUsers :: site -> Map.Map Text (AuthHardcodedLogin site)
isUserNameExists :: site -> Text -> HandlerT site IO Bool
```

But, maybe you talking about simplest way?  I could think about database check, but this requires one extra constraint to class -- `YesodAuthPersist`.

geraldus:
Ok, I think it is good point, that `isUserExists` should take `site` argument.  What about this?

```haskell
class (YesodAuth site) => YesodAuthHardcoded site where
  type AuthHardcodedLogin site

  -- | A list of all hardcoded uses names.
  hardcodedUsers :: site -> Set.Set Text
  hardcodedUsers _ = Set.empty

    -- | Check whether given user name exists among hardcoded names.
  isUserNameExists :: site -> Text -> HandlerT site IO Bool
  isUserNameExists site = return . flip Set.member (hardcodedUsers site)

  -- | Validate given user name with given password.
  validatePassword :: Text -> Text -> HandlerT site IO Bool

```

geraldus:
What about this?

```haskell
class (YesodAuth site) => YesodAuthHardcoded site where
  type AuthHardcodedLogin site

  -- | A list of all hardcoded uses names.
  hardcodedUsers :: site -> Map.Map Text (AuthHardcodedLogin site)
  hardcodedUsers _ = Map.empty

  -- | Check whether given user name exists among hardcoded names.
  isUserNameExists :: site -> Text -> HandlerT site IO Bool
  isUserNameExists site username =
    do mayUser <- getUser site username
       return (maybe False (const True) mayUser)

  -- | Return user entity by its name.
  getUser :: site -> Text -> HandlerT site IO (Maybe (AuthHardcodedLogin site))
  getUser site username = return (Map.lookup username (hardcodedUsers site))

  -- | Validate given user name with given password.
  validatePassword :: Text -> Text -> HandlerT site IO Bool

```

gregwebs:
Ahh, I think I wasn't really understanding things when I asked that question. My suggestion would be

``` haskell
isUserNameExists :: Text -> HandlerT site IO Bool
```

gregwebs:
Ahh, I think I wasn't really understanding things when I asked that question. My suggestion would be

``` haskell
isUserNameExists :: Text -> HandlerT site IO Bool
isUserNameExists _ = return False
```

gregwebs:
For these typeclass configurations it is better to have a few general, flexible, easy to understand functions than lots of little functions.

Also, it would be better to rename to `doesUserNameExists`

mrkkrp:
Maybe even `doesUserNameExist`.

geraldus:
Ok, I've simplified a class and removed all default implementations.

geraldus:
@gregwebs I've updated documentation and tested this plugin locally.

gregwebs:
Thanks! Why does this require MonadThrow ?

geraldus:
@gregwebs I have no idea, I wanted to put login widget in top level, and type signature inferred by GHC contains this constraint.  But it is possible not to depend on `exceptions` package and avoid `MonadThrow` constraint by moving widget definition in `where` block.  I've updated code.

gregwebs:
thanks! This is on hackage now

Issue #1101, opened 2015-11-23
Missing convenience function for custom status code + JSON return type


bitemyapp:
Ended up using sendResponseStatus + toJSON, but the lack of obvious apparent way to `returnJson` with a different status code can get new people turned around.

Snap avoids this by having the ability to set status codes inside MonadSnap, like so:

```
modifyResponse $ setResponseStatus 500 "Internal Server Error"
```

Then you can just `>>` that with your JSON return.

While technically this is accommodated with `sendResponseStatus` via `ToTypedContent`, it's not going to be obvious to new people that all they have to do is wrap their value in `toJSON`. I realize you don't want to make one-off concretions of `sendResponseStatus` for each return type, but something like "return status 201 + JSON" is going to be pretty common for anyone wanting to make a REST API in Yesod.

Only thing I could find via Googling that came close to documenting how this worked was [Brisbin's handy blog post](https://pbrisbin.com/posts/writing_json_apis_with_yesod/), but even that doesn't specifically address returning JSON (only Text values).

gregwebs:
Yeah, Right now error handling is too hard-coded.
I have been meaning to commit some code I created for dealing with this.

@snoyberg what do you think?


``` haskell
module ApiError
  ( ApiError(..)
  , send
  , sends
  , messageAndErrors
  ) where

import ClassyPrelude
import Yesod hiding (Unauthorized)
import Network.HTTP.Types
import FrontRow.Helpers.Error
import qualified Data.Aeson.Types as A

type Handler a = forall site.  (YesodPersist site, Yesod site)
              => HandlerT site IO a

data ApiError
  = Forbidden (Maybe Text) -- ^ 403
  | Unauthorized           -- ^ 401
  -- Keep this open so that it is easy to defined error types where they are used
  -- for type-safety, none of the ApiError constructors should have a ToJSON instance
  | forall err. (Eq err, Show err, ToJSON err) => BadRequest err

instance Show ApiError where
  show Unauthorized    = "Unauthorized"
  show (Forbidden msg) = "Forbidden " <> (unpack $ fromMaybe "" msg)
  show (BadRequest e)  = "BadRequest " <> show e

-- Hacky, but only used in tests
instance Eq ApiError where
  Unauthorized    == Unauthorized    = True
  (Forbidden msg1) == (Forbidden msg2) = msg1 == msg2
  -- using ToJSON (we could be slighly hackier and use Show) is pretty hacky
  -- but this is only needed for testing purposes
[...]

PR #1105, opened 2015-11-25
JSON-specific sendResponseStatus

bitemyapp:
Addressing #1101 

Name is a little verbose. Not married to this method of dealing with this problem.


gregwebs:
This saves typing `to` with the goal of providing documentation.
The only thing one can send is a response, so I would favor `sendStatusJson`.

bitemyapp:
@gregwebs I like that. I will update it accordingly.

snoyberg:
I'm good with this. I'll wait for @gregwebs to merge since he had made some comments.

gregwebs:
thanks! Sorry for the delay

PR #1112, opened 2015-12-02
Minor auth tweaks

mrkkrp:
This fixes a typo and bumps password strength to compensate 2 years. Value 16 is based on file `PasswordStore.hs` which I suppose was originally written in 2011 and proposed password strength back then was 12. In 2013 proposed strength was 14, so extrapolating this linearly (linearly with respect to “strength” value, and exponentially with respect to computational speeds) we get 16 in 2015, which is close to its end. 16 should be good enough, but I'm not an expert in this area.


snoyberg:
Thanks!

PR #1116, opened 2015-12-06
Added an example with email auth and an ses mailer

lethjakman:
This should help beginners get started with email authentication and
utilize Amazon SES for reliable mailing instead of sendmail because
sendmail is inherently unreliable with consumer based IP addresses.

Points of interest:
* Pulls SES authentication details from config/secrets.yaml to keep the
  secrets out of the source code
* Needs the from email address to be confirmed with Amazon SES

### DIFF
--- a/demo/auth/config/secrets.yaml
+++ b/demo/auth/config/secrets.yaml
@@ -1,0 +1,2 @@
+accessKey: <your access key>
+secretKey: <your secret key>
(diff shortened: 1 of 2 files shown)


gregwebs:
great! Would it make sense to link to this from YesodAuthEmail?

lethjakman:
Yes, I think it would. Would you like me to add an additional change for that?

gregwebs:
yes, we need to make sure people can find the examples! sorry for the late reply

Issue #1018, opened 2015-06-23
stack exec -- yesod devel


legrostdg:
Not sure if the issue is in yesod-bin or in stack, but after building a yesod based project and installing yesod-bin with stack, I can't run "yesod devel":

    $ stack exec -- yesod devel
    Yesod devel server. Press ENTER to quit
    Application can be accessed at:
    
    http://127.0.0.1:3000
    https://127.0.0.1:3443
    If you wish to test https capabilities, you should set the following variable:
      export APPROOT=https://127.0.0.1:3443
    
    cabal: Use of GHC's environment variable GHC_PACKAGE_PATH is incompatible with
    Cabal. Use the flag --package-db to specify a package database (it can be used
    multiple times).

snoyberg:
It's a problem in Cabal, in that it won't recognize the `GHC_PACKAGE_PATH` environment variable, and yesod devel which uses cabal. The proper solution will be to move yesod devel off of cabal, which will hopefully be part of the current GSoC project to move over to ide-backend. In the meanwhile, unfortunately yesod devel is not compatible with stack, at least not directly as we'd like.

legrostdg:
Is there a workaround? I noticed the same problem with propellor https://propellor.branchable.com, which I guess won't be updated to ide-backend...

https://github.com/haskell/cabal/pull/2270 may be a fix?

snoyberg:
It depends on what you mean by "workaround." You can use `env` to disable the GHC_PACKAGE_PATH environment variable (and we can provide a flag to `exec` to automate that). But in the case of yesod devel, that will mean you end up not having the packages in scope.

Another possibility would be to tweak `yesod devel` to parse GHC_PACKAGE_PATH, turn it into command line parameters for cabal, and unset the environment variable. Which presumably is what cabal should be doing anyway. propellor could do the same thing.

legrostdg:
closed by a7cccf2a7c5df8b26da9ea4fdcb6bac5ab3a3b75
Thanks!

snoyberg:
You're fast, the blog post isn't even out yet!

geraldus:
Oh, cool, thanks!  Now I can use `stack` (:

legrostdg:
@snoyberg you are the fast one!

I just updated my yesod git clone by chance and noticed the latest commit :).

tfiala:
I seem to be hitting this issue with 'stack exec -- yesod test':

```
stack exec -- yesod test
[...]

PR #1123, opened 2015-12-11
update testing instructions according to #1018

danse:


danse:
I was having errors running `stack exec -- yesod test` after scaffolding a new app, then i found on #1018 that other users had the same problems and that the general suggestion is to use `stack test` directly. So i thought to update the text

snoyberg:
Please send this PR to the postgres branch of the yesod-scaffold repo: https://github.com/yesodweb/yesod-scaffold

danse:
[Here](https://github.com/yesodweb/yesod-scaffold/pull/115) is that pull request. I will try to do similar pull requests on the other relevant branches. In general i try to avoid using branches in order to logically group code, it makes reuse and maintenance harder in my opinion. Branches just were not designed for that

danse:
For example, in this case i need to do a full text search for "yesod test". That was easier here than in the different branches

danse:
Luckily i can use this pull request as an index of branches to look in :P

danse:
Pull requests from 115 to 120 on `yesod-scaffold` correspond to this

PR #1124, opened 2015-12-11
Fix references to ‘Yesod.Core.Handler’

mrkkrp:
The links don't work:

http://haddock.stackage.org/nightly-2015-12-11/yesod-core-1.4.17/Yesod-Core.html#v:Handler

because Haddock's syntax for module references is double quotes `"Yesod.Core.Handler"`. Also I removed the incomplete phrase “Use this midle”.

Note that Emacs eliminated trailing whitespace. I committed that too, because trailing whitespace is nothing good.


snoyberg:
Thanks! In the future, I'd recommend whitespace-only changes be kept in separate commits

mrkkrp:
Sure. One question: I see you keep descriptions as very long lines. Is this on purpose? Should new descriptions be written like this?

snoyberg:
I don't care one way or another.

PR #1129, opened 2015-12-23
Add error class when help text is an error.

silky:
Does anyone care for this feature?

It doesn't appear as part of the Bootstrap3 spec - http://getbootstrap.com/css/#forms-control-validation - but it does actually appear to be useful.

For example, I have a field which has a permanent tool-tip; explaining what it is. But when someone doesn't fill it out correctly, but the validation only happens on the server, then the error gets sent back as a `help-block` that sits *below* the original help text. It is styled in exactly the same way.

The parent element having the `is-error` class is pretty useless, because if I were to style both of them as errors, it would show the tooltip help text as an error, *as well* as the error, which I don't think is ideal.

Opening this up for comment by way of pull request. Feel free to reject if it's not a good solution for everyone, and I'll just use something like this locally (or feel free to suggest a better way).

Thanks!

### DIFF
--- a/yesod-form/Yesod/Form/Bootstrap3.hs
+++ b/yesod-form/Yesod/Form/Bootstrap3.hs
@@ -191,7 +191,7 @@ helpWidget view = [whamlet|
     $maybe tt <- fvTooltip view
       <span .help-block>#{tt}
     $maybe err <- fvErrors view
-      <span .help-block>#{err}
+      <span .help-block .error-block>#{err}
 |]
 
 


gregwebs:
makes sense to me, but maybe it should be a slightly less generic class name?

silky:
okay; what name do you recommend?

`error-block` seems on par with `help-block`; but perhaps `error-info-block` or something?

i think the `help-block` selector already is pretty specific, so adding another one let's us customise it jointly (`.help-block.error-block`) so that's pretty specific.

gregwebs:
My fear is that since `.help-block` exists in bootstrap that `.error-block` will also. But maybe that will actually be fine.

PR #1144, opened 2016-01-13
Allow subsites within hierarchical routes

ajnsit:
This allows subsites to work within hierarchical routes like so -

    /sub SubR:
      /hello HelloSubR HelloSubRoute getHelloSubRoute

`HelloSubR` takes the subsite route as an argument which needs to be folded through the parent route constructors. Not doing so is definitely a bug, regardless of whether hierarchical subsites are supported.

### DIFF
--- a/yesod-core/Yesod/Routes/TH/Dispatch.hs
+++ b/yesod-core/Yesod/Routes/TH/Dispatch.hs
@@ -176,11 +176,12 @@ mkDispatchClause MkDispatchSettings {..} resources = do
                     subDispatcherE <- mdsSubDispatcher
                     runHandlerE <- mdsRunHandler
                     sub <- newName "sub"
+                    sroute <- newName "sroute"
                     let sub2 = LamE [VarP sub]
                             (foldl' (\a b -> a `AppE` b) (VarE (mkName getSub) `AppE` VarE sub) dyns)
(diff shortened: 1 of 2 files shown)


snoyberg:
Can you add a test case of something that didn't work previously and does work now?

ajnsit:
Do you mean a testcase in the test suite for yesod? I tried adding a test case but I couldn't immediately figure out how the test suite works. Does the test suite currently test any subsite functionality? I can use that as a template and try again when I have some more time.

Meanwhile, I have created a small yesod project to demonstrates the problem (https://github.com/ajnsit/yesod-subsite-test). If I try to compile this project without this PR, I get error messages (https://github.com/ajnsit/yesod-subsite-test/blob/master/error_master.txt).

Looking at the error message - It tries to assign the type `Route HelloSub -> Route App` to `SubR SubsiteNotOkR`. With this PR, the TH will generate `SubR . SubsiteNotOkR` instead which has the correct type.

snoyberg:
Yes, I meant a yesod-core test suite case. You can look in
test/YesodCoreTest for some simple mostly-standalone tests. WaiSubsite, for
example, includes a subsite in it, which may be a good template.

ajnsit:
Done! Thanks for the pointer to WaiSubsite! It was extremely easy to add a test case to that.

snoyberg:
Thanks!

PR #1145, opened 2016-01-13
Pass hierarchical route arguments to subsites

ajnsit:
This is in continuation to https://github.com/yesodweb/yesod/pull/1144 but as a separate PR because this feature may or may not be desired.

Assuming a route definition like this -

    /sub/#Text SubR:
      /hello/#Int HelloSubR HelloSubRoute getHelloSubRoute

With this change, both the #Text and #Int arguments will be passed to "getHelloSubRoute". Without this change (but with https://github.com/yesodweb/yesod/pull/1144), it would be just #Int.

This is in line with regular nested routes -

    /sub/#Text SubR:
      /hello/#Int HelloSubR GET

Where "getHelloSubR" will be passed both #Text and #Int.

### DIFF
--- a/yesod-core/Yesod/Routes/TH/Dispatch.hs
+++ b/yesod-core/Yesod/Routes/TH/Dispatch.hs
@@ -176,9 +176,10 @@ mkDispatchClause MkDispatchSettings {..} resources = do
                     subDispatcherE <- mdsSubDispatcher
                     runHandlerE <- mdsRunHandler
                     sub <- newName "sub"
+                    let allDyns = extraParams ++ dyns
                     sroute <- newName "sroute"
                     let sub2 = LamE [VarP sub]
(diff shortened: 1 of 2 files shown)


snoyberg:
Two requests:

* Can you add a test case for this change as well like the previous one?
* Can you rebase against master?

ajnsit:
Hmm, I tried adding a test case but started getting compile errors in the generated TH. I had thought this works but I may have skipped actually testing it with nested arguments in Yesod (the change works fine with wai-routes https://github.com/ajnsit/wai-routes/commit/951e46b553c2b8a4daf6758a98e192f86b00dd60).

I'll spend some time looking at it and update the pull request when I figure it out.

snoyberg:
Thanks! For future reference, edits to a comment don't trigger a ping, which is why I didn't notice this initially

ajnsit:
Great, thanks!

PR #1149, opened 2016-01-20
Croatian translation

vlatkoB:


yesod-form/Yesod/Form/I18n/Croatian.hs
```diff
@@ -0,0 +1,26 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Yesod.Form.I18n.Croatian where
+
+import Yesod.Form.Types (FormMessage (..))
+import Data.Monoid (mappend)
+import Data.Text (Text)
+
+englishFormMessage :: FormMessage -> Text
```
snoyberg: This should be `croatianFormMessage`

vlatkoB: Sorry, must have been hasty.

PR #1155, opened 2016-02-02
Document feed entry enclosures

chreekat:
Hopefully this will prevent the next lost soul from bleakly facing a new, mysterious 3-tuple

### DIFF
--- a/yesod-newsfeed/Yesod/FeedTypes.hs
+++ b/yesod-newsfeed/Yesod/FeedTypes.hs
@@ -28,6 +28,12 @@ data Feed url = Feed
     , feedEntries     :: [FeedEntry url]
     }
 
+-- | RSS and Atom allow for linked content to be enclosed in a feed entry.
+-- This represents the enclosed content.
+--
+-- Atom feeds ignore 'enclosedSize' and 'enclosedMimeType'.
+--
+-- @since 1.6
 data EntryEnclosure url = EntryEnclosure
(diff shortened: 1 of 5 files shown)


snoyberg:
Pinging @nek0 for feedback. How do you feel about this change?

Note: this will be a major version bump, but I'm OK with doing so in this case.

nek0:
Looks very good. It's more clear now.

snoyberg:
@chreekat could you add some extra comments to this as well, including the `@since` annotations and a major version bump?

chreekat:
Will do.

yesod-newsfeed/Yesod/FeedTypes.hs
```diff
@@ -27,11 +28,27 @@ data Feed url = Feed
...
     { feedEntryLink    :: url
     , feedEntryUpdated :: UTCTime
     , feedEntryTitle   :: Text
     , feedEntryContent :: Html
-    , feedEntryEnclosure :: Maybe (url, Int, Text)
+    , feedEntryEnclosure :: Maybe (EntryEnclosure url)
+      -- ^ Allows enclosed data: RSS \<enclosure> or Atom \<link
+      -- rel=enclosure>
+      --
+      -- @since 1.5
```
chreekat: The @\since here is for the field itself, not the particular type of the field.

snoyberg:
Thanks!

PR #1157, opened 2016-02-08
Adjustments for split read and write DB

pseudonom:
Goes with https://github.com/yesodweb/persistent/pull/534


pseudonom:
Hm, does this permission denied error in travis make sense to anyone? Is it a transient thing that triggering a rebuild might solve?
```
Cloning into '/home/travis/build/yesodweb/yesod/.stack-work/downloaded/338de38a6fac895a0dfdfd64cd279e13cc4286a72fd4ed186255fe57a7ebe3c1.tmp'...
Warning: Permanently added the RSA host key for IP address '192.30.252.131' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
```

snoyberg:
Do you have a submodule using ssh auth?

pseudonom:
Ah, I did add a git reference in `stack.yaml` to point to the PRed version of persistent.

```yml
 - location:
      git: git@github.com:pseudonom/persistent.git
      commit: 0edba747e49c816e94c235bced16978cb729b175
    subdirs:
    - persistent
```

Is there a way to make this work with travis or should I just temporarily vendor that version of `persistent` in a subdirectory?

snoyberg:
Just switch to the https URL instead, it should work just fine.

pseudonom:
Ah, thanks! Didn't realize `stack` supported both. Looks like everything should be good to go now?

pseudonom:
Bumped it to work with the newly released `persistent-2.5`.

yesod-auth/Yesod/Auth.hs
```diff
@@ -500,10 +500,10 @@ class (YesodAuth master, YesodPersist master) => YesodAuthPersist master where
     getAuthEntity :: AuthId master -> HandlerT master IO (Maybe (AuthEntity master))
 
     default getAuthEntity
-        :: ( YesodPersistBackend master
-               ~ PersistEntityBackend (AuthEntity master)
+        :: ( YesodPersistBackend master ~ backend
+           , BaseBackend backend ~ PersistEntityBackend (AuthEntity master)
```
gregwebs: Can use the new constraint kind synonym here.

gregwebs:
looks good other than it could use `PersistRecordBackend`. I will let @snoyberg merge this.

snoyberg:
To avoid upgrade headaches, we should keep compatibility with the old version of persistent via CPP. It's a pain, but it's better to keep the pain centralized in Yesod itself and not affect users. Would you be able to update the PR?

pseudonom:
Sure, I can do that.

pseudonom:
Okay, `CPP`ed it.

snoyberg:
Thanks!

PR #1158, opened 2016-02-09
Add Summernote WYSIWYG html field

geraldus:
This is raw attempt to bring Summernote support.  I used `Nic.hs` as example.

Summernote is MIT licenced.


geraldus:
I have created special project to test this
https://github.com/geraldus/yesod-summernote-test

snoyberg:
Is there an advantage to having this in yesod-form itself as opposed to a separate yesod-form-summernote package? The advantage of the latter is that it would be free to make breaking API changes without being constrained to the yesod-form release cycle.

geraldus:
I added this because there is NIC edit which looks a bit outdated (unmaintained), and I thought that Summernote could be a good replacement.

snoyberg:
If I were to do things again right now I'd keep that module in a separate
package too, it's basically grandfathered in now. I'd be happy putting a
comment at the top of that module (and the package description for
yesod-form) pointing to a separate summernote package.

geraldus:
Hmm… It seems to be a good reason to put very first package on Hackage (I do not have anything yet).  Ok, I'll create a separate package and  then I can update this PR (force-pushing mentioned comments) or you can close this and I'll send anther one.  Likely it will be at the end of week, quite short in time right now.

geraldus:
@snoyberg finally, I was able to do this :D I've uploaded `yesod-form-richtext` to Hackage.
I've added comment in `Nic` module, but didn't catch what do you mean by
> and the package description for yesod-form

Are you talking about description in cabal file or haddock comment in `Yesod/Form.hs`?

_By the way, I guess you should know how frequently hackage packages list is updated, I wonder how long it will take to my package get in?_

snoyberg:
Yes, I was thinking of the cabal file and/or readme

geraldus:
Ok, will do and ping you again (:

geraldus:
@snoyberg done

snoyberg:
Thanks!

geraldus:
:relaxed:

PR #1161, opened 2016-02-15
Have the yesod-auth login form use a CSRF token

### LINKED ISSUE #1159: defaultCsrfMiddleware not working with Yesod.Auth.Hardcoded
this might be realted with [this issue](https://github.com/yesodweb/yesod/issues/1151)

Although the cookie is set (and set) the handler will reject the hardcoded login (POST at `./auth/page/hardcoded/login`):

> A valid CSRF token wasn't present in HTTP headers or POST parameters. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection.

MaxGabriel:
Closes #1159

Based on reading this [StackOverflow Post](http://stackoverflow.com/questions/6412813/do-login-forms-need-tokens-against-csrf-attacks) and skimming [this paper](http://seclab.stanford.edu/websec/csrf/csrf.pdf), using CSRF protection on login forms protects against a vulnerability where an attacker submits their own username/password in the login form. Later, the user uses the real site, but doesn't realize they're logged in as the attacker. This creates vulnerabilities like:

1. If the site logs the user's activity for them (e.g. recently watched videos on YouTube, previous searches on Google), the attacker can see this information by logging in. This allows the attacker to track/spy on the user.
2. The user adds sensitive information to the account, like credit card information, the attacker can login and potentially steal that information or use it on the site.

I don't think this vulnerability applies to the `Yesod.Auth.Hardcoded` plugin because the attacker couldn't create an account of their own.

However:

* If I understand the example in `Yesod.Auth.Hardcoded`, one use case is to share one login form that works for both the Hardcoded plugin as well as normal database-backed username/password login, in which case having a CSRF token makes sense
* I don't see a downside to having the CSRF token there
* It makes the Hardcoded plugin work with the CSRF middleware

Does this sound like the right solution?

### DIFF
--- a/yesod-auth/Yesod/Auth/Hardcoded.hs
+++ b/yesod-auth/Yesod/Auth/Hardcoded.hs
@@ -162,12 +162,11 @@ authHardcoded =
     dispatch _ _ = notFound
     loginWidget toMaster = do
       request <- getRequest
-      let tokenKey = ("_token" :: Text) -- This value taken from yesod-form's postHelper. Not ideal that it's hard-coded in two places.
       [whamlet|
         $newline never
         <form method="post" action="@{toMaster loginR}">
           $maybe t <- reqToken request
-            <input type=hidden name=#{tokenKey} value=#{t}>
+            <input type=hidden name=#{defaultCsrfParamName} value=#{t}>
           <table>
             <tr>
               <th>_{Msg.UserName}
--- a/yesod-auth/Yesod/Auth/Hardcoded.hs
+++ b/yesod-auth/Yesod/Auth/Hardcoded.hs
@@ -160,10 +160,14 @@ authHardcoded =
   where
     dispatch "POST" ["login"] = postLoginR >>= sendResponse
     dispatch _ _ = notFound
-    loginWidget toMaster =
+    loginWidget toMaster = do
+      request <- getRequest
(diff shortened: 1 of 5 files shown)


yesod-auth/Yesod/Auth/Hardcoded.hs
```diff
@@ -160,10 +160,14 @@ authHardcoded =
   where
     dispatch "POST" ["login"] = postLoginR >>= sendResponse
     dispatch _ _ = notFound
-    loginWidget toMaster =
+    loginWidget toMaster = do
+      request <- getRequest
+      let tokenKey = ("_token" :: Text) -- This value taken from yesod-form's postHelper. Not ideal that it's hard-coded in two places.
```
snoyberg: I think we can use [defaultCsrfParamName](http://haddock.stackage.org/lts-5.3/yesod-core-1.4.19/Yesod-Core-Handler.html#v:defaultCsrfParamName) in both places instead.

snoyberg:
@gregwebs You've touched the CSRF code more recently than I have, any thoughts?

yesod-auth/Yesod/Auth/Hardcoded.hs
MaxGabriel: Good idea

PR #1165, opened 2016-02-18
Fixed registerHandler CSRF issue

### LINKED ISSUE #1164: Email auth plugin doesn't provide CSRF
I tried to use email authentication with a recently upgraded version of Yesod. When I tried to login or create a new user I got a CSRF token issue. I found a way to fix this and attempted to integrate a fix into the yesod-auth library. I couldn't find any tests for auth, but it did build so I'm filing a PR. I only applied this for the Register handler, if you would like I'm more than willing to do the work elsewhere as well. 

Side note: I think it'd be a good idea to pull the templates out into files that could be referenced from yesod-auth as well as overridden in a local project. Is this something that's possible/the project would be interested in?

lethjakman:
The default register handler for email authentication didn't provide a
CSRF token. I provided one by using a monadic form helper.

Should fix #1164

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -301,13 +301,13 @@ getRegisterR = registerHandler
 -- Since: 1.2.6
 defaultRegisterHandler :: YesodAuthEmail master => AuthHandler master Html
 defaultRegisterHandler = do
-    ((f,widget),e) <- lift $ runFormPost registrationForm
+    ((_,widget),enctype) <- lift $ runFormPost registrationForm
     toParentRoute <- getRouteToParent
     lift $ authLayout $ do
         setTitleI Msg.RegisterLong
         [whamlet|
             <p>_{Msg.EnterEmail}
-            <form method="post" action="@{toParentRoute registerR}">
+            <form method="post" action="@{toParentRoute registerR}" enctype=#{enctype}>
                 <div id="registerForm">
                     ^{widget}
                 <button .btn>_{Msg.Register}
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -322,7 +322,7 @@ defaultRegisterHandler = do
                 fsAttrs = []
             }
 
-            (emailRes, emailView) <- mreq textField emailSettings Nothing
+            (emailRes, emailView) <- mreq emailField emailSettings Nothing
 
             let userRes = UserForm <$> emailRes
             let widget = do
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -315,8 +315,8 @@ defaultRegisterHandler = do
     where
         registrationForm extra = do
             let emailSettings = FieldSettings {
-                fsLabel = "Email" ,
                 fsTooltip = Nothing ,
+                fsLabel = SomeMessage Msg.Email,
                 fsId = Just "email",
                 fsName = Just "email",
                 fsAttrs = []
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -315,8 +315,8 @@ defaultRegisterHandler = do
     where
         registrationForm extra = do
             let emailSettings = FieldSettings {
-                fsTooltip = Nothing ,
                 fsLabel = SomeMessage Msg.Email,
+                fsTooltip = Nothing,
                 fsId = Just "email",
                 fsName = Just "email",
                 fsAttrs = []
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -319,7 +319,7 @@ defaultRegisterHandler = do
                 fsTooltip = Nothing,
                 fsId = Just "email",
                 fsName = Just "email",
-                fsAttrs = []
+                fsAttrs = [("autofocus", "")]
             }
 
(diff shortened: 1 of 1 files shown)


yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -299,18 +301,37 @@ getRegisterR = registerHandler
...
+    ((f,widget),e) <- lift $ runFormPost registrationForm
+    toParentRoute <- getRouteToParent
     lift $ authLayout $ do
         setTitleI Msg.RegisterLong
         [whamlet|
             <p>_{Msg.EnterEmail}
-            <form method="post" action="@{tp registerR}">
+            <form method="post" action="@{toParentRoute registerR}">
                 <div id="registerForm">
-                    <label for=#{email}>_{Msg.Email}:
```
MaxGabriel: The existing code is using Msg.Email, so it'll use a localized word for Email. Can you use `Msg.Email` in your PR?

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -299,18 +301,37 @@ getRegisterR = registerHandler
...
+        registrationForm extra = do
+            let emailSettings = FieldSettings {
+                fsLabel = "Email" ,
+                fsTooltip = Nothing ,
+                fsId = Just "email",
+                fsName = Just "email",
+                fsAttrs = []
+            }
+
+            (emailRes, emailView) <- mreq textField emailSettings Nothing
```
MaxGabriel: This is using `textField`, which creates an [input of type `text`](http://hackage.haskell.org/package/yesod-form-1.4.6/docs/Yesod-Form-Fields.html#v:textField). The previous code created an input of type `email`, which has advantages like:

* Browsers can validate the email address before form submission
* CSS can select email form inputs
* Mobile browsers can display an email-specific keyboard

I think [`emailField`](http://hackage.haskell.org/package/yesod-form-1.4.6/docs/src/Yesod-Form-Fields.html#emailField) should be used so that the form is of the correct type.

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -299,18 +301,37 @@ getRegisterR = registerHandler
 -- Since: 1.2.6
 defaultRegisterHandler :: YesodAuthEmail master => AuthHandler master Html
 defaultRegisterHandler = do
-    email <- newIdent
-    tp <- getRouteToParent
+    ((f,widget),e) <- lift $ runFormPost registrationForm
+    toParentRoute <- getRouteToParent
```
MaxGabriel: Really like this rename, `toParentRoute` is much clearer than `tp`

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -299,18 +301,37 @@ getRegisterR = registerHandler
...
     lift $ authLayout $ do
         setTitleI Msg.RegisterLong
         [whamlet|
             <p>_{Msg.EnterEmail}
-            <form method="post" action="@{tp registerR}">
+            <form method="post" action="@{toParentRoute registerR}">
                 <div id="registerForm">
-                    <label for=#{email}>_{Msg.Email}:
-                    <input ##{email} type="email" name="email" width="150" autofocus>
```
MaxGabriel: The old code has this hard-coded width of 150. I that you dropped this—better to handle it with CSS, which makes it easier for the end user to change.

MaxGabriel: The old code had the `autofocus` attribute. Should that be used in the new code? I'm not sure if it's really important/good.

MaxGabriel:
This is a good PR, nice work @lethjakman. I made a few comments, mostly about maintaining some existing behavior from the old version.

yesod-auth/Yesod/Auth/Email.hs
lethjakman: Yes, I'm more than happy to, the reason I didn't is because when I set it I get:
"Couldn't match expected type ‘SomeMessage master1’
          with actual type ‘Msg.AuthMessage’"

Is there a reasonable way to convert the type?

yesod-auth/Yesod/Auth/Email.hs
lethjakman: Oooo I agree. I didn't realize that was a type. Thank you.

lethjakman:
I will make those changes and a few others when I get back. Thank you for the review! Did you see my comment on pulling the form out into a text file [here](https://github.com/yesodweb/yesod/issues/1164)? I'd like to be able to easily customize the form html wrappers without overriding a handler. However, I wasn't sure if it was possible or how one would go about doing it.

MaxGabriel:
> Did you see my comment on pulling the form out into a text file here?

I'm not sure if that would work or not, I sort of don't think it would work that way? Maybe @snoyberg could comment on that?

yesod-auth/Yesod/Auth/Email.hs
MaxGabriel: I wasn't sure either, but based on "it compiles" you just do:
```
fsLabel = SomeMessage Msg.Email,
```

yesod-auth/Yesod/Auth/Email.hs
lethjakman: I think it's a good thing to have. It's probably not necessary, but it seems like a good practice to me.

MaxGabriel:
This LGTM, but I want to give others a chance to review before merging.

lethjakman:
Since you like it I'm going to do the same thing to the other forms, would you like that in a separate PR or for me to add it here?

MaxGabriel:
@lethjakman Separate PR I guess, since I already merged this one

PR #1166, opened 2016-02-19
[WIP] improve stack detection for yesod-bin

luigy:
this came up in commercialhaskell/stack#1807

Currently when running `yesod keter` stack is only used if a `stack.yaml` file is found from where it was ran.

Expected:

if `STACK_YAML` is present in the `env` then assume `stack` 
```sh
export STACK_YAML=../stack.yaml
yesod keter   # would use stack to build (currently would use cabal)
```

and if running inside `stack exec` then use `stack`
I think is safe to add this by detecting the `STACK_EXE` env var
```sh
# also uses cabal since stack.yaml is not present in cwd
stack --stack-yaml=../stack.yaml exec -- yesod keter
```

hmm not sure if to make the local bin path relative to stack.yaml or keep the current behavior

### DIFF
--- a/yesod-bin/Keter.hs
+++ b/yesod-bin/Keter.hs
@@ -6,11 +6,12 @@ module Keter
 import Data.Yaml
 import qualified Data.HashMap.Strict as Map
 import qualified Data.Text as T
+import System.Environment (getEnvironment)
 import System.Exit
 import System.Process
 import Control.Monad
 import System.Directory hiding (findFiles)
-import Data.Maybe (mapMaybe)
+import Data.Maybe (mapMaybe,isJust,maybeToList)
 import Data.Monoid
 import System.FilePath ((</>))
 import qualified Codec.Archive.Tar as Tar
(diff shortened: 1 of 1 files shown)


snoyberg:
There's a build failure on GHC 7.8 (due to missing `<$>`), can you fix that? And thanks for the patch!

luigy:
Will do, but I haven't handled the second case, yet.  Is it ok if it also assumes to use `stack` if `STACK_EXE` is set?

snoyberg:
Yes. One other thing you may want to add is checking if stack.yaml exists in any ancestor directories.

All that said, I wonder if it's possible to rely on Stack itself for this, such as by trying to run `stack query` and see if it errors out.

luigy:
I updated the PR with your `stack query` suggestion.

Configurations passed directly to `stack` like `stack exec --stack-yaml=subdir/stack.yaml -- yesod keter` would be tricker to implement.  Setting the `STACK_YAML` env var should be used instead when the configuration is not part of direct parent.

snoyberg:
Thanks!

PR #1168, opened 2016-02-28
change OAuth Twitter ID, screen_name → user_id

kakkun61:
When using Twitter OAuth, [`credsIdent`](http://hackage.haskell.org/package/yesod-auth-1.4.12/docs/Yesod-Auth.html#v:credsIdent) returns *screen_name*. Twitter's *screen_name* is not good to use as ID because it is mutable.

The best way to identify Twitter account is to use *user_id*. *user_id* is immutable.

This change may request to change some user codes.

see: [Changing your username | Twitter Help Center](https://support.twitter.com/articles/14609) (username means *screen_name* here.)

Twitter OAuth [`credsExtra`](http://hackage.haskell.org/package/yesod-auth-1.4.12/docs/Yesod-Auth.html#v:credsExtra) is below.

```haskell
[ ("oauth_token", "foo")
, ("oauth_token_secret", "bar")
, ("user_id", "9820592")
, ("screen_name", "kakkun61")
, ("x_auth_expires", "0")
]
```

### DIFF
--- a/yesod-auth-oauth/Yesod/Auth/OAuth.hs
+++ b/yesod-auth-oauth/Yesod/Auth/OAuth.hs
@@ -4,6 +4,7 @@ module Yesod.Auth.OAuth
     ( authOAuth
     , oauthUrl
     , authTwitter
+    , authTwitterUsingUserId
     , twitterUrl
     , authTumblr
     , tumblrUrl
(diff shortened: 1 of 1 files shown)


snoyberg:
Unfortunately this approach will silently break all existing applications using Twitter. How about providing a separate function that takes a parameter indicating which value should be used as the identifier, and add a note explaining the issue to the docs?

kakkun61:
I added `twitterId` function and document comment.

```haskell
twitterId :: Yesod m => Creds m -> Text
```

snoyberg:
I don't think this will achieve what you're aiming for though, since `credsIdent` is what yesod-auth itself will use for identifying users. What I was recommending is that you make a new function that allows the user to select which field is used as the `credsIdent`, so that (1) current users aren't broken by an upgrade, but (2) new users can make a more intelligent selection of their identifier.

kakkun61:
Sorry for my confusion. Did you mean to add a function like below?

```Haskell
authTwitterUsingUserId :: YesodAuth m
                       => ByteString -- ^ Consumer Key
                       -> ByteString -- ^ Consumer Secret
                       -> AuthPlugin m
authTwitterUsingUserId key secret = authOAuth … (mkExtractCreds "twitter" "user_id")
```

snoyberg:
Yes, something along those lines. Having some comments explaining why this function should be preferred over the old one is very useful too.

kakkun61:
Please change “Since 1.4.x.x” comment to the version which this PR is release on.

snoyberg:
Merged and released, thanks!

PR #1174, opened 2016-03-06
Fixed CSRF token for login page

lethjakman:
Previously this page was using a hard coded form that would throw a CSRF token error. I also added the label back to the registration form.


MaxGabriel:
LGTM

snoyberg:
Thanks!

PR #1175, opened 2016-03-06
Minor doc patch - sendStatusJSON is since 1.4.18

sajith:
I'm sure this is trivially obvious. :-)

Commit 6a60dac introduced `sendStatusJSON` on Nov 25, 2015; yesod-core 1.4.18 was uploaded to hackage on Dec 17.


psibi:
Thanks!

PR #1176, opened 2016-03-07
Auth password handler CSRF

lethjakman:
I added a monadic form handler for the set password form. This also includes the changes from #1174. I'm not sure how you would like to handle that.


MaxGabriel:
> This also includes the changes from #1174

If you rebase against master—something like:
```
git checkout master
git pull upstream master
git checkout auth_set_password_csrf
git rebase master
git push origin auth_set_password_csrf --force
```

then your PR won't have the changes from #1174.

You can probably do this in other ways with git too, this is just my typical workflow

lethjakman:
Not a problem. Should be better now.

yesod-auth/Yesod/Auth/Message.hs
```diff
@@ -60,6 +60,7 @@ data AuthMessage =
     | ProvideIdentifier
     | SendPasswordResetEmail
     | PasswordResetPrompt
+    | CurrentPassword
```
snoyberg: This part is actually a breaking change, and will cause errors in all other translations. I'm OK including it, but we need to include at least a dummy value for the other languages supported by yesod-auth.

lethjakman: Ah ok, that makes sense. What do you want for the dummy value? I think I can get a few of translations so I'll toss those in at the same time.

snoyberg: Using the English string initially makes sense. After this gets merged in, you could send a message to the mailing list asking people to provide translations for their languages.

lethjakman: Alright, I added a few translations I received from friends as well as dummy text.

snoyberg:
Thanks!

yesod-auth/Yesod/Auth/Message.hs
lethjakman: Hey, I was looking to send an email to get translations. Is this the mailing list you were referring to? https://mail.haskell.org/mailman/listinfo/web-devel

snoyberg: It's the Yesod mailing list:

http://groups.google.com/group/yesodweb

You can find out about Yesod community resources at:

http://www.yesodweb.com/page/community

Issue #1180, opened 2016-03-09
Line of dashes throws "Invalid resource line" in route file


mrkkrp:
Since `--` starts line comments, I think it makes sense to use lines of dashes to divide sections of route files for readability, just how they can be used in Haskell. This however results in:

```
    /home/mark/projects/programs/my-project/Foundation.hs:52:21:
        Exception when trying to run compile-time code:
          Invalid resource line: ----------------------------------------------------------------------------
        Code: parseRoutesFile "config/routes"
        In the splice: $(parseRoutesFile "config/routes")
```

Is this intentional?

snoyberg:
It's the current documented behavior, but I see no harm in adding support
for comment lines. Interested in trying out a PR? Should be relatively
simple.

mrkkrp:
I think I will get to this in a few days.

snoyberg:
Resolved by #1182

PR #1182, opened 2016-03-11
Allow lines of dashes in route files

mrkkrp:
See #1180. This seems to handle it pretty well. Is there anything else I should update? (Docs?)


snoyberg:
No, this looks perfect, thanks!

snoyberg:
I especially like the minimalism of the patch and the update to the test suite :)

PR #1187, opened 2016-03-16
Multiple session messages.

sir-murray:
Sites often need to set multiple messages because of redirects causing several actions, or several unrelated warnings on one page.
For example a user just logged in so there should be a "successfully logged in" message, but they also have a "password is going to expire in 10 days" warning and an info box telling them they have a new notification.

Also adds a status to session messages, this is useful when you want to render different kinds of messages differently - e.g. errors in red and success messages in green.

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -437,17 +437,17 @@ defaultForgotPasswordHandler = do
         |]
   where
     forgotPasswordForm extra = do
-         (emailRes, emailView) <- mreq emailField emailSettings Nothing
+        (emailRes, emailView) <- mreq emailField emailSettings Nothing
     
-         let forgotPasswordRes = ForgotPasswordForm <$> emailRes
-         let widget = do
-             [whamlet|
-                 #{extra}
-                 ^{fvLabel emailView}
-                 ^{fvInput emailView}
-             |]
+        let forgotPasswordRes = ForgotPasswordForm <$> emailRes
+        let widget = do
+            [whamlet|
+                #{extra}
+                ^{fvLabel emailView}
+                ^{fvInput emailView}
+            |]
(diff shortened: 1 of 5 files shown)


yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -437,17 +437,17 @@ defaultForgotPasswordHandler = do
...
-	 return (forgotPasswordRes, widget)
+        let forgotPasswordRes = ForgotPasswordForm <$> emailRes
+        let widget = do
+            [whamlet|
+                #{extra}
+                ^{fvLabel emailView}
+                ^{fvInput emailView}
+            |]
+        return (forgotPasswordRes, widget)
+
```
sir-murray: there was a tab here, in case you're wondering what this seemingly redundant change is.

yesod-core/Yesod/Core/Handler.hs
```diff
@@ -521,30 +524,63 @@ clearUltDest = deleteSession ultDestKey
...
+    enlist = pairup . S.split W8._nul
+    pairup [] = []
+    pairup [x] = []
+    pairup (s:v:xs) = (decode s, preEscapedToHtml (decode v)) : pairup xs
+    decode = decodeUtf8With lenientDecode
+
+-- | Calls 'addMessage' with an empty status
+setMessage :: MonadHandler m => Html -> m ()
+setMessage = addMessage ""
+{-# DEPRECATED setMessage "Please use addMessage instead" #-}
```
snoyberg: I like this overhaul, but I'm not in favor of deprecating the existing functions, I think they can be useful within a single application.

sir-murray:
@snoyberg I removed the DEPRECATED directives.
I would rebase, but I think github will delete comments on diffs when I do.

snoyberg:
Thanks!

PR #1191, opened 2016-03-17
Better error provenance for stuff invoking withResponse'

### LINKED ISSUE #1185: Yesod Tests addToken could use a nicer error
Getting 

```
There was no response, you should make a request
```

Because you forgot to `GET` before using `addToken` in a POST is misery itself.

bitemyapp:
I tried to take a light touch, mostly mucking with the prime variants of functions.

Here's how it looks:

screenshot from 2016-03-17 14-35-08

This is intended to resolve #1185

### DIFF
--- a/yesod-test/yesod-test.cabal
+++ b/yesod-test/yesod-test.cabal
@@ -1,5 +1,5 @@
 name:               yesod-test
-version:            1.6.0.0
+version:            1.5.1.0
 license:            MIT
 license-file:       LICENSE
 author:             Nubis <nubis@woobiz.com.ar>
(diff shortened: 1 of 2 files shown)


codygman:
:+1: Seems like an improvement to me.

bitemyapp:
The test failure seems spurious, could be wrong.

yesod-test/yesod-test.cabal
```diff
@@ -1,5 +1,5 @@
 name:               yesod-test
-version:            1.5.0.1
+version:            1.6.0.0
```
snoyberg: What's the reason for the major version bump? It looks like this only changes internal functions.

bitemyapp: Oh so it does. I'd managed that on accident, thinking I'd changed the type signature of something that _was_ exported. Nuts. I'll fix it, sorry!

snoyberg:
Thanks Chris!

PR #1193, opened 2016-03-20
Add translation to Danish.

asjo:
As requested on the mailinglist[1] here is another translations of the messages in Yesod.Auth.

[1] http://permalink.gmane.org/gmane.comp.lang.haskell.yesod/9181


snoyberg:
Thanks!

Issue #1190, opened 2016-03-17
yesod-test does not follow redirects


mwotton:
We have some code that posts login details:

```
loginAsUser email password = do
  -- You _must_ get the page or addToken will fail!
  get BTLoginR
  statusIs 200
  request $ do
    setMethod "POST"
    addPostParam "email" email
    addPostParam "password" password
    addPostParam "remember-me" "yes"
    setUrl BTLoginR
    addToken
```

Once we called this code, we expected the token to be set, and addToken to work on the next post.
Unfortunately it doesn't: as far as i can tell, this is because addToken is looking at the previous response to scan for a CSRF token, and the 303 REDIRECT response from /login empty. If yesod-test followed the redirect, or offered some way of following the redirect, we would get the logged-in homepage, and I think everything would be kosher,  but as it doesn't, the next post fails.

Changing it to addTokenFromCookie works, but both should work, I think.
(If I have missed some obvious way to tell yesod-test to follow redirects, please forgive me.)

This is with

```
  yesodMiddleware = defaultCsrfMiddleware . sessionMiddleware . defaultYesodMiddleware
```

set,

```
✗ stack list-dependencies | grep yesod
classy-prelude-yesod 0.12.5
yesod 1.4.2
yesod-auth 1.4.12
yesod-core 1.4.19
yesod-form 1.4.6
yesod-newsfeed 1.5
yesod-persistent 1.4.0.3
yesod-static 1.5.0.3
```

snoyberg:
Following redirects in general wouldn't be good behavior for yesod-test (since we'll often want to test that a route really does return a 30X response). But having an alternate function that follows redirects would seem like a logical addition. Would that address your concern here? Would you be interested in trying out a PR for that?

mwotton:
Yep, that's what I need. Would something like "followLastRedirect" that
looks at the last response and errors out if it wasn't an la redirect be
the right way to approach it?

snoyberg:
Yeah, sounds good

MaxGabriel:
Closing; this is implemented by #1194

PR #1194, opened 2016-03-21
followRedirect as discussed in https://github.com/yesodweb/yesod/issues/1190

mwotton:
PR for #1190

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -695,12 +695,15 @@ get url = request $ do
     setUrl url
 
 -- | Follow a redirect, if the last response was a redirect.
+-- | Return Left with an error message if not a redirect
+-- | Return Right with the redirected URL if it was.
+--
 -- ==== __Examples__
 -- > get HomeR
 --
 -- > followRedirect
 followRedirect :: Yesod site
-               =>  YesodExample site (Either T.Text ())
+               =>  YesodExample site (Either T.Text T.Text)
 followRedirect = do
   mr <- getResponse
   case mr of
@@ -711,7 +714,8 @@ followRedirect = do
        else do
          case lookup "Location" (simpleHeaders r) of
           Nothing -> return $ Left "followRedirect called, but no location header set"
-          Just h -> get (TE.decodeUtf8 h) >> return (Right ())
+          Just h -> let url = TE.decodeUtf8 h in
+                     get url  >> return (Right url)
 
 -- | Sets the HTTP method used by the request.
 --
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -695,16 +695,15 @@ get url = request $ do
     setUrl url
 
 -- | Follow a redirect, if the last response was a redirect.
--- | (We consider a request a redirect if the status is
--- | 301, 302, 303, 307 or 308, and the Location header is set.)
--- |
--- | Return Left with an error message if not a redirect
--- | Return Right with the redirected URL if it was.
+-- (We consider a request a redirect if the status is
+-- 301, 302, 303, 307 or 308, and the Location header is set.)
 --
 -- ==== __Examples__
 --
 -- > get HomeR
 -- > followRedirect
+followRedirect :: Yesod site
+               =>  YesodExample site (Either T.Text T.Text) -- ^ 'Left' with an error message if not a redirect, 'Right' with the redirected URL if it was
 followRedirect :: Yesod site
                =>  YesodExample site (Either T.Text T.Text)
 followRedirect = do
@@ -713,7 +712,7 @@ followRedirect = do
    Nothing ->  return $ Left "followRedirect called, but there was no previous response, so no redirect to follow"
    Just r -> do
      if not ((H.statusCode $ simpleStatus r) `elem` [301, 302, 303, 307, 308])
-       then return $ Left  "followRedirect called, but previous request was not a redirect"
+       then return $ Left "followRedirect called, but previous request was not a redirect"
        else do
          case lookup "Location" (simpleHeaders r) of
           Nothing -> return $ Left "followRedirect called, but no location header set"
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -695,12 +695,13 @@ get url = request $ do
     setUrl url
 
 -- | Follow a redirect, if the last response was a redirect.
+-- | (We consider 301, 302, 303, 307 and 308 as redirects.)
 -- | Return Left with an error message if not a redirect
 -- | Return Right with the redirected URL if it was.
 --
 -- ==== __Examples__
--- > get HomeR
 --
+-- > get HomeR
 -- > followRedirect
 followRedirect :: Yesod site
                =>  YesodExample site (Either T.Text T.Text)
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -704,8 +704,6 @@ get url = request $ do
 -- > followRedirect
 followRedirect :: Yesod site
                =>  YesodExample site (Either T.Text T.Text) -- ^ 'Left' with an error message if not a redirect, 'Right' with the redirected URL if it was
-followRedirect :: Yesod site
-               =>  YesodExample site (Either T.Text T.Text)
 followRedirect = do
   mr <- getResponse
   case mr of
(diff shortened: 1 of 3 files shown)


yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
...
 
+-- | Follow a redirect, if the last response was a redirect.
+-- ==== __Examples__
+-- > get HomeR
+--
+-- > followRedirect
+followRedirect :: Yesod site
+               =>  YesodExample site ()
+followRedirect = do
+  mr <- getResponse
```
MaxGabriel: I think the error messages that @bitemyapp made for `withResponse` are pretty good, what do you think about using that for this function? https://github.com/yesodweb/yesod/pull/1191/files

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
     setMethod "GET"
     setUrl url
 
+-- | Follow a redirect, if the last response was a redirect.
```
MaxGabriel: It would be good if this said what qualified as a redirect (301 or 303 HTTP code + "Location" header in response).

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
...
+--
+-- > followRedirect
+followRedirect :: Yesod site
+               =>  YesodExample site ()
+followRedirect = do
+  mr <- getResponse
+  case mr of
+   Nothing ->  failure "no response, so no redirect to follow"
+   Just r -> do
+     if not ((H.statusCode $ simpleStatus r) `elem` [301,303])
```
MaxGabriel: I'm not super familiar this, but should other status codes be considered? https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection

It seems like 308 is a valid HTTP code for a redirect?

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
...
+       then failure "followRedirect called, but previous request was not a redirect"
+       else do
+         case lookup "Location" (simpleHeaders r) of
+          Nothing -> failure "No location header set"
```
MaxGabriel: It'd be good if this mentioned `followRedirect` so people know where the error came from

mwotton: good idea.

mwotton:
after trying to use this in practice, you seem to quite often want to verify that a redirect did fail, so exceptions are a bit cumbersome. refactored to return Either Text ().

yesod-test/Yesod/Test.hs
mwotton: yes, that's valid. changed to

```     if not ((H.statusCode $ simpleStatus r) `elem` [301, 302, 303, 307, 308]) ```

File: yesod-test/Yesod/Test.hs (hunk 1 above)
Comment (mwotton) [inline yesod-test/Yesod/Test.hs]: I think it's a different case. When you don't put the CSRF token in, it's almost always a mistake, and the messages are a useful way to find that out. In the redirect case (at least in my use case), you may actually want to check that a redirect has not happened, so it's easier to inspect an Either type than it is to trawl through messages.

Comment (MaxGabriel) [inline yesod-test/Yesod/Test.hs]: Ah got it; I wrote my comment before I saw the `Either` change

File: yesod-test/test/main.hs [hunk 5]
```diff
@@ -213,9 +215,28 @@ main = hspec $ do
...
+            followRedirect
+            statusIs 200
+            bodyContains "we have been successfully redirected"
 
 
+        yit "returns a Left when no redirect was returned" $ do
+            get ("/" :: Text)
+            statusIs 200
+            r <- followRedirect
+            liftIO $ assertBool "expected exception" $ isLeft r
```
MaxGabriel: Is `"expected exception"` left over from the previous implementation?

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
     setMethod "GET"
     setUrl url
 
+-- | Follow a redirect, if the last response was a redirect.
+-- ==== __Examples__
+-- > get HomeR
+--
+-- > followRedirect
+followRedirect :: Yesod site
+               =>  YesodExample site (Either T.Text ())
```
MaxGabriel: Would it be more useful to return the URL that was redirected to in the `Right` case? I don't know if that's actually useful, I just originally saw `Either Text ()` and assumed it was `Left <URL redirected to>` and `Right <no redirect>`

MaxGabriel: Either way (hah), the left/right cases should definitely be documented in the haddocks

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
     setMethod "GET"
     setUrl url
 
+-- | Follow a redirect, if the last response was a redirect.
+-- ==== __Examples__
```
MaxGabriel: There needs to be an empty haddock line before `-- ==== __Examples__`; it currently looks like this:
image

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,25 @@ get url = request $ do
...
+-- > get HomeR
+--
```
MaxGabriel: This example would make more sense if there was no empty line. Right now `get HomeR` and `followRedirect` are appearing as separate code blocks, as if they were separate examples or something

MaxGabriel:
Cool PR, think this is a great idea btw :+1:

yesod-test/test/main.hs
mwotton: yep. :) fixed, thanks.

yesod-test/Yesod/Test.hs
mwotton: yes, good call. I was trying to preserve the idea that Left was an error with some context, rather than inverting the typical semantics of Maybe, but given I have the slot I might as well pass back something useful.

yesod-test/Yesod/Test.hs
```diff
@@ -695,12 +695,15 @@ get url = request $ do
     setUrl url
 
 -- | Follow a redirect, if the last response was a redirect.
+-- | Return Left with an error message if not a redirect
+-- | Return Right with the redirected URL if it was.
```
MaxGabriel: I don't think this is doing what you wanted:

image

mwotton:
@MaxGabriel I think I've addressed all your suggestions - please let me know if I've missed anything.

yesod-test/Yesod/Test.hs
```diff
@@ -693,6 +694,29 @@ get url = request $ do
...
+-- > followRedirect
+followRedirect :: Yesod site
+               =>  YesodExample site (Either T.Text T.Text)
+followRedirect = do
+  mr <- getResponse
+  case mr of
+   Nothing ->  return $ Left "followRedirect called, but there was no previous response, so no redirect to follow"
+   Just r -> do
+     if not ((H.statusCode $ simpleStatus r) `elem` [301, 302, 303, 307, 308])
+       then return $ Left  "followRedirect called, but previous request was not a redirect"
```
MaxGabriel: Extra space here: `Left  "followRedirect`

MaxGabriel:
@mwotton Great work, one last thing though:
```
-- | Follow a redirect, if the last response was a redirect.
-- | (We consider a request a redirect if the status is
-- | 301, 302, 303, 307 or 308, and the Location header is set.)
-- |
-- | Return Left with an error message if not a redirect
-- | Return Right with the redirected URL if it was.
``` 

is being rendered like this:

image

I think you want something like this:

```
-- | Follow a redirect, if the last response was a redirect.
-- (We consider a request a redirect if the status is
-- 301, 302, 303, 307 or 308, and the Location header is set.)
--
-- ==== __Examples__
--
-- > get HomeR
-- > followRedirect
followRedirect :: Yesod site
               =>  YesodExample site (Either T.Text T.Text) -- ^ 'Left' with an error message if not a redirect, 'Right' with the redirected URL if it was
```

which renders as:

image

I also moved the `Left`/`Right` documentation location, but your way is fine too.

(You can run `stack haddock yesod-test`, then open the index.html file it prints out at the end to test this)

mwotton:
Good lord, I just can't Haddock for nuts, can I... took your version.

MaxGabriel:
@mwotton Hah, Haddock can be super finicky, and there's a lot of tricks to it. Maybe I should write a short guide about it.

MaxGabriel:
Anyway, this PR looks good to merge to me; nice job. I'll let it sit for a day to give people a chance to look at it.

mwotton:
@MaxGabriel we good? :)

MaxGabriel:
Thanks for reminding me, merged

PR #1195, opened 2016-03-22
Added imported addMessage functions to Auth

lethjakman:
These functions weren't being exported which was causing Yesod to not
compile.

### DIFF
--- a/yesod-auth/Yesod/Auth.hs
+++ b/yesod-auth/Yesod/Auth.hs
@@ -64,6 +64,7 @@ import Network.HTTP.Client (Manager, Request, withResponse, Response, BodyReader
 import qualified Network.Wai as W
 
 import Yesod.Core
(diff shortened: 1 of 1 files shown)


snoyberg:
Are you sure this is necessary? Shouldn't those functions be exported from Yesod.Core? Perhaps there's a missing lower bound on yesod-core in yesod-auth?

lethjakman:
I can play with that. What do you mean by "missing lower bound"? I couldn't get yesod to compile until I added that because it couldn't find addMessageI or addMessage. I can certainly move it if there's a better place.

snoyberg:
I think you're using different versions of yesod-core. Based on the information provided in this PR, though, it's impossible to know what's actually going on.

lethjakman:
My bad.  I did a stack clean && stack build and the issue disappeared without the change.

snoyberg:
No worries, glad it's working

PR #1196, opened 2016-03-23
Add French translation for CurrentPassword

kokotchY:
As requested on the mailinglist, here is French translation of the message CurrentPassword in Yesod.Auth.


psibi:
Thanks!

PR #1200, opened 2016-03-28
Log a warning when a CSRF error occurs

### LINKED ISSUE #1192: CSRF rejections should be logged by default
I've made a custom CSRF middleware privately that logs that a request was rejected. This is needed because the user will never see any of the log output from their handler and the body seems to be empty when this happens.

I could respin this with a parameterization that takes a logLevel and then update the scaffolds if people are open to this.

MaxGabriel:
(Closes #1192)

So, this is a breaking change because it adds the `MonadLogger` constraint to `checkCsrfHeaderOrParam`. However, I find it exceptionally unlikely that would cause issues, since this function is mainly used by higher-level middleware, is pretty new, and would probably be called from a `Handler` that already has logging capabilities.

The current log message is the same as what's returned in the response body:

>    28/Mar/2016:12:15:41 -0700 [Warn#yesod-core] A valid CSRF token wasn't present in HTTP headers or POST parameters. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection. @(yesod_LSS9gnOrLHV9K8GzIkReQP:Yesod.Core.Handler ./Yesod/Core/Handler.hs:1445:5)

* Thoughts on the error message?
* Thoughts a breaking change being ok?

cc @bitemyapp

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.4.20.1
+
+* Log a warning when a CSRF error occurs [#1200](https://github.com/yesodweb/yesod/pull/1200)
+
 ## 1.4.20
 
 * `addMessage`, `addMessageI`, and `getMessages` functions
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -252,6 +252,7 @@ import qualified Yesod.Core.TypeCache as Cache
 import qualified Data.Word8 as W8
 import qualified Data.Foldable as Fold
 import Data.Default
+import Control.Monad.Logger (MonadLogger, logWarnS)
 
 get :: MonadHandler m => m GHState
 get = liftHandlerT $ HandlerT $ I.readIORef . handlerState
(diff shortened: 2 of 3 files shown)


bitemyapp:
Include a stable URL that documents the issue perhaps? A lot people (myself included) use terminals that make it easy to open recognizable URLs.

Other than that, looks good.

MaxGabriel:
Can I ask where you were running into this error? I noticed that the default test environment in the yesod-scaffolding wouldn't log something like this and wondered if that was contributing to the problem

bitemyapp:
We had logs turned up.

Happened in dev, yesod-test has some stuff that fails at runtime if you don't make a GET before-hand. I already added logging for this.

MaxGabriel:
@bitemyapp any suggestions for a stable URL? Ideally I'd link to this: http://haddock.stackage.org/lts-5.10/yesod-core-1.4.19/Yesod-Core-Handler.html#g:28 but I don't know of a way to get a stable link to it, besides hardcoding an specific version.

bitemyapp:
@MaxGabriel stackage needs the latest version defaulting behavior Hackage has, but for LTS, it looks like.

MaxGabriel:
Related: fpco/stackage-server#107

MaxGabriel:
@snoyberg Do you think this is enough of a breaking change to warrant 1.5 or could this be snuck in as 1.4.20?

snoyberg:
To be clear, the breaking change you're referring to is the introduction of the `MonadLogger` constraint? I have no objection to that going into 1.4, since all instances of `MonadHandler` are also instances of `MonadLogger`.

MaxGabriel:
Ah my bad, didn't realize that.

snoyberg:
It's a mistake in my original design, MonadLogger should have been a superclass of MonadHandler (unless history itself defeated me here). Either way, I see no problem with this being in the 1.4 series.

PR #1201, opened 2016-03-29
Minor improve of Russian translation

geraldus:
Current translation of `ConfirmPass` message is not very good: currently used word is a noun and means "confirmation" (of something).  So, when used as label for form field this word could bring some confusion to user, e.g. what exactly should he/she confirm?  Proposed translation a bit wordy, but very certain.


snoyberg:
Thanks!

PR #1205, opened 2016-04-03
Provide CSRF token in Dummy login form

league:
Similar to #1159 which added CSRF support for Yesod.Auth.Hardcoded, this change applies it to Yesod.Auth.Dummy.


MaxGabriel:
LGTM. If you'd like you can add yourself to the changelog as 1.4.13.1 + bump the .cabal file version. Otherwise I'll merge tomorrow

MaxGabriel:
Thanks @league !

PR #1209, opened 2016-04-11
Fix addToken_ needing a trailing space to work.

maxigit:
The fix can add spaces in place where none or only one where expected.
The css parser has been modified to remove trailing or multiple spaces.
This might be a bit more lax that official CSS spec.

### DIFF
--- a/yesod-test/test/main.hs
+++ b/yesod-test/test/main.hs
@@ -140,7 +140,7 @@ main = hspec $ do
                 htmlAnyContain "p" "World"
                 htmlAnyContain "p" "Moon"
                 htmlNoneContain "p" "Sun"
-            yit "CSRF token" $ do
+            yit "finds the CSRF token by css selector" $ do
                 get ("/form" :: Text)
                 statusIs 200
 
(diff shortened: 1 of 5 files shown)


MaxGabriel:
Having the same name as you makes it very surreal to get emails about your PRs :)

maxigit:
Indeed ;-)

MaxGabriel:
This LGTM

yesod-test/test/main.hs
```diff
@@ -140,6 +140,18 @@ main = hspec $ do
                 htmlAnyContain "p" "World"
                 htmlAnyContain "p" "Moon"
                 htmlNoneContain "p" "Sun"
+            yit "CSRF token" $ do
```
MaxGabriel: All of these test descriptions are terse and yours follows that pattern, but I think it would be much clearer to mention that the test is looking up a CSRF token with a CSS query. As it is, it's hard to tell at a glance how this test is different from `yit "labels"` above

MaxGabriel:
@maxigit Do you mind adding this patch to the Changelog, and bumping the version of `yesod-test` to 1.5.1.1 in the `.cabal` file?

yesod-test/test/main.hs
maxigit: I agree with you but I'm not about what you are suggesting.

MaxGabriel: Sorry, here's what I mean. I think this line should be:

```haskell
yit "Lookup CSRF token with CSS query" $ do
```

maxigit:
done

PR #1211, opened 2016-04-14
Added yesod button and link form helper

lethjakman:
These helpers should allow you to create a simple link or button that
will include a CSRF token and allow you to send any restful request type

### DIFF
--- a/yesod-core/Yesod/Core/FormHelpers.hs
+++ b/yesod-core/Yesod/Core/FormHelpers.hs
@@ -5,8 +5,11 @@
 module Yesod.Core.FormHelpers
     (
       httpLink,
+      httpLinkParams,
       buttonLink,
-      link
+      buttonLinkParams,
+      link,
+      linkParams,
     ) where
 
 import Control.Monad.IO.Class (MonadIO)
@@ -20,8 +23,8 @@ import Data.String (IsString)
 import Text.Lucius
 import Data.Text
 
-httpLink :: (MonadIO m, MonadBaseControl IO m, MonadThrow m, ToWidget site c) => (Text -> c) -> Text -> Route site -> Text -> WidgetT site m ()
-httpLink buttonCSS methodType path text = do
+httpLinkParams :: (MonadIO m, MonadBaseControl IO m, MonadThrow m, ToWidget site c) => (Text -> c) -> [(Text, Text)] -> Text -> Route site -> Text -> WidgetT site m ()
+httpLinkParams buttonCSS params methodType path text = do
     request <- getRequest
     buttonClass <- newIdent
     toWidget $ formMethodWrapper
@@ -36,19 +39,22 @@ httpLink buttonCSS methodType path text = do
     formMethodWrapper contents = 
       [whamlet|
         $if isFormSupported
-          <form action=@{path} method=#{methodType}>
+          <form action=@?{(path, queryString)} method=#{methodType}>
             ^{contents}
         $else
           <form action=@?{(path, queryString)}>
             ^{contents}
       |]
     queryString
-      | isFormSupported = []
-      | otherwise = [("_method", methodType)]
+      | isFormSupported = params
+      | otherwise = [("_method", methodType)] ++ params
     isFormSupported = toUpper methodType == "GET" || toUpper methodType == "POST"
 
-buttonLink :: (MonadIO m, MonadBaseControl IO m, MonadThrow m) => Text -> Route site -> Text -> WidgetT site m ()
-buttonLink = httpLink style
+httpLink :: (MonadIO m, MonadBaseControl IO m, MonadThrow m, ToWidget site c) => (Text -> c) -> Text -> Route site -> Text -> WidgetT site m ()
+httpLink buttonCSS = httpLinkParams buttonCSS []
+
+buttonLinkParams :: (MonadIO m, MonadBaseControl IO m, MonadThrow m) => [(Text, Text)] -> Text -> Route site -> Text -> WidgetT site m ()
+buttonLinkParams params = httpLinkParams style params
   where
     style ident =
       [lucius|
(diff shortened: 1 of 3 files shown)


lethjakman:
This is my working example and suggestion of how I think a form helper should work. This would add the capability to generate "links" with CSRF tokens that can perform any simple restful operations. (A good example is a logout button.)

One quick comment on something I didn't know if there was a best practice for: The IsString restriction [here](https://github.com/yesodweb/yesod/pull/1211/files#diff-6586f361eb553b385efd5bd533df56d6R78) is flexible, but it seems by not restricting it to Text or String I have to manually specify that somewhere and I can not do it in the hamlet source. This means I have to define a temporary variable in the handler specifying the type, which is a bit ugly. Would it be better to restrict it to a Text or a String in this circumstance, or is there some better way that I don't know of to handle this?

yesod-core/Yesod/Core/FormHelpers.hs
```diff
@@ -0,0 +1,79 @@
...
+    HttpDELETE -> False
+
+httpLink :: (MonadIO m, MonadBaseControl IO m, MonadThrow m, ToMarkup a) => FormStyle -> HttpMethod -> Route site -> a -> WidgetT site m ()
+httpLink formType methodType path text = do
+  request <- getRequest
+  let method = (getMethodType methodType) :: String
+      standardHttpMethod = isStandardHttpMethod methodType
+  toWidget
+    [whamlet|
+      <form action=@{path} method=#{method}>
```
snoyberg: This will generate invalid HTML. Only `get` and `post` can be used as the method

yesod-core/Yesod/Core/FormHelpers.hs
```diff
@@ -0,0 +1,79 @@
...
+        $maybe token <- reqToken request
+          $if standardHttpMethod
+            <input name=_method value=#{method}>
```
snoyberg: This only works by chance, since we're using URL encoded body in this form and not multipart. Arguably multipart will never be used, but I'd still rather see the `action` value just contain the query string directly, e.g. `action=@?{path, queryString}`, where `queryString = if standardHttpMethod then [] else [("_method", method)]`

yesod-core/Yesod/Core/FormHelpers.hs
```diff
@@ -0,0 +1,79 @@
...
+import Control.Monad.Trans.Control (MonadBaseControl (..))
+import Yesod.Routes.Class
+import Control.Monad.Trans.Resource (MonadResource (..), MonadThrow (..))
+import Yesod.Core.Widget
+import Yesod.Core.Handler
+import Text.Blaze (ToMarkup)
+import Data.String (IsString)
+import Text.Lucius
+
+data HttpMethod = HttpGET | HttpPOST | HttpPATCH | HttpUPDATE | HttpDELETE
```
snoyberg: What value does this add instead of just using our more standard textual representation of a method? The HTTP spec leaves the number of HTTP methods open-ended. And this list already contains at least one notable omission: `PUT`.

yesod-core/Yesod/Core/FormHelpers.hs
```diff
@@ -0,0 +1,79 @@
...
+          <input type=hidden name=#{defaultCsrfParamName} value=#{token}>
+        $case formType
+          $of LinkForm
+              <button class=buttonLink> #{text}
+          $of ButtonForm
+              <button> #{text}
+    |]
+  toWidget
+    [lucius|
+      .buttonLink {
```
snoyberg: To avoid potential conflict with other code, it would be better to generate a unique identifier via `newIdent` and use that instead of a class name.

yesod-core/Yesod/Core/FormHelpers.hs
lethjakman: Ahhh, I see what you're saying. I did that because my understanding is that setting the method= to an invalid method defaults to GET and I thought it would make it a bit more clear as to what is happening in the html. I will default it to POST.

yesod-core/Yesod/Core/FormHelpers.hs
lethjakman: OK. That makes sense, thank you.

yesod-core/Yesod/Core/FormHelpers.hs
lethjakman: I was attempting to add type security to the HTTP methods to force it to match up with one of the routes. This initially started when I was trying to use the Yesod generated routes for a form action to deduce what the correct HTTP action would be, but I couldn't figure out a good way to do that. I can remove this and just use text. 

This also leads me to another thought: What about making an addition to the generated routes that pairs together the action and the method so you can query which method someone expected in forms and such and make sure they're always hitting the correct method and the correct path? The reason I started this change was because my GET request was being made, but failing on the server because the end route was actually a POST.

yesod-core/Yesod/Core/FormHelpers.hs
lethjakman: Sure, my thought on using .buttonLink was to allow people to override the CSS in their own applications. Will that still be possible, or is that a wanted capability?

snoyberg: If you want people to be able to set their own CSS, I'd probably have an override for setting the ID and/or class be present in the arguments to this function.

lethjakman:
What are your thoughts on these changes?

snoyberg:
I have some overall concerns:

- the names of these functions are pretty general, and don't imply exactly what they're doing, especially for something to be exposed from the main module of Yesod
- there are lots of parameters to the functions, with some types repeating, making it easy to misuse
- it's unclear that this fully captures the generality of the functionality

It may make more sense for this functionality to start in a separate package and, after it had stabilized, consider moving it into yesod-core proper.

lethjakman:
Sure! Thank you for taking a look at my idea. I appreciate the input a lot. I'll post a link to the package when I get some time. If you have any suggestions on articles that may help me improve the design it would be appreciated.

StevenXL:
@lethjakman Hi. I'm closing the PR due to `1` age of the PR and `2` the discussion leads me to believe you will make a separate PR in the future. Thanks!

PR #1213, opened 2016-04-15
Fixes for ghc-8.0 and persistent 2.5

erikd:


erikd:
This failed in travis-ci because persistent 2.5 has not yet been formally released.

snoyberg:
@gregwebs do you want to merge this after releasing persistent?

gregwebs:
2.5 was released yesterday. I just clicked the redo button on Travis

snoyberg:
Cool, thanks!

erikd:
Still not building. I'll take a look some time today.

erikd:
Building yesod with ghc-8.0 will also depend on https://github.com/prowdsponsor/nonce/issues/3

felixonmars:
yesod-form still fails to build after the patch:

```
Yesod/Form/Fields.hs:660:22:
    Could not deduce (persistent-2.5:Database.Persist.Class.PersistStore.BaseBackend
                        (YesodPersistBackend site)
                      ~ YesodPersistBackend site)
    from the context (YesodPersist site,
                      PersistEntity a,
                      PersistQuery (PersistEntityBackend a),
                      PathPiece (Key a),
                      RenderMessage site msg,
                      YesodPersistBackend site ~ PersistEntityBackend a)
      bound by the type signature for
                 optionsPersist :: (YesodPersist site, PersistEntity a,
                                    PersistQuery (PersistEntityBackend a), PathPiece (Key a),
                                    RenderMessage site msg,
                                    YesodPersistBackend site ~ PersistEntityBackend a) =>
                                   [Filter a]
                                   -> [SelectOpt a]
                                   -> (a -> msg)
                                   -> HandlerT site IO (OptionList (Entity a))
      at Yesod/Form/Fields.hs:(648,19)-(657,58)
    Relevant bindings include
      optionsPersist :: [Filter a]
                        -> [SelectOpt a]
                        -> (a -> msg)
                        -> HandlerT site IO (OptionList (Entity a))
        (bound at Yesod/Form/Fields.hs:658:1)
    In the second argument of `($)', namely `selectList filts ords'
    In a stmt of a 'do' block: pairs <- runDB $ selectList filts ords
    In the second argument of `($)', namely
      `do { mr <- getMessageRender;
            pairs <- runDB $ selectList filts ords;
            return
            $ map
                (\ (Entity key value)
                   -> Option
                        {optionDisplay = mr (toDisplay value),
                         optionInternalValue = Entity key value,
                         optionExternalValue = toPathPiece key})
                pairs }'

Yesod/Form/Fields.hs:686:22:
    Could not deduce (persistent-2.5:Database.Persist.Class.PersistStore.BaseBackend
                        (YesodPersistBackend site)
                      ~ YesodPersistBackend site)
    from the context (YesodPersist site,
                      PersistEntity a,
                      PersistQuery (PersistEntityBackend a),
                      PathPiece (Key a),
                      RenderMessage site msg,
                      YesodPersistBackend site ~ PersistEntityBackend a)
      bound by the type signature for
                 optionsPersistKey :: (YesodPersist site, PersistEntity a,
                                       PersistQuery (PersistEntityBackend a), PathPiece (Key a),
                                       RenderMessage site msg,
                                       YesodPersistBackend site ~ PersistEntityBackend a) =>
                                      [Filter a]
                                      -> [SelectOpt a]
                                      -> (a -> msg)
                                      -> HandlerT site IO (OptionList (Key a))
      at Yesod/Form/Fields.hs:(672,6)-(682,42)
    Relevant bindings include
      optionsPersistKey :: [Filter a]
                           -> [SelectOpt a]
                           -> (a -> msg)
                           -> HandlerT site IO (OptionList (Key a))
        (bound at Yesod/Form/Fields.hs:684:1)
    In the second argument of `($)', namely `selectList filts ords'
    In a stmt of a 'do' block: pairs <- runDB $ selectList filts ords
    In the second argument of `($)', namely
      `do { mr <- getMessageRender;
            pairs <- runDB $ selectList filts ords;
            return
            $ map
                (\ (Entity key value)
                   -> Option
                        {optionDisplay = mr (toDisplay value), optionInternalValue = key,
                         optionExternalValue = toPathPiece key})
                pairs }'
```

erikd:
I have a fix for `yesod-form` and now working on `yesod-auth`.

erikd:
The travis-ci failure seems to be stack related. I'm still building with cabal.

snoyberg:
This is a real error, since the lower bounds are too strict. We need support for both old and new persistent via CPP.

erikd:
@snoyberg You mean in `yesod-persistent` we need to support both old and new?

snoyberg:
Yes

erikd:
Ok, working on it.

snoyberg:
Thanks!

erikd:
@gregwebs @snoyberg If we want yesod to support both old persisitent (version < 2.5) and new persistent (version == 2.5.*) as well as supporting ghc-8.0, then we need to release a version of persistent-template that has my ghc-8.0 fixes, but without the read/write split.

Happy to do this, but just want to make sure that is what you want.

erikd:
If I checkout the `persistent-template/2.1.6` tag as a branch I can then cherry pick commit 3adc3e6df021 which then builds and tests ok with both ghc-7.10 and ghc-8.0. I'd do a pull request, but I'm not sure how to do a pull request against something that is not a branch.

erikd:
Hopefully this latest set of commits works for persistent < 2.5 as well as persistent == 2.5 as well as working for ghc-7.10 and ghc-8.0.

erikd:
Oh crap, using `<$>` breaks for ghc-7.8.

mgsloan:
Usage of liftM is fine :)

erikd:
Bah! Seem to have messed up a merge somehwere.

erikd:
Nope, commit d99de615547d messed up my WIP patches.

erikd:
Yay! That turned out to be a lot more work than I initially expected.

snoyberg:
Thanks for doing it, will review today

yesod-core/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -8,26 +8,30 @@ module Yesod.Routes.TH.RenderRoute
...
 import Data.Monoid (mconcat)
 #endif
 
 -- | Generate the constructors of a route data type.
-mkRouteCons :: [ResourceTree Type] -> ([Con], [Dec])
-mkRouteCons =
-    mconcat . map mkRouteCon
+mkRouteCons :: [ResourceTree Type] -> Q ([Con], [Dec])
+mkRouteCons rttypes =
+    mconcat <$> mapM mkRouteCon rttypes
```
snoyberg: Was this change necessary?

yesod-core/test/YesodCoreTest/Exceptions.hs
```diff
@@ -41,7 +41,7 @@ case500 :: IO ()
 case500 = runner $ do
     res <- request defaultRequest
     assertStatus 500 res
-    assertBody "FOOBAR" res
+    assertBodyContains "FOOBAR" res
```
snoyberg: It's strange that this test case needed to change, what motivated that tweak?

erikd: In ghc-8.0, `error` includes a call stack and the default `Show` instance for exceptions prints the callstack.

```> ghci -XScopedTypeVariables
GHCi, version 8.0.0.20160411: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/erikd/.ghci
ghci > import Control.Exception
ghci > catch (error "FOOBAR") (\(e :: SomeException) -> print e)
FOOBAR
CallStack (from HasCallStack):
```

yesod-core/Yesod/Routes/TH/RenderRoute.hs
erikd: Yes, `mkRouteCon` used to be pure (returning `Dec`), but now it needs to run in the `Q` monad (now returning `Q Dec`) so that `mapM conT` can be used in the `MIN_VERSION_template_haskell(2,11,0)` case.

erikd: Now that I've had my morning coffee, I can give a more coherent explanation :).

As described in the GHC 8.0 mirgation guide ( https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0 ), the `DataD` constructor has changed:

```
    - dataD :: CxtQ -> Name -> [TyVarBndr]               -> [ConQ] -> [Name] -> DecQ
    + dataD :: CxtQ -> Name -> [TyVarBndr] -> Maybe Kind -> [ConQ] -> CxtQ   -> DecQ
```

Specifiacally the last field has changed from `[Name]` to `CxtQ` which is just a type synonym for `Q Cxt`. The existing code for ghc < 8.0 already had value of type `[Name]` and the way to convert value `xs :: [Name]` to `CxtQ` is to use `mapM conT xs`.

Once we need the `Q` monad there, the whole function needs to be in the `Q` monad.

yesod-core/test/YesodCoreTest/Exceptions.hs
snoyberg: I'm still not following, sorry. There's a change here from `assertBody` to `assertBodyContains`, which is the difference between "is the body exactly X" and "does the body contain X." What's the relevance of `error` to that change?

yesod-core/Yesod/Routes/TH/RenderRoute.hs
snoyberg: Thanks for the explanation, I wasn't looking at enough of the code.

erikd: :-).

yesod-core/test/YesodCoreTest/Exceptions.hs
erikd: Hmm, i really should not respond to github issues before my morning coffee :).

Without the change from `assertBody` to `assertBodyContains` after building with ghc-8.0, the test fail with:

```
  1) Test.Exceptions 500
       uncaught exception: WaiTestFailure (WaiTestFailure "Expected response body \"FOOBAR\", but received \"FOOBAR\\nCallStack (from HasCallStack):\\n  error, called at test/YesodCoreTest/Exceptions.hs:25:12 in main:YesodCoreTest.Exceptions\"")
```

which points directly at line 25 which is:

```
getRootR = error "FOOBAR" >> return ()
```

So, `error` is relevant, because `error` now has a `CallStack`.

snoyberg: Ah, OK, that makes sense. I'm on board with that change then. Merging, thank you for all the work on this!

Issue #1218, opened 2016-04-20
'yesod keter' tries to use 'stack' even when it shouldn't


bermanjosh:
With yesod-bin 1.4.18, running `yesod keter` tries to use stack (as opposed to cabal) even when it shouldn't (there's no `STACK_EXE` or `STACK_YAML` environment variable, and running `stack` command returns: "-bash: stack: command not found"). 

The specific error yesod throws is: 
> yesod: stack: readProcessWithExitCode: runInteractiveProcess: exec: does not exist (No such file or > directory)

I've reproduced on both Ubuntu and CentOS.

bermanjosh:
Just talking to myself:
Could be that
`            (ec,_,_) <- readProcessWithExitCode "stack" ["query"] ""` [here](https://github.com/yesodweb/yesod/blob/master/yesod-bin/Keter.hs#L82)
is unhandled.

PR #1219, opened 2016-04-20
error handling when checking for stack binary

bermanjosh:
For #1218

### DIFF
--- a/yesod-bin/Keter.hs
+++ b/yesod-bin/Keter.hs
@@ -79,8 +79,8 @@ keter cabal noBuild noCopyTo buildArgs = do
 
     unless noBuild $ do
         stackQueryRunSuccess <- do
-            (ec,_,_) <- readProcessWithExitCode "stack" ["query"] ""
-            return (ec == ExitSuccess)
+            eres <- try $ readProcessWithExitCode "stack" ["query"] "" :: IO (Either SomeException (ExitCode, String, String))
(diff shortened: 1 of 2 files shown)


yesod-bin/Keter.hs
```diff
@@ -79,8 +79,8 @@ keter cabal noBuild noCopyTo buildArgs = do
 
     unless noBuild $ do
         stackQueryRunSuccess <- do
-            (ec,_,_) <- readProcessWithExitCode "stack" ["query"] ""
-            return (ec == ExitSuccess)
+            eres <- try $ readProcessWithExitCode "stack" ["query"] "" :: IO (Either SomeException (ExitCode, String, String))
```
snoyberg: Catching all exceptions (via `SomeException`) is dangerous, as it can swallow up async exceptions. Does catching just `IOException` fix this?

yesod-bin/yesod-bin.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-bin
-version:         1.4.18
+version:         1.4.19
```
snoyberg: It would be better to make a patch version bump (1.4.18.1)

bermanjosh:
@snoyberg You're right, I'll double-check the exception type and make those changes tonight / tomorrow morning.

snoyberg:
Awesome, thanks!

bermanjosh:
@snoyberg Done

PR #1226, opened 2016-05-01
Add laxSameSiteSessions and strictSameSiteSessions

bobjflong:
Adds the ability to mark the Yesod session with SameSite options.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -21,3 +21,4 @@ extra-deps:
 - yaml-0.8.17
 - nonce-1.0.2
 - persistent-2.5
+- cookie-0.4.2
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -391,7 +391,7 @@ strictSameSiteSessions = sameSiteSession sameSiteStrict
 sameSiteSession :: SameSiteOption -> IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
 sameSiteSession s = (fmap . fmap) secureSessionCookies
   where
-    sameSite cookie = cookie { setCookieSameSite = (Just s) }
+    sameSite cookie = cookie { setCookieSameSite = Just s }
     secureSessionCookies = customizeSessionCookies sameSite
 
 -- | Apply a Strict-Transport-Security header with the specified timeout to
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -49,7 +49,7 @@ import qualified Text.Blaze.Html5                   as TBH
 import           Text.Hamlet
 import           Text.Julius
 import qualified Web.ClientSession                  as CS
-import           Web.Cookie                         (parseCookies)
+import           Web.Cookie                         (parseCookies, sameSiteLax, sameSiteStrict, SameSiteOption)
 import           Web.Cookie                         (SetCookie (..))
 import           Yesod.Core.Types
 import           Yesod.Core.Internal.Session
@@ -365,6 +365,22 @@ sslOnlySessions = (fmap . fmap) secureSessionCookies
   where
     setSecureBit cookie = cookie { setCookieSecure = True }
     secureSessionCookies = customizeSessionCookies setSecureBit
+
+-- | Helps defend against CSRF attacks by setting the SameSite attribute on
+-- session cookies to "Lax".
+laxSameSiteSessions :: IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
+laxSameSiteSessions = sameSiteSession sameSiteLax
+
+-- | Helps defend against CSRF attacks by setting the SameSite attribute on
+-- session cookies to "Strict".
+strictSameSiteSessions :: IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
+strictSameSiteSessions = sameSiteSession sameSiteStrict
+
+sameSiteSession :: SameSiteOption -> IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
+sameSiteSession s = (fmap . fmap) secureSessionCookies
+  where
+    sameSite cookie = cookie { setCookieSameSite = (pure s) }
+    secureSessionCookies = customizeSessionCookies sameSite
 
 -- | Apply a Strict-Transport-Security header with the specified timeout to
 -- all responses so that browsers will rewrite all http links to https
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -391,7 +391,7 @@ strictSameSiteSessions = sameSiteSession sameSiteStrict
 sameSiteSession :: SameSiteOption -> IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
 sameSiteSession s = (fmap . fmap) secureSessionCookies
   where
-    sameSite cookie = cookie { setCookieSameSite = (pure s) }
+    sameSite cookie = cookie { setCookieSameSite = (Just s) }
     secureSessionCookies = customizeSessionCookies sameSite
 
 -- | Apply a Strict-Transport-Security header with the specified timeout to
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -368,11 +368,15 @@ sslOnlySessions = (fmap . fmap) secureSessionCookies
 
 -- | Helps defend against CSRF attacks by setting the SameSite attribute on
 -- session cookies to "Lax".
+--
+-- Since 1.4.21
 laxSameSiteSessions :: IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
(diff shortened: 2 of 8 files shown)


stack.yaml
```diff
@@ -21,3 +21,4 @@ extra-deps:
 - yaml-0.8.17
 - nonce-1.0.2
 - persistent-2.5
+- cookie-0.4.2
```
MaxGabriel: The [current `cookie` package constraint is `>= 0.4.1`](https://github.com/yesodweb/yesod/blob/465210d131ab7a82807e1a0fd2a02d3c0063d025/yesod-core/yesod-core.cabal#L45), but the SameSite options were added in 0.4.2. Can you bump the `cookie` constraint in the cabal file to handle this?

yesod-core/Yesod/Core/Class/Yesod.hs
```diff
@@ -366,6 +366,22 @@ sslOnlySessions = (fmap . fmap) secureSessionCookies
     setSecureBit cookie = cookie { setCookieSecure = True }
     secureSessionCookies = customizeSessionCookies setSecureBit
 
+-- | Helps defend against CSRF attacks by setting the SameSite attribute on
+-- session cookies to "Lax".
+laxSameSiteSessions :: IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
+laxSameSiteSessions = sameSiteSession sameSiteLax
+
+-- | Helps defend against CSRF attacks by setting the SameSite attribute on
+-- session cookies to "Strict".
```
MaxGabriel: Can you add `@since 1.4.21` or `-- Since 1.4.21` to the new APIs?

stack.yaml
bobjflong: Done, but without this update the resolver fails with:

```
Failure when adding dependencies:    
      cookie: needed (>=0.4.2 && <0.5), 0.4.1.6 found (latest applicable is 0.4.2)
    needed for package: yesod-core-1.4.20.2
```

yesod-core/Yesod/Core/Class/Yesod.hs
bobjflong: done!

yesod-core/Yesod/Core/Class/Yesod.hs
```diff
@@ -366,6 +366,22 @@ sslOnlySessions = (fmap . fmap) secureSessionCookies
     setSecureBit cookie = cookie { setCookieSecure = True }
     secureSessionCookies = customizeSessionCookies setSecureBit
 
+-- | Helps defend against CSRF attacks by setting the SameSite attribute on
+-- session cookies to "Lax".
```
MaxGabriel: I think this documentation would be more helpful if it briefly explained what this attribute does, and/or linked to an external resource with a good explanation, similar to `sslOnlySessions`'s documentation. That would help users who are trying to figure out how Yesod protects against CSRF attacks, but aren't familiar with this specific method of prevention. What do you think of something like this?

```
-- | Helps defend against CSRF attacks by setting the SameSite attribute on
-- session cookies to "Lax". See the <https://tools.ietf.org/html/draft-west-first-party-cookies IETF Draft for details>.
```

or more detailed:

```
-- | Helps defend against CSRF attacks by setting the SameSite attribute on
-- session cookies to "Lax". This will cause browsers to not include your session cookie
-- when a write request is made from www.attacker.com to your site. HTTP requests using
-- @GET@, @HEAD@, @OPTIONS@ or @TRACE@ aren't affected by this setting.
-- 
-- See the <https://tools.ietf.org/html/draft-west-first-party-cookies IETF Draft for Details>.
-- 
-- /Warning:/ SameSession cookies were introduced mid-2016, so many browsers won't support them.
-- You should use SameSite session cookies in combination with normal CSRF prevention methods.
laxSameSiteSessions :: ...
```

```
-- | Helps defend against CSRF attacks by setting the SameSite attribute on
-- session cookies to "Strict". The "Strict" option functions the same as the "Lax" option,
-- but also removes cookies for @GET@, @HEAD@, @OPTIONS@ and @TRACE@ HTTP methods.
strictSameSiteSessions :: ...
```

Caveat: I only learned about the SameSite cookie attribute when you opened this PR, so you probably know more about what this documentation should say

bobjflong: I see what you are saying. As the spec is still evolving it's best if people do up to date research when considering whether or not to use this option. I'll think about how to clearly document that.

yesod-core/Yesod/Core/Class/Yesod.hs
bobjflong: I've been thinking - maybe this functionality should be a plugin as opposed to in core? I think this because the spec is not quite stable yet (there was some ambiguity in this [rack](https://github.com/rack/rack/pull/1051) discussion). However for a plugin to be possible we must bump the Yesod `cookie` version, which I could do in another PR if that sounds good?

MaxGabriel: I'm happy for this to be in core. It seems like a pretty small piece of functionality and core is where the rest of the CSRF protection stuff is. I don't think potential changes to the spec are a huge deal for `yesod-core`: unless the spec changes so much that the lax/strict options are dropped completely, all of this code is pretty generic.

Even if there was a big change, personally I don't see a huge difference between a new plugin vs keeping these functions in core, and clearly marking them as unstable APIs that may need to change in accordance with the spec.

Maybe that's being too kind to potential breaking changes—I know Michael and Greg have worked really hard to not make major breaking changes. Anyone have thoughts on this?

yesod-core/Yesod/Core/Class/Yesod.hs
MaxGabriel: > As the spec is still evolving

Hm, ok. Here are some key things that might be worth covering:

* Warn that browser support is recent/forthcoming. Otherwise, people might use SameSite cookies and they'll be confused why it isn't working for them, or it might work for them and they won't realize many of their users aren't protected. A link to caniuse.com would be great, but they don't [support it yet, unfortunately](https://github.com/Fyrd/caniuse/issues/16130).
* Warn that the yesod API may need to change, since the spec is still in flux, if you think that might need to happen
* Suggest checking the latest draft of the spec. Is this URL https://tools.ietf.org/html/draft-west-first-party-cookies sufficient? It takes you the latest version of the spec—not sure if they intend to move it from that site later.

bobjflong: thought about this, and I think we should have:

* Link to https://datatracker.ietf.org/doc/draft-west-first-party-cookies/
* Mention that the cookie option is new and changeable.
* Explain briefly what the (current) difference is between Lax and Strict.

bobjflong: https://github.com/yesodweb/yesod/pull/1226/commits/9b0caaf2cf81834a1067704e7c999a82f677053a

yesod-core/Yesod/Core/Class/Yesod.hs
```diff
@@ -366,6 +366,34 @@ sslOnlySessions = (fmap . fmap) secureSessionCookies
...
+--
+-- This option is liable to change in future versions of Yesod as the spec evolves.
+-- View more information <https://datatracker.ietf.org/doc/draft-west-first-party-cookies/ here>.
+--
+-- Since 1.4.21
+strictSameSiteSessions :: IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
+strictSameSiteSessions = sameSiteSession sameSiteStrict
+
+sameSiteSession :: SameSiteOption -> IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
+sameSiteSession s = (fmap . fmap) secureSessionCookies
```
bobjflong: I'm a little concerned that this helper is now part of the API (this module does not have an export list). @MaxGabriel do you think I should factor this back into `laxSameSiteSessions` and `strictSameSiteSessions`?

MaxGabriel: `Yesod.Core.Class.Yesod` isn't listed under [`exposed-modules` in the cabal file](https://github.com/yesodweb/yesod/blob/465210d131ab7a82807e1a0fd2a02d3c0063d025/yesod-core/yesod-core.cabal#L74-L83), so I don't think `sameSiteSession` would become public API.

MaxGabriel: `Yesod.Core.Class.Yesod` isn't listed under [`exposed-modules` in the cabal file](https://github.com/yesodweb/yesod/blob/465210d131ab7a82807e1a0fd2a02d3c0063d025/yesod-core/yesod-core.cabal#L74-L83), so I don't think `sameSiteSession` would become public API.

MaxGabriel: `Yesod.Core.Class.Yesod` isn't listed under [`exposed-modules` in the cabal file](https://github.com/yesodweb/yesod/blob/465210d131ab7a82807e1a0fd2a02d3c0063d025/yesod-core/yesod-core.cabal#L74-L83), so I don't think `sameSiteSession` would become public API.

bobjflong: 👍

MaxGabriel:
This is good to merge I think

snoyberg:
Thanks! Sorry for the delayed merge

PR #1227, opened 2016-05-02
Adds cookie-0.4.2

bobjflong:
cookie-0.4.2 brings supports for the SameSite cookie attribute

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -21,3 +21,4 @@ extra-deps:
 - yaml-0.8.17
 - nonce-1.0.2
 - persistent-2.5
+- cookie-0.4.2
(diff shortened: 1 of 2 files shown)


snoyberg:
What's the motivation for making this change? yesod-core can already build with cookie 0.4.2.

bobjflong:
It builds but it won't typecheck with middleware that use the new options, because the Cookie versions mismatch:

```haskell
lax :: IO (Maybe SessionBackend) -> IO (Maybe SessionBackend)
lax s = (fmap . fmap) secureSessionCookies
  where
    sameSite cookie = cookie { setCookieSameSite = (pure sameSiteLax) }
    secureSessionCookies = customizeSessionCookies sameSite
```

snoyberg:
yesod-core is compatible with both older and new cookie packages, so closing.

PR #1237, opened 2016-06-04
yesod-core needs bytestring >= 0.10

bergmark:
```
Yesod/Core/Handler.hs:544:28:
    Not in scope: `L.toStrict'
```

I added a revision to fix the GHC 7.4 build: https://hackage.haskell.org/package/yesod-core-1.4.20.2/revisions/


snoyberg:
Thanks!

PR #1241, opened 2016-06-20
Added support for aeson's toEncoding function (>= 0.11)

lippling:
@snoyberg so here is the first draft.


lippling:
Build failed with: `The job exceeded the maxmimum time limit for jobs, and has been terminated.`

snoyberg:
I've restarted that build. Initial review of the code looked good, I'll go through more closely after the build succeeds.

snoyberg:
The build failure was just due to a slow cache upload, so this is GTG.

PR #1243, opened 2016-06-27
Cleanup of yesod-core code

mschristiansen:
I've silenced some compiler warnings and added +100 suggestions from `hlint`.

It consists entirely of trivial changes:
- remove redundant parenthesis
- remove redundant import
- fmap instead of liftM
- etc.

All test passing with GHC 8.0.1 and GHC 7.10.3. Not tested with earlier versions.

### DIFF
--- a/yesod-core/Yesod/Core/Internal/Session.hs
+++ b/yesod-core/Yesod/Core/Internal/Session.hs
@@ -61,7 +61,7 @@ clientSessionDateCacher validity = do
       , updateFreq   = 10000000 -- 10s
       }
 
-    return $! (getClientSessionDateCache, return ())
+    return (getClientSessionDateCache, return ())
   where
     getUpdated = do
       now <- getCurrentTime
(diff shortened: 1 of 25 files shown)


snoyberg:
Build fails with GHC 7.8.4

erikd:
Yeah, looks like you removed a `Functor` constraint (ghc 8.0 warns about about redundant constraints) but for ghc 7.8, `Functor` is not a super-class of `Monad` so the `Functor` constraint is needed.

yesod-core/Yesod/Core/Internal/Session.hs
```diff
@@ -63,7 +61,7 @@ clientSessionDateCacher validity = do
       , updateFreq   = 10000000 -- 10s
       }
 
-    return $! (getClientSessionDateCache, return ())
```
mschristiansen: Any concerns about removing this strict application?

mschristiansen:
Fixed the issues causing GHC 7.8.4 to fail.

yesod-core/Yesod/Core/Internal/Session.hs
snoyberg: No, that should be removed. It's probably a relic of a refactor, forcing evaluation of a tuple like this never makes sense.

snoyberg:
Thanks!

PR #1248, opened 2016-07-11
Default CSRF tokens to the root path "/"

### LINKED ISSUE #1247: CSRF failures due to duplicate XSRF-TOKEN values
screenshot from 2016-07-08 13-09-39

Presently, what you see in the screenshot will work because the XSRF-TOKEN values are the same, however if they differ, you'll get that CSRF token middleware error we all know and love.

I do not know how two different XSRF tokens were generated (no repro), but it's not clear to me why there isn't a single XSRF token for the `/` path which covers requests being made from other pages anyway.

I read this thread: https://github.com/yesodweb/yesod/issues/1016

It did not clear up for me why there should be multiple `X-XSRF-TOKEN` values in the cookies.

AFAICT, this code is responsible for the duplicates: 
https://github.com/yesodweb/yesod/blob/8bbe91cbfe68a87735f91251c91535abeb40ad9d/yesod-core/Yesod/Core/Handler.hs#L858-L859

Some snippets:

```haskell
addHeader :: MonadHandler m => Text -> Text -> m ()
addHeader a = addHeaderInternal . Header (encodeUtf8 a) . encodeUtf8

setCookie :: MonadHandler m => SetCookie -> m ()
setCookie = addHeaderInternal . AddCookie

-- | Internal use only, not to be confused with 'setHeader'.
addHeaderInternal :: MonadHandler m => Header -> m ()
addHeaderInternal = tell . Endo . (:)

tell :: MonadHandler m => Endo [Header] -> m ()
tell hs = modify $ \g -> g { ghsHeaders = ghsHeaders g `mappend` hs }

data GHState = GHState
    { ghsSession :: SessionMap
    , ghsRBC     :: Maybe RequestBodyContents
    , ghsIdent   :: Int
    , ghsCache   :: TypeMap
    , ghsCacheBy :: KeyedTypeMap
    , ghsHeaders :: Endo [Header]
    }
```

MaxGabriel:
* The default path of cookies is the current path making the request
  * e.g. an AJAX request made from http://example.com/foo/bar would be /foo
  * This causes multiple CSRF tokens to build up as you navigate a site
  * This will cause errors if the CSRF tokens have different values, and an invalid token is sent.
* Closes #1247

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,8 @@
+## 1.4.23.1
+
+* Don't allow sending multiple cookies with the same name to the client, in accordance with [RFC 6265](https://tools.ietf.org/html/rfc6265). This fixes an issue where multiple CSRF tokens were sent to the client. [#1258](https://github.com/yesodweb/yesod/pull/1258)
+* Default CSRF tokens to the root path "/", fixing an issue where multiple tokens were stored in cookies, and using the wrong one led to CSRF errors [#1248](https://github.com/yesodweb/yesod/pull/1248)
+
 ## 1.4.23
 
 * urlParamRenderOverride method for Yesod class [#1257](https://github.com/yesodweb/yesod/pull/1257)
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -491,13 +491,17 @@ csrfCheckMiddleware handler shouldCheckFn headerName paramName = do
 
 -- | Calls 'csrfSetCookieMiddleware' with the 'defaultCsrfCookieName'.
 --
+-- The cookie's path is set to @/@, making it valid for your whole website.
+--
 -- Since 1.4.14
 defaultCsrfSetCookieMiddleware :: Yesod site => HandlerT site IO res -> HandlerT site IO res
-defaultCsrfSetCookieMiddleware handler = csrfSetCookieMiddleware handler (def { setCookieName = defaultCsrfCookieName })
+defaultCsrfSetCookieMiddleware handler = setCsrfCookie >> handler
 
 -- | Takes a 'SetCookie' and overrides its value with a CSRF token, then sets the cookie. See 'setCsrfCookieWithCookie'.
 --
 -- For details, see the "AJAX CSRF protection" section of "Yesod.Core.Handler".
+--
+-- Make sure to set the 'setCookiePath' to the root path of your application, otherwise you'll generate a new CSRF token for every path of your app. If your app is run from from e.g. www.example.com\/app1, use @app1@. The vast majority of sites will just use @/@.
 --
 -- Since 1.4.14
 csrfSetCookieMiddleware :: Yesod site => HandlerT site IO res -> SetCookie -> HandlerT site IO res
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -1358,7 +1358,7 @@ stripHandlerT (HandlerT f) getSub toMaster newRoute = HandlerT $ \hd -> do
 --
 -- The form-based approach has the advantage of working for users with Javascript disabled, while adding the token to the headers with Javascript allows things like submitting JSON or binary data in AJAX requests. Yesod supports checking for a CSRF token in either the POST parameters of the form ('checkCsrfParamNamed'), the headers ('checkCsrfHeaderNamed'), or both options ('checkCsrfHeaderOrParam').
 --
--- The easiest way to check both sources is to add the 'defaultCsrfMiddleware' to your Yesod Middleware.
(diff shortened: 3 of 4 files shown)


MaxGabriel:
@bitemyapp

bitemyapp:
screenshot from 2016-07-11 00-27-18

been a web dev for 8 years, working dev for 10 and today-I-learned

bitemyapp:
No objections here since this looks like a cleaner solution than my hack.

I do have one question, how is uniqueness preserved in the cookie store? is it keyed by key+path? As long as we believe this'll be guaranteed-unique and not duplicate then :+1: from me.

MaxGabriel:
> I do have one question, how is uniqueness preserved in the cookie store? is it keyed by key+path? As long as we believe this'll be guaranteed-unique and not duplicate then 👍 from me.

At least for Yesod (not sure if Wai has something different), it looks like the cookies [are parsed directly from the HTTP text](https://github.com/yesodweb/yesod/blob/8bbe91cbfe68a87735f91251c91535abeb40ad9d/yesod-core/Yesod/Core/Internal/Request.hs#L93-L104), so you can get multiple cookies with the same name:

```haskell
req <- getRequest
traceShowM "Cookies are: "
traceShowM $ reqCookies req
```

```
"Cookies are: "
[("XSRF-TOKEN","yoxecmpgmd"),("_SESSION","1q/bVJn7lnz+oigJTv2b+NosvuXa1EJXtM7Yfg8kp7uWRVXiZeoobGLM0iKZl/KCO3bi5bmwrAapH8P5hrcfpscvoh7aUtPQ7AqCJNXMmuKMLHmGDOjjnpm42A+CTE89u8UQIAJ+f/Y="),("XSRF-TOKEN","s6fpSkuJqv")]
```

So there's nothing guaranteeing uniqueness based on name alone, but that's OK because this PR gives the CSRF tokens all the same name. The CSRF cookies only persist for a session, so the old CSRF tokens with the wrong path will go away when the browser's closed.

I don't think this really answers your question though, because I think we need to find the code that sends cookies back to the browser to make sure it accepts cookies with the same name, as long as they have a different path.

bitemyapp:
I think the tension here is between making everything that is valid HTTP expressible vs. what is sensible.

I'm not really sure what duplicate cookies of the same name are supposed to mean (multiplicity? dubious.) and most web stacks I've worked with haven't supported it.

I'd like to clear up why duplicate keys are expressible and what they mean for this particular feature if possible. Changing the default path doesn't fix this problem by itself, does this change make duplicate values of the same key going out the door impossible? Is that desirable?

MaxGabriel:
Ok, I was working off the idea that multiple cookies of the same name are HTTP expressible, so Yesod should support them. 

**My understanding of cookies:**

When a client makes a request to a server, the server can send back one or more cookies using Set-Cookie header(s) (e.g. `Set-Cookie: lang=en-US; Expires=Wed, 09 Jun 2021 10:18:14 GMT`). When the client makes a request to the server, it sends just the cookie key/value pair (e.g. `Cookie: lang=en-US`). 

The cookies persist on the client until the browser deletes them, based on their expiration settings/the user deleting them manually. The server can delete cookies early by sending back a new cookie with the same name and an expires date in the past.

**This issue:**

Clients can build up multiple cookies with the same name when they have different paths. This can happen as the user browses your site and different paths store different cookies. It can also happen when the server sends out multiple cookies with different paths in a single request, like so:

```haskell
setCookie def { setCookieName = "name", setCookieValue = "value1", setCookiePath = Just "/foo" }
setCookie def { setCookieName = "name", setCookieValue = "value2", setCookiePath = Just "/" }
```

I'm pretty sure the issue we ran into here was that as users navigated between different paths, different cookies built up, since the CSRF code only sets the cookie once per request.

As far as preventing multiple duplicate cookies from going out, RFC 6265 does have this to say:

>   Servers SHOULD NOT include more than one Set-Cookie header field in
>   the same response with the same cookie-name.  (See Section 5.2 for
>   how user agents handle this case.)

I can't find where in [Section 5.2](https://tools.ietf.org/html/rfc6265#section-5.2) they expand on that. It sort of surprises me the RFC didn't say "with the same cookie-name [and path and domain]". Anyway, do you think there's any value in preventing multiple cookies from going out if they have the same name? It seems like an edge case, and one only triggered by a user actively trying to do it.

Finally, it seems wrong to me to remove duplicate incoming cookies from APIs like `reqCookies`. The server can't prevent this happening, the best it can do is choose one of them. But that just makes it harder to tell there's an issue going on causing multiple cookies to be sent from the client. The current way (a list of key-value pairs in tuples) is fine for lookups, too.

bitemyapp:
>Anyway, do you think there's any value in preventing multiple cookies from going out if they have the same name? It seems like an edge case, and one only triggered by a user actively trying to do it.

I think the server should only send one value out of a particular name, preferring the "newest" possible value where a tiebreaker is needed. That seems more correct based on what the standard says.

I don't think I'd do anything about duplicates when it comes to _accepting_ the data though. The author of the server needs to decide how, if at all, they want to handle that _but_ most users are relying on middleware that will silently fail in weird ways (like CSRF middleware did for us) when they encounter this. YMMV.

PR #1250, opened 2016-07-14
Document recommended usage of the CSRF middleware

### LINKED ISSUE #1246: isAuthorized not working with defaultCsrfMiddleware
I found the following behaviour:

if I use 'defaultCsrfMiddleware' in Foundation.hs, then the 'isAuthorized' function is not called, or does not work as expected

what I did (LTS-6.6):

stack new my-project7 yesod-sqlite

in Foundation.hs use `defaultCsrfMiddleware` and `isAuthorized` returns always `Unauthorized`

    yesodMiddleware = defaultCsrfMiddleware
    ...
    -- Routes not requiring authentication.
    -- isAuthorized (AuthR _) _ = return Authorized
    -- isAuthorized FaviconR _ = return Authorized
    -- isAuthorized RobotsR _ = return Authorized
    -- Default to Authorized for now.
    isAuthorized _ _ = return $ Unauthorized "not authorized"

I would expect that all pages return a 403 Forbidden, but this is not the case. I can still browse all pages.
This seems to be a problem with defaultCsrfMiddleware, or do I misunderstand something?

BTW With the defaultYesodMiddleware it works fine.

thanks,
Alex.

MaxGabriel:
Closes #1246 

This PR isn't explicit about the CSRF middleware not providing the authorization checks—thoughts on if that's necessary? I mainly wanted to give out a code snippet that people could copy and paste in, and imply it's meant to be chained together with `defaultYesodMiddleware`.

I also think this PR will help significantly: https://github.com/yesodweb/yesod-scaffold/pull/127

image

### DIFF
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -454,6 +454,18 @@ csrfSetCookieMiddleware handler cookie = setCsrfCookieWithCookie cookie >> handl
 --
 -- For details, see the "AJAX CSRF protection" section of "Yesod.Core.Handler".
 --
+-- You can add this chain this middleware together with other middleware like so:
+--
+-- @
+-- 'yesodMiddleware' = 'defaultYesodMiddleware' . 'defaultCsrfMiddleware'
+-- @
+--
+-- or:
+--
+-- @
+-- 'yesodMiddleware' app = 'defaultYesodMiddleware' $ 'defaultCsrfMiddleware' $ app
+-- @
+--
 -- Since 1.4.14
 defaultCsrfMiddleware :: Yesod site => HandlerT site IO res -> HandlerT site IO res
 defaultCsrfMiddleware = defaultCsrfSetCookieMiddleware . defaultCsrfCheckMiddleware


MaxGabriel:
@agreif Do you think this would be sufficient? You're the best qualified person to answer I think

snoyberg:
LGTM, awesome!

agreif:
I just tested it, and it works fine.
Thanks for documenting it in the docs and in the scaffolding.

If I now read the "old" comment in the scaffolding that said "Some users may also want to add the defaultCsrfMiddleware..." then I understand what was meant by the "add" :) 
It was a misunderstanding on my side, I thought I should use `defaultCsrfMiddleware` instead of `defaultYesodMiddleware`. I was confused by the names... they both are `default...Middleware`, so I thought they both do the basic `isAuthorized`. For me a default function is not depending on some other function.
I personally would have understood it better if the CsrfMiddleware was not named `default` but something like `additionalCsrfMiddleware` with the usage this way: 
`defaultYesodMiddleware . additionalCsrfMiddleware`.

But the issue is solved now, and most importantly it is clearly documented.
Thanks again for the solution!

Alex.

PR #1257, opened 2016-08-09
urlParamRenderOverride method for Yesod class

ran-lasgalen:
This method replaces urlRenderOverride because the latter lacks support for query string.

While it is possible to keep API and just fix the yesodRender to add params to the result of urlRenderOverride, this solution would be incomplete in user control over query string and possibly ineffective in terms of performance. So I decided to add a new method, make that fix a default definition for it to keep thing backward compatible, and deprecate urlRenderOverride.

The fix is inside yesod-core but the patch also touches haddocks in yesod-static, replacing the name of method.


snoyberg:
Great approach!

PR #1258, opened 2016-08-10
What I did to avoid the duplicate cookie problem

bitemyapp:
Putting this up because this is what we're still using at the moment, not sure you remember that conversation or not.

From this: https://github.com/yesodweb/yesod/issues/1247

### DIFF
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -724,7 +724,11 @@ invalidArgsI msg = do
 -- | Set the cookie on the client.
 
 setCookie :: MonadHandler m => SetCookie -> m ()
-setCookie = addHeaderInternal . AddCookie
+setCookie sc = do
+  addHeaderInternal (DeleteCookie name value)
+  addHeaderInternal (AddCookie sc)
+  where name = setCookieName sc
+        value = setCookieValue sc
 
 -- | Helper function for setCookieExpires value
 getExpires :: MonadIO m
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -725,10 +725,10 @@ invalidArgsI msg = do
 
 setCookie :: MonadHandler m => SetCookie -> m ()
 setCookie sc = do
-  addHeaderInternal (DeleteCookie name value)
+  addHeaderInternal (DeleteCookie name path)
   addHeaderInternal (AddCookie sc)
   where name = setCookieName sc
-        value = setCookieValue sc
+        path = maybe "/" id (setCookiePath sc)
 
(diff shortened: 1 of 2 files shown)


snoyberg:
CC @MaxGabriel

bitemyapp:
@snoyberg  By the by, part of the reason I want this out of our `packages:` listing in our `stack.yaml` is that it seems like Stack ends up rebuilding everything in our packages list unconditionally, even stuff with `extra-dep` and that I know is cached. Any idea what might be going on there?

MaxGabriel:
I think this PR matches the behavior recommended in [RFC 6265](https://tools.ietf.org/html/rfc6265):

> Servers SHOULD NOT include more than one Set-Cookie header field in
the same response with the same cookie-name. (See Section 5.2 for
how user agents handle this case.)

And it solves the issue, so I think this should be merged, and #1248 should be merged as well

bitemyapp:
@snoyberg :+1: to what @MaxGabriel said :)

snoyberg:
Go for it if I don't get too it first. I'm on quasi vacation right now, so
I might be slower than usual.

bitemyapp:
@snoyberg I don't have commit bit, do you @MaxGabriel?

MaxGabriel:
Ok, both PRs are merged now, had to fix a changelog conflict in mine.

lloucas-imvu:
> I think this PR matches the behavior recommended in RFC 6265:
> 
> Servers SHOULD NOT include more than one Set-Cookie header field in
> the same response with the same cookie-name. (See Section 5.2 for
> how user agents handle this case.)
> And it solves the issue, so I think this should be merged, and #1248 should be merged as well

[https://github.com/yesodweb/yesod/blob/8bbe91cbfe68a87735f91251c91535abeb40ad9d/yesod-core/Yesod/Core/Internal/Response.hs#L90](https://github.com/yesodweb/yesod/blob/8bbe91cbfe68a87735f91251c91535abeb40ad9d/yesod-core/Yesod/Core/Internal/Response.hs#L90)

It seems to me that this explicitly does include 2 Set-Cookie header fields with the same name.
In fact I am seeing this behaviour after updating to yesod-core 1.4.24

> < Set-Cookie: sandboxCookie=; path=/; expires=Thu, 01-Jan-1970 00:00:00 GMT
> < Set-Cookie: sandboxCookie=12345; Path=/; Expires=Wed, 01-Jan-2020 00:00:00 GMT; Domain=.d.c
> < Set-Cookie: sandboxCookie-id=; path=/; expires=Thu, 01-Jan-1970 00:00:00 GMT
> < Set-Cookie: sandboxCookie-id=12345; Path=/; Expires=Wed, 01-Jan-2020 00:00:00 GMT; Domain=.d.c

PR #1262, opened 2016-08-17
Explain what sslOnlyMiddleware really does

chreekat:
Doc updated per
<https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security>.

I was tipped off to the discrepancy when my site worked totally fine over
http, in spite of the claim, "This middleware makes a site functionally
inaccessible over vanilla http in all standard browsers."


snoyberg:
@gregwebs @MaxGabriel I think the two of your are more familiar with this than I, can you review?

gregwebs:
Thanks!

Issue #1266, opened 2016-08-26
Yesod cache does not store nested cached action results


CthulhuDen:
I am talking about per-request cache (`Yesod.Core.Handler.cached`).

When [`cached'](https://github.com/yesodweb/yesod/blob/master/yesod-core/Yesod/Core/Handler.hs#L997) executes action which also calls `cached`, outer `cached` will override new cache entries having been created during inner `cached` invocations. I suggest merging cached types map instead of overwriting it [here](https://github.com/yesodweb/yesod/blob/master/yesod-core/Yesod/Core/Handler.hs#L1006).

snoyberg:
SGTM. Want to send a PR?

PR #1268, opened 2016-08-28
cached and cachedBy will not overwrite global state changes

CthulhuDen:
Fixes issue #1266 (only it also prevents overwriting any part of the changed state, not only cache. If actions to be cached were not supposed to change global state it should have been documented).


snoyberg:
Thanks!

PR #1271, opened 2016-09-01
Add bodyNotContains function

amitaibu:
PR adds a function to assert a text doesn't appear in the body of the response

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -395,6 +395,13 @@ bodyContains text = withResponse $ \ res ->
   liftIO $ HUnit.assertBool ("Expected body to contain " ++ text) $
     (simpleBody res) `contains` text
 
+-- | Assert the last response doesn't have the given text. The check is performed using the response
+-- body in full text form.
+bodyNotContains :: String -> YesodExample site ()
+bodyNotContains text = withResponse $ \ res ->
+  liftIO $ HUnit.assertBool ("Expected body not to contain " ++ text) $
+    not . contains (simpleBody res) text
+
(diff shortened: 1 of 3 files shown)


yesod-test/Yesod/Test.hs
```diff
@@ -395,6 +395,13 @@ bodyContains text = withResponse $ \ res ->
   liftIO $ HUnit.assertBool ("Expected body to contain " ++ text) $
     (simpleBody res) `contains` text
 
+-- | Assert the last response doesn't have the given text. The check is performed using the response
+-- body in full text form.
```
snoyberg: Can you add a `@since` comment, as well as bump the .cabal version field and add a comment to the ChangeLog.md? This looks good.

snoyberg: Oh, also need to export the identifier from the module.

snoyberg:
The build fails on Travis.

amitaibu:
Tests seem to be fixed apart of one that is [stuck](https://travis-ci.org/yesodweb/yesod/jobs/156792968#L1132). Maybe we should restart that test?

build__2106_-_yesodweb_yesod_-_travis_ci

snoyberg:
Restarted

amitaibu:
The failing environment seems like it didn't restart. Should we move it to allowed failings?

psibi:
I have restarted it again.

snoyberg:
I merged to master since all seems to be in order. We'll see if it breaks the master build too somehow.

amitaibu:
Cool, thanks folks. Excited about my first yesod core PR :)

PR #1274, opened 2016-09-02
Add an alternative to Yesod.Test that threads a mock variable throughout.

whittle:
This PR is not intended to be merged in its current state, as it contains a great deal of duplication, and something should be done about the documentation. Before I put any more effort into this, however, I would like to check if there is any interest in this addition. 

If a PR is an inconvenient venue to have such a discussion, I will happily close it and open an issue, but I did want to show that there is code to support such a feature request, however rudimentary. 

If mocking support gets added to yesod-test, I would also be interested in adding support for it to the yesod scaffolding.


snoyberg:
With such a large, new API like this, it probably makes sense to start it off in a separate package to allow it to iterate/evolve, instead of including in a core package that has a stable API already.

StevenXL:
@whittle Closing this PR due to age and since it seems like a separate package was suggested. Thank you.

PR #1278, opened 2016-09-21
Add instance of MonadHandler and MonadWidget for ExceptT

Dridus:
I noticed there aren't any lifting instances for ExceptT like there are for ErrorT. Hopefully it's uncontroversial to add them, but let me know if I should approach this with a different process.


snoyberg:
There's a build failure with GHC 7.8.4. Can you either use CPP to avoid those instances on older transformers, or use transformers-compat?

Dridus:
Ah whoops, sorry about that. I pushed a new commit which makes them conditional on `__GLASGOW_HASKELL__ >= 710`, hopefully that's acceptable.

snoyberg:
The condition should really be on the transformers package itself, not GHC.
It is possible to force different versions.

Dridus:
okay, I'll change it to `MIN_VERSION_transformers(0,4,0)`. I didn't previously realize that was possible.

Dridus:
okay, I think this latest commit addresses that. let me know if it's suitable

snoyberg:
Thanks!

PR #1282, opened 2016-09-27
yesod-bin: Make it build with latest optparse-applicative

erikd:
Also bump dependency lower bound.


snoyberg:
I'd rather keep compatibility with older versions. Would it be possible to update this with a CPP approach?

erikd:
According to Travis, the latest `optparse-applicative` breaks criterion, so maybe an upper bound makes more sense.

erikd:
I switched to a CPP fix as suggested.

There was one failure:
```
Codec.Compression.Zlib: compressed data stream format error (incorrect data check)
```
which has nothing to do with my change.

snoyberg:
Awesome, thanks!

PR #1287, opened 2016-10-22
Added Yesod.Form.I18n.Spanish

lamg:
Spanish internationalization for yesod-forms


snoyberg:
Thank you! This module needs to be listed in the cabal file, could you add it? Also, it would be good to have a version bump and change log entry.

lamg:
I think is done, sorry for the small mess.

snoyberg:
Thanks!

PR #1290, opened 2016-11-04
languages reflects setLanguage

snoyberg:


snoyberg:
I haven't heard any yays or nays on this yet, so leaving blocked for now.

paul-rouse:
Speaking for myself, it won't have any adverse effect - which hardly seems worth saying, except that if everyone thinks that way, no-one says anything!

psibi:
Agree with Paul here. But, I can say `yay` anyway. :)

snoyberg:
Exactly the feedback I needed, thanks 👍

Issue #1291, opened 2016-11-04
htmlString don't work in julius 


jairoGilC:
Hi I have the following htmlString in the `homepage.julius` that work fine:

    var i = 1;
    var html = 
        '<div  class="form-group form-inline" >' +
          '<div class="form-group" >' + 
            '<div for="documentType' + i +'"> TD </div>' +
            '<select name="documentType' + i +'" id="documentType' + i +'" data-select="documentType' + i +'">' +
              '<option value="1">1'+
              '<option value="2">2'+
              '<option value="3">3'+
              '<option value="4">4'+
              '<option value="7">7'+
            '</select>'            +
            '<spam for="documentType' + i +'" id="documentType' + i +'Spam" class="help-block error-block">Error</spam>' +
          '</div>' +   

But when add any other tag, the autogen file It goes blank, so I copy and paste the code in the autogen File and work fine so I suppose that the error is in the julius parse.

    var i = 1;
    var html = 
        '<div  class="form-group form-inline" >' +
          '<div class="form-group" >' + 
            '<div for="documentType' + i +'"> TD </div>' +
            '<select name="documentType' + i +'" id="documentType' + i +'" data-select="documentType' + i +'">' +
              '<option value="1">1'+
              '<option value="2">2'+
              '<option value="3">3'+
              '<option value="4">4'+
              '<option value="7">7'+
            '</select>'            +
            '<spam for="documentType' + i +'" id="documentType' + i +'Spam" class="help-block error-block">Error</spam>' +
          '</div>' +     
             
          '<div class="form-group" >' +
            '<div for="documentNumber' + i +'"> Numero de documento </div>' +
            '<input type=number name="documentNumber' + i +'' + i +'" id="documentNumber' + i +'" data-int="documentNumber' + i +'">' +
            ' <spam for="documentNumber' + i +'" id="documentNumber' + i +'Spam" class="help-block error-block">Error</spam>' +
          '</div>' +   
        '</div>' ;

paul-rouse:
This seems to be the minifying, not actually julius.  I have just reported an issue against language-javascript (https://github.com/erikd/language-javascript/issues/58).

You should be able to circumvent it by breaking your expression for `html` into several shorter pieces.  The problem is the complexity of the expression, and my testing suggests that the problem is the number of `+ i +` groups - ie those which imply a conversion - not simply the number of plus signs.

paul-rouse:
[...]

PR #1294, opened 2016-11-09
Don't keep partial autogen file when exception occurs

paul-rouse:
The problem is this: an (async) exception occurring while an autogen file was being written resulted in an incomplete file being left behind.  When it was checked again - even after rebuilding the code - the hash matched and the partial content was served.  The solution is simply to delete the file on any exception.

#1291 had rather odd symptoms because of this, presumably because of something like a ctrl-C while minifying was taking a ridiculous time to complete.

### DIFF
--- a/yesod/ChangeLog.md
+++ b/yesod/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.4.3.1
+
+*  Handle exceptions while writing a file in `addStaticContentExternal`
+
 ## 1.4.3
 
 * Switch to `Data.Yaml.Config`
--- a/yesod/Yesod/Default/Util.hs
+++ b/yesod/Yesod/Default/Util.hs
@@ -17,9 +17,11 @@ module Yesod.Default.Util
 import qualified Data.ByteString.Lazy as L
 import Data.Text (Text, pack, unpack)
 import Yesod.Core -- purposely using complete import so that Haddock will see addStaticContent
-import Control.Exception (onException)
 import Control.Monad (when, unless)
-import System.Directory (doesFileExist, createDirectoryIfMissing, removeFile)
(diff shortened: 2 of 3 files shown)


snoyberg:
This looks fine, but what's really funny is that I'm working on a function to do something very similar for this in conduit-extra.

paul-rouse:
Maybe a more general solution is the answer, then.

snoyberg:
I'll share the function here in a moment, but the basic idea is: a conduit version of the cautious-file package.

snoyberg:
Here's the commit

psibi:
@snoyberg I think you might have missed the commit.

snoyberg:
Doh, thanks. https://github.com/snoyberg/conduit/commit/73b5fc28f1bb3ae52095c404ff64f07ba0da92ac

paul-rouse:
Oh yes - it's nice that it handles pulling the plug as well.  I am happy to try using conduit instead in addStaticContentExternal, but give me a couple of days - I have some real work to do for the rest of today and tomorrow!

snoyberg:
No pressure, Using cautious-file is also an option here, though we already
have the conduit dependency, so that might be better. I'm just blown away
that I got the notification on your work here as I was writing the test
case.

paul-rouse:
Yes it is quit a coincidence!  I might as well use it, since you've written it!

paul-rouse:
Here it is, revised to use sinkFileCautious

snoyberg:
Awesome, I tweaked the stack.yaml a bit, merged, and released. Thanks!

paul-rouse:
Did you mean/expect it still to be in a branch on github? And sorry I didn't look harder at the tests!

snoyberg:
Good catch, I pushed to master. (Second time I've made that mistake this
week...)

Issue #1295, opened 2016-11-13
AttributeValue instance for Enctype


andrewthad:
There's a typeclass in blaze-markup called `ToValue`. I would like to add an instance for `Enctype`. The reason is that I recently wrote a package that is just a thin wrapper around the `blaze-html` elements and attributes. Having this instance would be helpful for writing code that generates forms.

andrewthad:
I've put up a PR for this. It's small and pretty straightforward, but it's always nice to have someone else looking it over.

PR #1296, opened 2016-11-13
Added a ToValue instance for Enctype

andrewthad:
Resolves https://github.com/yesodweb/yesod/issues/1295


andrewthad:
Travis errors for GHC8, but I don't think it's related to the changes I made.

snoyberg:
Thanks!

PR #1302, opened 2016-11-19
yesod-auth: Fix CSRF security vulnerability in registerHelper function

psibi:
Return a 403 status code if the csrf tokens aren't matched. This currently
affects two endpoints: During registration and during password reset
forms.

This curl request demonstrates how this can be exploited to register new
email:

curl -i --header "Accept: application/json" --request POST -F
"email=sibi@psibi.in" http://localhost:3005/auth/page/email/register

With the patch applied, it will respond with this:

{"message":"Permission Denied. A valid CSRF token wasn't present in HTTP
headers or POST parameters. Because the request could have been forged,
it's been rejected altogether. Check the Yesod.Core.Handler docs of the
yesod-core package for details on CSRF protection."}

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE QuasiQuotes             #-}
 {-# LANGUAGE Rank2Types              #-}
 {-# LANGUAGE TypeFamilies            #-}
+{-# LANGUAGE DoAndIfThenElse         #-}
 -- | A Yesod plugin for Authentication via e-mail
 --
 -- This plugin works out of the box by only setting a few methods on the type class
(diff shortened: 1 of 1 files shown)


yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -709,6 +713,9 @@ setLoginLinkKey aid = do
     now <- liftIO getCurrentTime
     setSession loginLinkKey $ TS.pack $ show (toPathPiece aid, now)
 
+csrfErrorMessage :: Text
+csrfErrorMessage = "A valid CSRF token wasn't present in HTTP headers or POST parameters. Because the request could have been forged, it's been rejected altogether. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection."
```
MaxGabriel: It's a little weird because it's not an API the user would use, but I think exporting this error message from Yesod-core is preferable to pasting it here

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -383,19 +384,22 @@ registerHelper :: YesodAuthEmail master
...
-    let eidentifier = case midentifier of
-            Nothing -> Left Msg.NoIdentifierProvided
-            Just x
-                | Just x' <- Text.Email.Validate.canonicalizeEmail (encodeUtf8 x) ->
-                    Right $ normalizeEmailAddress y $ decodeUtf8With lenientDecode x'
-                | allowUsername -> Right $ TS.strip x
-                | otherwise -> Left Msg.InvalidEmailAddress
-
-    case eidentifier of
+    csrfToken <- lookupPostParam "_token"
```
MaxGabriel: The error message used below says the the CSRF token can come from the headers or in the post parameters, but afaict the code only checks the post parameters. You could use `checkCsrfHeaderOrParam` to fix this. Also this would obviate the need to paste the CSRF error message from below

psibi: Thanks for the feedback. Indeed `checkCsrfHeaderOrParam` is the right way to do. I have pushed the code with fixes.

MaxGabriel:
This LGTM, will merge Tuesday morning absent any other comments.

MaxGabriel:
Nice fix @psibi! 👍

PR #1306, opened 2016-11-24
Add default text/plain error formatting

### LINKED ISSUE #622: No default text/plain formatters for builtin errors
Accept: text/plain.*

Does not throw useful errors for 404 etc.

text/html and application/json have these default handlers.

Is there simple boilerplate that could be added to work around this for an app using json and plaintext formats?

jFransham:
Closes #622

I refactored `defaultErrorHandler` to be a case statement so I could put the helper functions in a `where` block. If that's bad style or unnecessary change I can get rid of it. Also, I couldn't find any tests for error messages returning the correct `content-type`, so I'd rather ask about that here than try to write my own and end up reinventing the wheel (obviously if there really are no tests for that, I'll be happy to write them).

### DIFF
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -627,81 +627,102 @@ widgetToPageContent w = do
 
 -- | The default error handler for 'errorHandler'.
 defaultErrorHandler :: Yesod site => ErrorResponse -> HandlerT site IO TypedContent
-defaultErrorHandler NotFound = selectRep $ do
-    provideRep $ defaultLayout $ do
-        r <- waiRequest
-        let path' = TE.decodeUtf8With TEE.lenientDecode $ W.rawPathInfo r
-        setTitle "Not Found"
-        toWidget [hamlet|
-            <h1>Not Found
-            <p>#{path'}
-        |]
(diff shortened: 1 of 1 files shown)


yesod-core/Yesod/Core/Class/Yesod.hs
```diff
@@ -627,81 +627,102 @@ widgetToPageContent w = do
...
--- if you specify an authRoute the user will be redirected there and
--- this page will not be shown.
-defaultErrorHandler NotAuthenticated = selectRep $ do
-    provideRep $ defaultLayout $ do
-        setTitle "Not logged in"
-        toWidget [hamlet|
-            <h1>Not logged in
-            <p style="display:none;">Set the authRoute and the user will be redirected there.
-        |]
+defaultErrorHandler resp = case resp of
```
snoyberg: Would it be possible to switch back to the original function pattern matching in place of the case statement, and just move the `where` clause definitions to top level functions? It would make it much easier to see in the history what the actual change here was.

jFransham: Sure, I thought it was possible this would be a problem.

snoyberg:
Bump

StevenXL:
@snoyberg I think it would be appropriate to close this PR given that #1520 was merged.

snoyberg:
Good call

Issue #1300, opened 2016-11-18
widgetToPageContent support the option to have julius files generate a script tag with crossorigin attribute


sbditto85:
Disclosure: I'm a Haskell and Yesod noob.

It would be nice to have the script tag generated when calling widgetToPageContent have the option to have attributes or be able to at least pass a crossorigin value.

Having my scripts load on a different sub domain prevents me from seeing the full error details when using window.onerror, if setup properly having a crossorigin="anonymous" allows me to then see those error details.

if https://github.com/yesodweb/yesod/blob/master/yesod-core/Yesod/Core/Class/Yesod.hs#L573 (<script src="#{s}">) were to be changed to allow it to have a crossorigin attribute that would be awesome.

I may try and hack a PR together, but not sure I could do this in an optimal way.

Thanks!

snoyberg:
I think one of the Attrs variants of addScript, like addScriptRemoteAttrs,
will do what you're looking for. Check out:

https://www.stackage.org/lts-7.9/hoogle?q=addScript&package=yesod-core

sbditto85:
It appears that I left out some details of what I'm trying to do.

For example, using the `yesod-mysql` template from stack, if I wanted to have the script tag generated from having the `homepage.julius` have attributes how would I do it? I thought that `addScriptAttrs` and `addScriptRemoteAttrs` only worked on "other" js files? From what little I know of  the code the previously mentioned link to the `regularScriptLoad` portion in the `widgetToPageContent` function is the only way to modify that particular script tag for the defaultLayout.

Happy to be corrected otherwise :)

P.S. thank you for your hard work and wonderful framework, without which I don't think I'd be able to convince my work to let me use Haskell.

snoyberg:
You're right about that: the functions I mentioned will only work for separate scripts, not for those generated from `addJulius`. That would require a modification to yesod-core.

I'm a bit out of the loop on this though. Can you explain why you'd need to set crossorigin for these scripts? My understanding what that this is only relevant for scripts coming from a different domain name.

sbditto85:
I'll tell you a generic version of my situation.

I used `urlRenderOverride` to set the `StaticR` base URL to be `static.domain.com` and the website is served under `www.domain.com`. I have multiple JS files in the `./static/scripts/` directory that are included using `addScriptAttrs` and then I have a `.julius` file that is generating a script with some values inserted. In one of the JS scripts I have something similar to the following code:

```
window.onerror = senderror;
function senderror() {
[...]

PR #1308, opened 2016-11-26
added jsAttributes for the script tag generated by julius files

sbditto85:
My first attempt at fixing issue #1300.

I tried a number of things and this is the one I settled on. Please let me know if there is a better way of doing it as I'm still learning Haskell and Yesod.

### DIFF
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -575,7 +575,7 @@ widgetToPageContent w = do
                 ^{mkScriptTag s}
             $maybe j <- jscript
                 $maybe s <- jsLoc
-                    ^{mkDefaultScriptTag s (jsAttributes master)}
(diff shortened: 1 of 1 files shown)


yesod-core/Yesod/Core/Class/Yesod.hs
```diff
@@ -570,7 +575,7 @@ widgetToPageContent w = do
                 ^{mkScriptTag s}
             $maybe j <- jscript
                 $maybe s <- jsLoc
-                    <script src="#{s}">
+                    ^{mkDefaultScriptTag s (jsAttributes master)}
```
snoyberg: You can use the `*{..}` syntax for multiple attributes, described at: http://www.yesodweb.com/book/shakespearean-templates#shakespearean-templates_attributes

sbditto85:
Awesome i figured i was missing something. Learn something new every day :)

snoyberg:
Thanks!

PR #1318, opened 2016-12-06
Exports some internals and fix version bounds

s9gf4ult:
Would be great to have this on Hackage.

### DIFF
--- a/yesod-core/Yesod/Core/Dispatch.hs
+++ b/yesod-core/Yesod/Core/Dispatch.hs
@@ -17,6 +17,8 @@ module Yesod.Core.Dispatch
     , mkYesodSubData
     , mkYesodDispatch
     , mkYesodSubDispatch
+      -- *** Helpers
+    , getGetMaxExpires
       -- ** Path pieces
     , PathPiece (..)
     , PathMultiPiece (..)
--- a/yesod-core/Yesod/Core/Dispatch.hs
+++ b/yesod-core/Yesod/Core/Dispatch.hs
@@ -24,6 +24,7 @@ module Yesod.Core.Dispatch
       -- * Convert to WAI
     , toWaiApp
     , toWaiAppPlain
+    , toWaiAppYre
     , warp
     , warpDebug
     , warpEnv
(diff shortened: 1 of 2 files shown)


snoyberg:
What's the reason for the lower version bound bump?

yesod-core/Yesod/Core/Dispatch.hs
```diff
@@ -17,13 +17,16 @@ module Yesod.Core.Dispatch
     , mkYesodSubData
     , mkYesodDispatch
     , mkYesodSubDispatch
+      -- *** Helpers
```
snoyberg: If we're going to add new exports, they need to have Haddocks and `@since` comments. Can you add those, and do a minor version bump in the .cabal file?

s9gf4ult: Ok.

s9gf4ult:
> What's the reason for the lower version bound bump?

compilation error: 

```
./yesod/yesod-core/Yesod/Core/Widget.hs:112:27-32:
    Could not deduce (H.ToMarkup TB.Builder)
      arising from a use of ‘toHtml’
    from the context (MonadWidget m, HandlerSite m ~ site)
      bound by the type signature for
                 toWidget :: (MonadWidget m, HandlerSite m ~ site) =>
                             TB.Builder -> m ()
      at Yesod/Core/Widget.hs:112:5-12
    In the second argument of ‘(.)’, namely ‘toHtml’
    In the expression: toWidget . toHtml
    In an equation for ‘toWidget’: toWidget = toWidget . toHtml
```

http://hackage.haskell.org/package/blaze-markup-0.7.1.1/changelog

```
- 0.7.1.0 
* Relax `QuickCheck` dependency to allow 2.9 
* Add text builder instances
```

`yesod-core` does not build without text builder instances, which is added since `blaze-markup-0.7.1`.

psibi:
I remember seeing such an error in Travis in my PR yesterday: https://travis-ci.org/yesodweb/yesod/jobs/181651004#L438

yesod-core/Yesod/Core/Dispatch.hs
s9gf4ult: @snoyberg fixed

snoyberg:
Thanks!

PR #1321, opened 2016-12-12
Add insert400 and insert400_

whittle:
In yesod-persistent, add insert companion functions similar to get404, which: 

  * work like insert and insert_ when on the golden path
  * in the case of a uniqueness constraint violation, shortcircuit the handler to return a 400 Bad Request specifying the columns in the violated constraint

### DIFF
--- a/yesod-persistent/Yesod/Persist/Core.hs
+++ b/yesod-persistent/Yesod/Persist/Core.hs
@@ -168,9 +168,17 @@ getBy404 key = do
 -- | Create a new record in the database, returning an automatically
 -- created key, or raise a 400 bad request if a uniqueness constraint
 -- is violated.
-insert400 :: (MonadIO m, PersistUniqueWrite backend, PersistRecordBackend record backend)
-          => record
-          -> ReaderT backend m (Key record)
+--
+-- @since 1.4.1
+#if MIN_VERSION_persistent(2,5,0)
+insert400 :: (MonadIO m, PersistUniqueWrite backend, PersistRecordBackend val backend)
+          => val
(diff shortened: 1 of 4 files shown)


yesod-persistent/Yesod/Persist/Core.hs
```diff
@@ -163,7 +165,30 @@ getBy404 key = do
         Nothing -> notFound'
         Just res -> return res
 
+-- | Create a new record in the database, returning an automatically
+-- created key, or raise a 400 bad request if a uniqueness constraint
+-- is violated.
+insert400 :: (MonadIO m, PersistUniqueWrite backend, PersistRecordBackend record backend)
+          => record
+          -> ReaderT backend m (Key record)
```
snoyberg: Looks like the lack of CPP code as used in `getBy404` would require a `persistent >= 2.5` in the .cabal file. Ideally we'd keep support for older persistent versions. Is doing so a problem?

yesod-persistent/Yesod/Persist/Core.hs
```diff
@@ -163,7 +165,30 @@ getBy404 key = do
         Nothing -> notFound'
         Just res -> return res
 
+-- | Create a new record in the database, returning an automatically
+-- created key, or raise a 400 bad request if a uniqueness constraint
+-- is violated.
```
snoyberg: Can you do a minor version bump, update the changelog, and add `@since` comments to these haddocks?

snoyberg:
The build failure is just a Travis timeout, merging. Thanks!

PR #1340, opened 2017-01-31
Fix for bug in identifyForm with empty forms.

### LINKED ISSUE #1072: identifyForm mishandles empty forms
Looks like identifyForm simply removes fields of a form to make it missing. However, for already-empty forms, that's a noop, so they wrongly pass validation successfully.

The usecase seems pretty legitimate to me, e.g. delete actions on a particular route (which need no parameters). Of course, one can always work around it by providing a different route and handler for each form.

jprider63:
Fixes #1072

### DIFF
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -352,6 +352,7 @@ identifyForm identVal form = \fragment -> do
                     | otherwise = id
     ( res', w) <- eraseParams (form fragment')
 
+    -- Empty forms now properly return FormMissing. [#1072](https://github.com/yesodweb/yesod/issues/1072)
     let res = if missing then FormMissing else res'
     return ( res, w)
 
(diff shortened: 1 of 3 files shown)


yesod-form/Yesod/Form/Functions.hs
```diff
@@ -336,21 +336,24 @@ identifyForm identVal form = \fragment -> do
...
     -- Run the form proper (with our hidden <input>).  If the
     -- data is missing, then do not provide any params to the
     -- form, which will turn its result into FormMissing.  Also,
     -- doing this avoids having lots of fields with red errors.
     let eraseParams | missing   = local (\(_, h, l) -> (Nothing, h, l))
                     | otherwise = id
-    eraseParams (form fragment')
+    ( res', w) <- eraseParams (form fragment')
+
+    let res = if missing then FormMissing else res'
```
snoyberg: To ensure this doesn't get reverted in the future: can you add comment linking to the issue?

yesod-form/Yesod/Form/Functions.hs
```diff
@@ -62,7 +62,7 @@ import Yesod.Core
 import Yesod.Core.Handler (defaultCsrfParamName)
```
snoyberg: Can you provide a patch version bump in the cabal file, and add a ChangeLog entry about the change?

snoyberg:
Looks good, just some housekeeping requests. Thanks!

jprider63:
Done. Let me know if I should change anything else.

snoyberg:
Thanks!

snoyberg:
Thanks!

PR #1346, opened 2017-02-06
Allow for a  custom Email login widget

alx741:
I needed to have custom CSS for the Email login page, but the email login widget seems to be hard coded in yesod-auth-email and doesn't export the `defaut...Handler` function that register, forgot password and set password do (https://www.stackage.org/haddock/lts-7.19/yesod-auth-1.4.16/Yesod-Auth-Email.html#g:5).

So I added and exported a `defaultEmailLoginHandler`, that will allow the user to define a custom one, and thus change the login widget. **But** this is probably dumb as it isn't like the other functions: it doesn't produce a Handler, but a Widget and it takes the `toParent` function as argument.

Should only the names be adjusted to reflect what is actually going on (and the documentation as well) ? or is there a (sane) way to actually make it work as the other `default...Handler` functions?

Edit: This is related to this stackoverflow question: http://stackoverflow.com/questions/42015869/custom-widget-for-email-login-in-yesod?noredirect=1#comment71234149_42015869

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -298,7 +298,7 @@ class ( YesodAuth site
     -- Default: 'defaultEmailLoginHandler'.
     --
     -- @since: 1.2.6
-    emailLoginHandler :: YesodAuthEmail master => (Route Auth -> Route master) -> WidgetT master IO ()
+    emailLoginHandler :: (Route Auth -> Route site) -> WidgetT site IO ()
     emailLoginHandler = defaultEmailLoginHandler
 
 
@@ -534,7 +534,7 @@ defaultForgotPasswordHandler = do
   where
     forgotPasswordForm extra = do
         (emailRes, emailView) <- mreq emailField emailSettings Nothing
-    
+
         let forgotPasswordRes = ForgotPasswordForm <$> emailRes
         let widget = do
             [whamlet|
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -297,7 +297,7 @@ class ( YesodAuth site
     --
     -- Default: 'defaultEmailLoginHandler'.
     --
-    -- @since 1.4.11
+    -- @since 1.4.17
     emailLoginHandler :: (Route Auth -> Route site) -> WidgetT site IO ()
     emailLoginHandler = defaultEmailLoginHandler
 
@@ -360,7 +360,7 @@ getRegisterR = registerHandler
 
 -- | Default implementation of 'emailLoginHandler'.
 --
--- @since 1.2.6
+-- @since 1.4.17
 defaultEmailLoginHandler :: YesodAuthEmail master => (Route Auth -> Route master) -> WidgetT master IO ()
 defaultEmailLoginHandler toParent = do
         (widget, enctype) <- liftWidgetT $ generateFormPost loginForm
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -297,7 +297,7 @@ class ( YesodAuth site
     --
     -- Default: 'defaultEmailLoginHandler'.
     --
-    -- @since: 1.2.6
+    -- @since 1.4.11
     emailLoginHandler :: (Route Auth -> Route site) -> WidgetT site IO ()
     emailLoginHandler = defaultEmailLoginHandler
 
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -106,6 +106,7 @@ module Yesod.Auth.Email
     , loginLinkKey
     , setLoginLinkKey
      -- * Default handlers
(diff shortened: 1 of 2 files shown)


yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -346,8 +358,11 @@ authEmail =
 getRegisterR :: YesodAuthEmail master => HandlerT Auth (HandlerT master IO) Html
 getRegisterR = registerHandler
 
-emailLoginHandler :: YesodAuthEmail master => (Route Auth -> Route master) -> WidgetT master IO ()
-emailLoginHandler toParent = do
+-- | Default implementation of 'emailLoginHandler'.
+--
+-- @since 1.2.6
```
psibi: I don't think it should be from 1.2.6.

alx741: Not quite sure what to put there thought, the current tag is `1.4.10` but if this is used, it will be on `1.4.11` or the like...

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -290,6 +291,17 @@ class ( YesodAuth site
     normalizeEmailAddress :: site -> Text -> Text
     normalizeEmailAddress _ = TS.toLower
 
+    -- | Handler called to render the login page.
+    -- The default works fine, but you may want to override it in
+    -- order to have a different DOM.
+    --
+    -- Default: 'defaultEmailLoginHandler'.
+    --
+    -- @since: 1.2.6
```
psibi: This shouldn't have `:`

yesod-auth/Yesod/Auth/Email.hs
psibi: Yeah, it should be the next minor version.

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -290,6 +291,17 @@ class ( YesodAuth site
...
+    emailLoginHandler :: (Route Auth -> Route site) -> WidgetT site IO ()
+    emailLoginHandler = defaultEmailLoginHandler
+
```
psibi: Also you have to change the relevant function in `authEmail` for this to work.

alx741:
Thanks @psibi , `YesodAuthEmail` contains `emailLoginHandler` as a member function that is meant to be overwritten by the user. `authEmail` stays unchanged, as `emailLoginHandler` is now a class member function defined in terms of ` defaultEmailLoginHandler`

psibi:
@alx741 LGTM. Just to be sure, Were you able to add the custom DOM using it and was it working ?

I think we can merge this once the test passes.

alx741:
@psibi I'm trying to test this out, but a bit clueless about how to do so.

I tried to load `demo/auth/email_auth_ses_mailer.hs` into GHCi, but it errors out:

```
email_auth_ses_mailer.hs:19:18:
    Could not find module ‘Database.Persist.Sqlite’
    Perhaps you meant
      Database.Persist.Sql (from persistent-2.5@persi_AFLfH6GwRzz7PZt5tCzHRv)
      Database.Persist.Sql.Raw
      Database.Persist.Sql.Run
    Use -v to see a list of the files searched for.

email_auth_ses_mailer.hs:27:18:
    Could not find module ‘Network.Mail.Mime.SES’
    Perhaps you meant
      Network.Mail.Mime (from mime-mail-0.4.11@mimem_GfwvuUO1sCuAzWhs10QE4S)
    Use -v to see a list of the files searched for.
Failed, modules loaded: none.
```

Any pointers about how to try it out?

psibi:
I think you don't have the relevant packages in the cabal file. Have you added the package `persistent-sqlite` and `mime-mail-ses` in appropriate place ?

alx741:
Alright, I was finally able to try it out and there are good news, I was perfectly able to overwrite `emailLoginHandler` with a custom one and thus have a custom DOM. Also de tests passed so I think this is in good shape

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -290,6 +291,17 @@ class ( YesodAuth site
     normalizeEmailAddress :: site -> Text -> Text
     normalizeEmailAddress _ = TS.toLower
 
+    -- | Handler called to render the login page.
+    -- The default works fine, but you may want to override it in
+    -- order to have a different DOM.
+    --
+    -- Default: 'defaultEmailLoginHandler'.
+    --
+    -- @since 1.4.11
```
psibi: This has to be `1.4.17` since the function `emailLoginHandler` will be exported as part of the typeclass from it's next version.

yesod-auth/Yesod/Auth/Email.hs
psibi: @alx741 Again this has to be `1.4.17`

psibi:
@alx741 Awesome. I have left two comments regarding the version number used. Also, can you update the Changelog with your entry ? 

Apart from that everything seems good. I will wait for others to comment now. I will merge it in three days if nobody opposes (and provided this PR hasn't been already merged by then).

psibi:
Thanks.

PR #1347, opened 2017-02-07
Allow latest nightly

snoyberg:


sbditto85:
any reason the travis CI builds show as canceled? I wanted to try using nightly as i think it will solve one of my problems.

snoyberg:
Yes, I'm fixing this a different way, you can see it in the latest commits. This should be irrelevant to usage, all of the Yesod packages are already in the latest nightly.

sbditto85:
Please disregard my previous comment ... I was chasing a problem and figured it out

PR #1355, opened 2017-02-25
Update license to 2017

madnight:


snoyberg:
Thanks!

erikd:
Isn't that dat date the copyright date? What happens to the copyright on the work you did before this year? Are you giving that up?

I suspect that what you really want is to have a copyright date of `2012-2017`.

madnight:
@erikd git solves that problem, for example if you check out a source code version of 2012, then the copyright year should be 2012 and so on (works well if you update every year), but your solution is also very common and equally valid

psibi:
But for copyright the year that is mentioned in it should be the year on which it was released. So the year in the current Yesod LICENSE file isn't right, I think. For evolving products like software, a year range is typically used.

Reference: http://www.copyrightlaws.com/international/522/

snoyberg:
I've updated the LICENSE file with a range.

PR #1363, opened 2017-03-22
Adds curly brackets to route parser.

jprider63:
As discussed on the [google group](https://groups.google.com/forum/#!topic/yesodweb/DTDcDPT7Wps). I still need to implement the `&{}` syntax.

### DIFF
--- a/yesod-core/Yesod/Routes/Parse.hs
+++ b/yesod-core/Yesod/Routes/Parse.hs
@@ -13,7 +13,7 @@ module Yesod.Routes.Parse
     ) where
 
 import Language.Haskell.TH.Syntax
-import Data.Char (isUpper)
+import Data.Char (isUpper, isSpace)
 import Language.Haskell.TH.Quote
 import qualified System.IO as SIO
 import Yesod.Routes.TH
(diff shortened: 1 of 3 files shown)


yesod-core/Yesod/Routes/Parse.hs
```diff
@@ -102,6 +102,26 @@ resourcesFromString =
...
+-- | Splits a string by spaces, as long as the spaces are not enclosed by curly brackets (not recursive).
+splitSpaces :: String -> [String]
+splitSpaces "" = []
+splitSpaces str = 
+    let (rest, piece) = parse $ dropWhile isSpace str in
+    piece:(splitSpaces rest)
+
+    where 
+        parse :: String -> ( String, String)
+        parse ('{':s) = fmap ('{':) $ parseBracket s
```
snoyberg: Is there an advantage to retaining the surrounding braces? What about simply dropping them, and then removing the need for the `dynamicPieceFromString` below?

jprider63: I think that would be too permissive. Otherwise, lines could contain arbitrary braces in routes. For example, the following would probably parse:

    /m{}y/{route with}/{spaces} RouteR GET

I prefer the current stricter grammar that requires the braces follow `'#'` so that braces can only surround types in the route.

snoyberg: Fair enough, that makes sense.

snoyberg:
Thanks!

PR #1369, opened 2017-03-28
Added Yesod.Form.I18n.Chinese, also cleaned up chineseMessage in Yesod.Auth.Message

mingyuguo:


yesod-form/Yesod/Form/I18n/Chinese.hs
```diff
@@ -0,0 +1,26 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Yesod.Form.I18n.Chinese where
+
+import Yesod.Form.Types (FormMessage (..))
+import Data.Monoid (mappend)
+import Data.Text (Text)
+
+chineseFormMessage :: FormMessage -> Text
+chineseFormMessage (MsgInvalidInteger t) = "无效的整数: " `Data.Monoid.mappend` t
```
psibi: Just a style issue: To be consistent with the rest of the code, you can use `mappend` here also instead of `Data.Monoid.mappend`.

mingyuguo: for some reason, in all the language files (e.g., English.hs, Japanese.hs), the first line uses `Data.Monoid.mappend` while the other lines use just mappend.

snoyberg:
That's to avoid unused identifier warnings with various versions of GHC.
It's an ugly backwards compatibility hack when compiling with -Wall -Werror.

psibi:
Thanks for the PR!

yesod-form/Yesod/Form/I18n/Chinese.hs
psibi: Merged the PR. As mentioned by Michael it seems a backward compatibility thing. Thanks.

Issue #1373, opened 2017-04-07
Remove value attribute of input type=password


rkaminsk:
Hi, currently the passwordField creates an input with a value attribute (see code below). As far as I know it is really not a good idea to do this. Do yo agree that the attribute should be removed? -R

```
passwordField :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m Text
passwordField = Field
    { fieldParse = parseHelper $ Right
    , fieldView = \theId name attrs val isReq -> toWidget [hamlet|
$newline never
<input id="#{theId}" name="#{name}" *{attrs} type="password" :isReq:required="" value="#{either id id val}">
|]
    , fieldEnctype = UrlEncoded
    }
```

snoyberg:
Yes, it should be, looks like it was just a copy-paste error. Want to send
a PR?

rkaminsk:
Here you go.

PR #1374, opened 2017-04-08
remove value attribute from password field

rkaminsk:
see issue #1373

### DIFF
--- a/yesod-form/Yesod/Form/Fields.hs
+++ b/yesod-form/Yesod/Form/Fields.hs
@@ -267,9 +267,9 @@ $newline never
 passwordField :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m Text
 passwordField = Field
     { fieldParse = parseHelper $ Right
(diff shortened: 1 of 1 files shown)


yesod-form/Yesod/Form/Fields.hs
```diff
@@ -267,9 +267,9 @@ $newline never
 passwordField :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m Text
 passwordField = Field
     { fieldParse = parseHelper $ Right
-    , fieldView = \theId name attrs val isReq -> toWidget [hamlet|
+    , fieldView = \theId name attrs _val isReq -> toWidget [hamlet|
```
psibi: If we are not using `_val`, I think `_` is a better choice.

rkaminsk: In such cases I prefer `_val` over `_` to keep the name. Anyway, it has to match the style of the rest of the project. Changed this to `_`.

psibi:
Thanks!

PR #1375, opened 2017-04-11
Add assertNotEq

amitaibu:
Allow asserting values are _not_ equal, instead of doing something like `assertEq "Values are different" ("1" /= "2") True`


amitaibu:
Travis seems to be stuck..

psibi:
Thanks!

PR #1381, opened 2017-04-24
Fixing race condition in yesod-bin

psibi:
Stack build process emits line even after successful build process
which leads to the overwriting of the appPortVar with -1. This leads
it to a compile mode again. Pressing Return Key and rebuilding it
again makes it go, but that's just a workaround I have to do every now
and then to solve the actual issue.

I'm using a `MVar` based locking solution for fixing the race
condition introduced.

I have been using this patch for around 3 to 4 days in our workplace and we haven't seen any surprising behavior so far.

### DIFF
--- a/yesod-bin/Devel.hs
+++ b/yesod-bin/Devel.hs
@@ -231,26 +231,24 @@ stackSuccessString = "ExitSuccess"
 stackFailureString :: ByteString
 stackFailureString = "ExitFailure"
 
-data BuildOutput = Started
-                   deriving (Show, Eq, Ord)
-
-makeEmptyMVar :: MVar a -> IO ()
-makeEmptyMVar mvar = do
-  isEmpty <- isEmptyMVar mvar
-  case isEmpty of
-    True -> return ()
-    False -> takeMVar mvar >> return ()
-
-updateAppPort :: ByteString -> MVar (BuildOutput) -> TVar Int -> IO ()
-updateAppPort bs mvar appPortVar = do
-  isEmpty <- isEmptyMVar mvar
-  let hasEnd = isInfixOf stackFailureString bs || isInfixOf stackSuccessString bs
-  case (isEmpty,hasEnd) of
-    (True,False) -> do
-      putMVar mvar Started
-      atomically $ writeTVar appPortVar (-1 :: Int)
-    (_,False) -> return ()
-    (_,True) -> makeEmptyMVar mvar
+-- We need updateAppPort logic to prevent a race condition.
+-- See https://github.com/yesodweb/yesod/issues/1380
+updateAppPort :: ByteString -> TVar Bool -- ^ Bool to indicate if the
+                                         -- output from stack has
+                                         -- started. False indicate
+                                         -- that it hasn't started
+                                         -- yet.
+              -> TVar Int -> IO ()
+updateAppPort bs buildStarted appPortVar = do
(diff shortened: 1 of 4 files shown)


yesod-bin/Devel.hs
```diff
@@ -222,6 +225,33 @@ checkDevelFile =
...
+
+stackFailureString :: ByteString
+stackFailureString = "ExitFailure"
+
+data BuildOutput = Started
+                   deriving (Show, Eq, Ord)
+
+makeEmptyMVar :: MVar a -> IO ()
+makeEmptyMVar mvar = do
+  isEmpty <- isEmptyMVar mvar
```
snoyberg: It's safer to use tryTakeMVar

yesod-bin/Devel.hs
```diff
@@ -222,6 +225,33 @@ checkDevelFile =
...
+  case isEmpty of
+    True -> return ()
+    False -> takeMVar mvar >> return ()
+
+updateAppPort :: ByteString -> MVar (BuildOutput) -> TVar Int -> IO ()
+updateAppPort bs mvar appPortVar = do
+  isEmpty <- isEmptyMVar mvar
```
snoyberg: It seems like the logic here would be easier to follow as a TVar Bool instead of MVar Build output.

psibi:
@snoyberg Thanks for the PR, I have updated the PR. I have also added `DEBUG` flag to cabal file which IMO helps during development of `yesod-bin`. I will remove it if you feel it's not needed.

yesod-bin/Devel.hs
```diff
@@ -222,6 +227,42 @@ checkDevelFile =
...
+-- We need updateAppPort logic to prevent a race condition.
+-- See https://github.com/yesodweb/yesod/issues/1380
+updateAppPort :: ByteString -> TVar Bool -- ^ Bool to indicate if the
+                                         -- output from stack has
+                                         -- started. False indicate
+                                         -- that it hasn't started
+                                         -- yet.
+              -> TVar Int -> IO ()
+updateAppPort bs buildStarted appPortVar = do
+  hasStarted <- readTVarIO buildStarted
```
snoyberg: This looks like a race condition to me, at least in principle. Why not turn the entire function into a single `STM` block and perform the read and modifications atomically?

snoyberg:
I'm not opposed to some debugging output, but I'd rather see it happen with a `--verbose` flag than CPP. I'm indifferent on whether we keep the CPP in for now though.

psibi:
Thanks for the review. Indeed making it as a single STM block is better. I have updated the PR.

yesod/yesod.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod
-version:         1.4.5
+version:         1.4.6
```
snoyberg: Not sure if this was here before or not, is it supposed to be part of this PR?

snoyberg:
LGTM, thanks for the update. Merge when ready 👍

yesod/yesod.cabal
psibi: Sorry, that was unintended. Thanks for the catch, have reverted this.

psibi:
Thanks, will merge it once that Travis CI passes.

psibi:
I would love to see a new release of this version in Hackage.

PR #1387, opened 2017-05-04
Fix Haddock formatting for staticFilesList

steshaw:


psibi:
Did you check if it works ? I think you have to put `@` before and after double quotes (along with `\`) to make it work.

steshaw:
I did check that it works.

Here's what it used to look like (from stackage.org):
screenshot 2017-05-05 10 04 42

Here's what it looks like after my change (locally):
screenshot 2017-05-05 09 57 03

You don't need all that crazy escaping with ["bird track" code blocks](https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810780208).

psibi:
Thanks!

PR #1394, opened 2017-05-11
Add WaiSubsiteWithAuth

mchaver:
This newtype and instances will allow the WaiSubsite to pick up isAuthorized from the parent site. I followed the code for Static subsites. I am guessing it makes sense to have both `WaiSubsite` and `WaiSubsiteWithAuth` because it is possible that we want to have a WaiSubsite that ignores the parents middleware. Maybe there is a better name then `WaiSubsiteWithAuth`? From what I understand this will also pick up whatever middleware the parent site has.

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.4.34
+
+* Add `WaiSubsiteWithAuth`. [#1394](https://github.com/yesodweb/yesod/pull/1394)
+
 ## 1.4.33
 
 * Adds curly brackets to route parser. [#1363](https://github.com/yesodweb/yesod/pull/1363)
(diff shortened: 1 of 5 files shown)


snoyberg:
This looks good. Could you additionally:

* Bump the minor version number
* Make an entry in the ChangeLog
* Add a Haddock comment to `WaiSubsite` explaining the difference from `WaiSubsiteWithAuth`
* Add a Haddock comment to `WaiSubsiteWithAuth` referring to `WaiSubsite`'s docs, and a `@since` annotation for the new version

mchaver:
Looks like it failed to build in Travis for GHC 8.0.1. Yesod-core built fine but yesod-persistent did not.

snoyberg:
Thanks! We can ignore the build failure, it's spurious and unrelated.

PR #1395, opened 2017-05-12
Don't hardcode templates directory; make it a setting.

nh2:
This allows to load yesod applications into ghci together
with other packages, for example for `stack repl`.

Until now, the templates directory was hardcoded, which
required one to be inside the directory containing `templates`
for `stack repl .` to work.
That automatically forbade using `stack repl` in projects
that have more than one Yesod project.

This commit makes it possible to pass in the templates directory,
so that it can be passed as an absolute directory, so that it
would work from anywere.
To do this, the TemplateHaskell `location` function can be used,
for example for the following directory tree:

```
myproject.cabal
templates/
src/Settings.hs
```

the file `src/Settings.hs` could contain:

```haskell
import Language.Haskell.TH (location, stringE, loc_filename)
import System.FilePath     (takeDirectory, (</>))

widgetFileSettings :: WidgetFileSettings
widgetFileSettings = def
  { wfsTemplatesDirectory = $(location >>= \loc -> stringE $ takeDirectory (loc_filename loc) </> ".." </> "templates")
  }
```

This commit breaks `yesod`'s API and thus requires a version bump,
because the `globFile` function was exported and has its type
signature changed.

CC @chrisdone @snoyberg

### DIFF
--- a/yesod/Yesod/Default/Util.hs
+++ b/yesod/Yesod/Default/Util.hs
@@ -12,6 +12,7 @@ module Yesod.Default.Util
     , WidgetFileSettings
     , wfsLanguages
     , wfsHamletSettings
+    , wfsTemplatesDirectory
     ) where
 
 import qualified Data.ByteString.Lazy as L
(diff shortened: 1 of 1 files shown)


snoyberg:
It seems like it would be easy to achieve this without an API breakage by just defining a new function, any reason not to do that?

nh2:
> It seems like it would be easy to achieve this without an API breakage by just defining a new function, any reason not to do that?

Mainly that it would be confusing when you set `wfsTemplatesDirectory` to something that's not `templates`, but `globFile` would still return `templates/` in the front.

jezen:
After half a decade, it doesn't look like this one will be merged as is. I haven't felt this specific pain personally, and I think the state of tooling has changed a fair bit since this PR was opened. If the pain of not having this resurfaces then maybe it's worth addressing, but for now I'll close this PR.

Issue #1402, opened 2017-06-04
defaultErrorHandler causes inappropriate 406 reponses


whittle:
Copying over from a thread at https://groups.google.com/forum/#!topic/yesodweb/ed6MvRG_pVg

I’m seeing some odd behavior with a handler that uses `selectRep` to choose a representation. I’m on LTS-8.5. The handler in question looks like this: 

```
getVideoIconR :: VideoId -> Handler TypedContent
getVideoIconR i = do
  a <- runDB $ get404 i
  t <- searchForImageUrl a
  selectRep $ do
    provideRepType "image/*" $ (redirect t :: Handler ())
    provideJson $ object ["url" .= t]
```

I have specs covering four cases (excluding cases that test the behavior of `searchForImageUrl`): 

1. Request accepting "application/json" with a VideoId that exists in the database. _Responds with a 200 and a body containing the expected JSON._ 
2. Request accepting "image/*" with a VideoId that exists in the database. _Responds with a 303 and a Location header with the expected URL._ 
3. Request accepting "application/json" with a VideoId that does not exist in the database. _Responds with a 404._ 
4. Request accepting "image/*" with a VideoId that does not exist in the database. _Responds with a 406._ 

The last test case is not what I would expect; I would have thought that short-circuiting execution of the handler with `notFound` would prevent the server from attempting to find a representation for data that won’t be returned anyway. Failing that, I would expect both the representations to behave similarly. 

It looks like defaultErrorHandler uses selectRep to choose a representation for >=400 responses, which can generate an inappropriate 406 response when the error handler doesn’t explicitly provide an acceptable body. This theory is bolstered by replacing `defaultErrorHandler` in my app with a custom error handler that explicitly handles acceptance of "image/*" causes the app to respond with a 404 in the fourth case I listed previously.

snoyberg:
Can you follow up on this with a PR? Ideal would be if you included a fix for this, but even just PRing the failing test case would be great.

StevenXL:
Closing as this was addressed by #1540.

PR #1403, opened 2017-06-05
defaultErrorHandler causes inappropriate 406 reponses

whittle:
Issue originally raised at https://groups.google.com/forum/#!topic/yesodweb/ed6MvRG_pVg

### DIFF
--- a/yesod-core/test/YesodCoreTest/ErrorHandling.hs
+++ b/yesod-core/test/YesodCoreTest/ErrorHandling.hs
@@ -38,6 +38,11 @@ mkYesod "App" [parseRoutes|
 /file-bad-name FileBadNameR GET
 
 /good-builder GoodBuilderR GET
+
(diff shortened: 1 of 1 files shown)


whittle:
Currently only adds failing specs. PR created per @snoyberg’s request on #1402. I am currently examining options for fixing the issue.

snoyberg:
Thanks for following up on this. Based on your comments, I'm going to hold off until I see another ping from you on this. If there's something you'd like me to chime in with before, let me know.

StevenXL:
@whittle I've been trying to make the tests on this pass. While there are several different ways to get the correct status (each with different tradeoffs), the tests expect no `content-type` header, which is a bit trickier. 

I was wondering if you'd like to pair on this PR?

StevenXL:
See #1540

PR #1408, opened 2017-06-14
add clickOn function (closes #1405)

### LINKED ISSUE #1405: Yesod devel no longer recognising app is ready to serve after upgrading to LTS-8.13 from 6.9
This is going to be a bit vague but I apologise in advance - I'm not sure what code it would be required to share.

Immediately after starting yesod devel, you can go to

    http://mylocalhost/dev/root

.. to see the standard yesod devel building page. Previously, once our app had started, the page would redirect to ` http://mylocalhost/dev/root/login`. This is no longer happening.

Since upgrading to lts-8.13, the yesod devel splash reloads indefinitely. If you manually navigate to

    http://mylocalhost:3000/dev/root/login

.. then the app can be accessed. The following URLs are also not working whereas previously they were:

    http://mylocalhost:3000/dev/root
    http://mylocalhost/dev/root/login

How does yesod devel determine that an app is ready to serve? How do I determine why it is no longer doing so?

My guess is that `Application.hs` doesn't have some required new magic for yesod devel to work - what this magic is I haven't been able to determine.

mwotton:


yesod-test/Yesod/Test.hs
```diff
@@ -830,6 +831,16 @@ setUrl url' = do
         , rbdGets = rbdGets rbd ++ H.parseQuery (TE.encodeUtf8 urlQuery)
         }
 
+clickOn :: Yesod site => Query -> YesodExample site ()
```
MaxGabriel: Could you add documentation for this, and an `@since` Haddock annotation? Ideally the documentation would include a small example like most of the rest of yesod-test has (You could just copy an example from your tests and that would be fine)

yesod-test/test/main.hs
```diff
@@ -169,6 +170,15 @@ main = hspec $ do
                     addToken_ "body"
                 statusIs 200
                 bodyEquals "12345"
+            yit "can follow a link via clickOn" $ do
+              get ("/htmlWithLink" :: Text)
+              clickOn "a#thelink"
+              statusIs 200
```
MaxGabriel: Should this test that the actual new page is returned after the link is clicked?

yesod-test/ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## 1.5.7
+
+* Add clickOn
```
MaxGabriel: Could you link to this PR in the changelog so people can get further details? Just line for 1.5.4

yesod-test/Yesod/Test.hs
MaxGabriel: Whichever Github user `@since` is, I bet they get a lot of spurious notifications 🤔

mwotton: good catch, forgot about that one :)

yesod-test/test/main.hs
mwotton: Probably. Would be nice to just check URL of the last response, but i can't seem to do that - i'll check the content instead.

mwotton:
this seems to be failing because of a CI timeout.  @MaxGabriel - could you poke travis?

yesod-test/Yesod/Test.hs
```diff
@@ -830,6 +831,23 @@ setUrl url' = do
         , rbdGets = rbdGets rbd ++ H.parseQuery (TE.encodeUtf8 urlQuery)
         }
 
+
+-- | Click on a link defined by a CSS query
+--
+-- ==== __ Examples__
+--
+-- > get "/foobar"
+-- > clickOn "a#idofthelink"
```
MaxGabriel: Could you add an `@since` annotation here?

MaxGabriel:
Yes, poked. I made one more comment about a missing `@since` annotation

yesod-test/Yesod/Test.hs
```diff
@@ -830,6 +831,25 @@ setUrl url' = do
...
+--
+-- @since 1.5.7 
+clickOn :: Yesod site => Query -> YesodExample site ()
+clickOn query = do
+  withResponse' yedResponse ["Tried to invoke clickOn in order to read HTML of a previous response."] $ \ res ->
+    case findAttributeBySelector (simpleBody res) query "href" of
+      Left err -> failure $ query <> " did not parse: " <> T.pack (show err)
+      Right [[match]] -> get match
+      Right matches -> failure $ "Expected exactly one match for clickOn: got " <> T.pack (show matches)
```
MaxGabriel: I really like this implementation, very clean pattern matching.

yesod-test/Yesod/Test.hs
mwotton: done.

yesod-test/Yesod/Test.hs
mwotton: thanks :)

MaxGabriel:
Great PR!

Issue #1416, opened 2017-07-09
Add replaceOrCreateHeader in Yesod.Core.Handler


psibi:
It will be useful in certain scenarios: https://github.com/yesodweb/yesod/issues/1415

Other frameworks also sort of provide this functionality:

https://docs.djangoproject.com/en/1.11/ref/request-response/#setting-header-fields

PR #1417, opened 2017-07-13
Add replaceOrAddHeader function to Yesod.Core.Handler module

psibi:
Fixes https://github.com/yesodweb/yesod/issues/1416


yesod-core/Yesod/Core/Handler.hs
```diff
@@ -787,6 +789,30 @@ setHeader :: MonadHandler m => Text -> Text -> m ()
...
+  in modify $ \g -> g {ghsHeaders = replaceHeader header (ghsHeaders g)}
+  where
+    sameHeaderName :: Header -> Header -> Bool
+    sameHeaderName (Header n1 _) (Header n2 _) = n1 == n2
+    sameHeaderName _ _ = False
+
+    replaceHeader :: Header -> Endo [Header] -> Endo [Header]
+    replaceHeader header endo =
+      let allHeaders :: [Header] = appEndo endo []
+      in Endo
```
snoyberg: The structure here looks wrong, you're adding the new header to the front of the list, then any additional headers added after, and finally the previously defined header. Instead, I think you want:

```haskell
Endo $ \rest -> header : filter (\x -> not (sameHeaderName x header)) ++ rest
```

We _could_ optimize it by fusing the `filter` and `++` calls, but I don't think it's necessary.

psibi: @snoyberg 

> The structure here looks wrong, you're adding the new header to the front of the list, then any additional headers added after, and finally the previously defined header. 

I was under the impression that the ordering didn't matter. I have updated the code to fix it and have added additional test to ensure that any header isn't lost because of the above function.

paul-rouse:
Shouldn't the comparison with existing headers be case-insensitive?

psibi:
@paul-rouse You're right. But I think it's better fixed at the Eq instances of `Header` type rather than here ?

yesod-core/Yesod/Core/Handler.hs
paul-rouse: Hmm.. It looks as if the ordering is still wrong in your latest repo.  I am trying this

```
addHeader "Foo" "Bar"
replaceOrAddHeader "Foo" "Baz"
addHeader "Foo" "Dup"
replaceOrAddHeader "FOO" "Final"
```

and I get

```
FOO: Final
Foo: Baz
Vary: Accept, Accept-Language
Foo: Dup
```

I did have a quick look at this earlier in the week, and I ended up with a modified version of `tell` which also takes a filtering function `alter` wrapped up in `Endo`:
```
alterAndTell alter hs = modify f
  where f g = let altered = Endo (appEndo (alter `mappend` ghsHeaders g) [] ++)
              in g { ghsHeaders = altered `mappend` hs }
```

paul-rouse:
Wouldn't Eq on Header compare both the header name and the value?

psibi:
@paul-rouse Maybe I'm misunderstanding something, but can't we control the behavior anyway we like when we write our own manual instances of the type ?

paul-rouse:
Yes, but there's already a derived Eq instance for Header in Types, which I assumed was there for a reason.  Perhaps it isn't used - I haven't looked that far!

psibi:
@paul-rouse Ah okay. My proposal is just to remove the derived Eq instance and write our custom instance. :)

paul-rouse:
@psibi Did you also see my comment on the header ordering - I realise I didn't mention you, and I don't think github sends notifications for those.

psibi:
@paul-rouse Just seeing it and that indeed is an issue. I will comment there.

yesod-core/Yesod/Core/Handler.hs
psibi: @paul-rouse So, this should be the correct ordering - right ?
```
FOO: Final
Foo: Dup
Foo: Baz
```

paul-rouse: @psibi I don't think so.  The behaviour of `addHeader` is that a later call puts the header later in the result, so to be consistent, and also "natural", I think, the result should have been (since Yesod appears to add `Vary` before the handler code):
```
Vary: Accept, Accept-Language
Foo: Baz
Foo: Dup
FOO: Final
```
Obviously that's with the case sensitive match still - actually we want just the Vary and the last FOO, of course!

psibi: @paul-rouse Thanks for the feedback. I have fixed (hopefully!) the ordering logic with the latest push. With respect to the equality of case insensitive comparison, I will open a new issue.

paul-rouse:
@psibi I fear I may have contributed to a miscommunication (in the now obsolete comments).  This ordering is fine, but I don't think you necessarily have to keep the replacement in the original position.  The example I showed had *additions* out of order.

psibi:
@paul-rouse Ah, I was under the impression that you were concerned about the ordering there. :) I will fix the extra additions which you have illustrated in this issue soon: https://github.com/yesodweb/yesod/issues/1418

paul-rouse:
@psibi I *was* concerned about the ordering, but only to the extent that when `replaceOrAddHeader` *added* a header it was at the front, unlike `addHeader`.  Thinking again, I was probably being too strict, since the header is guaranteed unique after `replaceOrAddHeader`, and any subsequent `addHeader` with the same name will still append (which *does* matter).

psibi:
@paul-rouse With the latest push, it does case insensitive equality check on header name. This is the response we get for your above code right now:

```
Vary: Accept, Accept-Language
FOO: Final
```

paul-rouse:
Sounds great, but I'm not seeing a push at the moment :)

psibi:
@paul-rouse Ah, sorry about that. Pushed it now.

paul-rouse:
@psibi Yes that looks great!  I have tried a few more combinations of addHandler and replaceOrAddHandler and all is fine.  Sorry I was unclear and led us into an over-pedantic view of the ordering, but the order is certainly valid.  Thank you!

psibi:
@paul-rouse Thanks for the detailed feedback in the PR!

@snoyberg I'm merging this as the Travis test passes and the Windows build seems to be failing for reasons not related to this PR. Can you make a new `yesod-core` release ?

PR #1420, opened 2017-07-18
add-handler: add `src/` to file path if haskell source files in `src`

### LINKED ISSUE #1413: add-handler is broken since the scaffolding update
Since the [June 2017 scaffolding update](http://www.yesodweb.com/blog/2017/06/updated-yesod-scaffolding), yesod add-handler is broken. Running it produces this error:

    yesod: Application.hs: openFile: does not exist (No such file or directory)

It looks like yesod-bin/AddHandler.hs hasn't been updated to reference src/Application.hs. Should it be updated, or is add-handler dead, per the discussion on the [scaffolding update's pull request](https://github.com/yesodweb/yesod-scaffold/pull/147)?

Ephemera:
It's moved into `src` directory.

### DIFF
--- a/yesod-bin/AddHandler.hs
+++ b/yesod-bin/AddHandler.hs
@@ -5,9 +5,13 @@ import Prelude hiding (readFile)
 import System.IO  (hFlush, stdout)
 import Data.Char  (isLower, toLower, isSpace)
 import Data.List  (isPrefixOf, isSuffixOf, stripPrefix)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, listToMaybe)
 import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
+import Distribution.PackageDescription.Parse (readPackageDescription)
+import Distribution.PackageDescription.Configuration (flattenPackageDescription)
+import Distribution.PackageDescription (allBuildInfo, hsSourceDirs)
+import Distribution.Verbosity (normal)
 import System.Directory (getDirectoryContents, doesFileExist)
 import Control.Monad (unless)
 
@@ -31,7 +35,7 @@ cmdLineArgsError = "You have to specify a route name if you want to add handler 
 addHandler :: Maybe String -> Maybe String -> [String] -> IO ()
 addHandler (Just route) pat met = do
     cabal <- getCabal
-    checked <- checkRoute route
+    checked <- checkRoute route cabal
     let routePair = case checked of
           Left err@EmptyRoute -> (error . show) err
           Left err@RouteCaseError -> (error . show) err
@@ -54,7 +58,7 @@ addHandlerInteractive = do
         putStr "Name of route (without trailing R): "
         hFlush stdout
         name <- getLine
-        checked <- checkRoute name
+        checked <- checkRoute name cabal
         case checked of
             Left err@EmptyRoute -> (error . show) err
             Left err@RouteCaseError -> print err >> routeInput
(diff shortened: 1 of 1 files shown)


snoyberg:
Thank you!

There are still lots of old scaffolded sites out there, it would be better to check for the presence of both sets of files and then choose the appropriate one. Simplest version: if `src/` exists, use that, otherwise don't. Would you be able to implement that logic?

Ephemera:
@snoyberg 

I added some logic. It depends on `hs-source-dirs` in .cabal file.  
Is it reasonable?

yesod-bin/AddHandler.hs
```diff
@@ -214,3 +221,9 @@ mkHandler name pattern methods = unlines
 uncapitalize :: String -> String
 uncapitalize (x:xs) = toLower x : xs
 uncapitalize "" = ""
+
+getSrcDir :: FilePath -> IO FilePath
+getSrcDir cabal = do
+    pd <- readPackageDescription normal cabal
+    let dirs = concatMap hsSourceDirs . allBuildInfo $ flattenPackageDescription pd
+    return (dirs !! 0)
```
psibi: I'm not that comfortable seeing a partial function. Can you fix it ?

erikd:
Sorry, to be a pain in the neck, but can we have a better commit message than "Fix #XXX" ? Knowing the ticket number is useful, but a short text explanation is also important.

yesod-bin/AddHandler.hs
Ephemera: @psibi 

Fixed it.

yesod-bin/AddHandler.hs
```diff
@@ -214,3 +221,10 @@ mkHandler name pattern methods = unlines
 uncapitalize :: String -> String
 uncapitalize (x:xs) = toLower x : xs
 uncapitalize "" = ""
+
+getSrcDir :: FilePath -> IO FilePath
+getSrcDir cabal = do
+    pd <- flattenPackageDescription <$> readPackageDescription normal cabal
+    let buildInfo = allBuildInfo pd
+        srcDirs = concatMap hsSourceDirs buildInfo
+    return $ head srcDirs
```
psibi: `head` is also a partial function. You probably can throw impure exception here.  @snoyberg Would you be fine with a partial function or is there some other better way ?

snoyberg: I'm not 100% certain, but I _think_ this is worse than it seems. It's perfectly valid to omit any `hs-source-dirs`, which defaults to "use current directory." Unless I'm totally off, what we want here is `fromMaybe "." $ listToMaybe srcDirs`.

Ephemera: but it looks already having default value  with current directory for `hs-source-dirs`.
https://github.com/haskell/cabal/blob/master/Cabal/Distribution/PackageDescription/Configuration.hs#L580-L584

snoyberg: Interesting, I didn't know that. Nonetheless, it's not promised via the Cabal API, and it could change in the future, causing silent breakage on our end. I'd rather be explicit about that case.

Ephemera: That's right. I fixed it.

snoyberg:
Sorry, I put a bad commit on master. Can you pull or rebase?

Ephemera:
@snoyberg 

Sure, I rebased.

snoyberg:
Thanks! LGTM. @psibi did you have any comments on this?

psibi:
Nope. Looks good.

snoyberg:
Thanks!

PR #1421, opened 2017-07-19
Add implicit param HasCallStack to assertions

sestrella:
`yesod-test` current error messages looks like follows:

```
./Yesod/Test.hs:351: 
  1) testing status code
       Expected status was 401 but received status was 400
```
The error message is not easy to localize since it points to the module and line number where the assertion is defined instead of the place where it is called.

Adding [HasCallStack](http://hackage.haskell.org/package/call-stack-0.1.0/docs/Data-CallStack.html) to each assertion fix the error message since it passes the CallStack to the [HUnit](http://hackage.haskell.org/package/HUnit-1.6.0.0/docs/Test-HUnit-Base.html) matchers used by `yesod-test` under the hood. Here is an example of how an error message looks before and after the changes introduced by this PR:

Before

```
./Yesod/Test.hs:351: 
  1) testing status code
       Expected status was 401 but received status was 400
```

After

```
test/Handler/FooSpec.hs:24: 
  1) testing status code
       Expected status was 401 but received status was 400
```

### DIFF
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.5.8
+* Added implicit parameter HasCallStack to assertions.
+[#1421](https://github.com/yesodweb/yesod/pull/1421)
+
 ## 1.5.7
 
 * Add clickOn.
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -4,6 +4,8 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE ConstraintKinds #-}
 
 {-|
 Yesod.Test is a pragmatic framework for testing web applications built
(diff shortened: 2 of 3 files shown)


snoyberg:
Would you be able to switch this over to using conditional compilation for older GHCs? I'd rather that than add another dependency.

sestrella:
@snoyberg thank you for the comment. I guess I could copy the content of [Data.CallStack] defined at [call-stack](http://hackage.haskell.org/package/call-stack) into `Yesod.Test` module. `Data.CallStack` module looks like follows:

```hs

#if MIN_VERSION_base(4,8,1)
import qualified GHC.Stack as GHC
#endif

#if MIN_VERSION_base(4,9,0)
import           GHC.Stack (HasCallStack)
#elif MIN_VERSION_base(4,8,1)
type HasCallStack = (?callStack :: GHC.CallStack)
#else
import GHC.Exts (Constraint)
type HasCallStack = (() :: Constraint)
#endif

type CallStack = [(String, SrcLoc)]

callStack :: HasCallStack => CallStack
#if MIN_VERSION_base(4,9,0)
callStack = drop 1 $ GHC.getCallStack GHC.callStack
#elif MIN_VERSION_base(4,8,1)
callStack = drop 2 $ GHC.getCallStack ?callStack
#else
callStack = []
#endif
```

However, I'm actually exposing a hidden package rather than adding a new dependency since `call-stack` is a dependency of [HUnit](https://github.com/hspec/HUnit/blob/master/HUnit.cabal#L35) already. What do you think?

sestrella:
@snoyberg taking a second look, I found this change is not compatible with [older](https://travis-ci.org/yesodweb/yesod/jobs/255120712) stack resolvers such as `lts-3` since `call-stack` is not part of it. I'm going to add the type alias `HasCallStack` to `Yesod.Test` using conditional compilation as you originally suggested.

sestrella:
@snoyberg all changes are in place.

snoyberg:
Looks good. Could you also do a minor version bump and update the changelog?

sestrella:
@snoyberg sure thing! I just added a new entry to the CHANGELOG and bumped the minor version. Thank you for taking a look.

snoyberg:
Thanks!

PR #1399, opened 2017-05-18
Move yesod-auth to Cryptonite (from cryptohash)

psibi:
Partially addresses #1397

PR #1424, opened 2017-07-23
Convert yesod-static to cryptonite.

QuLogic:
I don't really know much Haskell, but I'd like to build with one less crypto lib. I tried to follow the changes in #1399 and it seems to build. Hopefully no other changes are necessary.


snoyberg:
Thanks!

PR #1432, opened 2017-08-21
Reduce verbosity using Monadic Forms

sestrella:
Building forms where some fields validations depend on other fields is simple using Monadic Forms, but as the own Yesod website mentioned:

> it does so at the cost of being more verbose.

Some of this verbosity is caused by the need of grouping the `FiedView site` returned by either `mreq` or `mopt`. Also on the other hand, `formToAForm` requires a Monadic Form like `MForm m (FormResult a, [FieldView site])` to able to transform to `AForm m a`. Consider the following scenario:

```hs
formToAForm $ do
  (field1F, field1V) <- mreq textField MsgField1 Nothing
  (field2F, field2V) <- mreq (checkWith field1F textField) MsgField2 Nothing
  (field3F, field3V) <- mreq (checkWith field1F textField) MsgField3 Nothing
  return
    ( MyForm <$> field1F
                     <*> field2F
                     <*> field3F
    , [field1V, field2V, field3V]
    )
```

While the previous example could be refactored to use Applicative Form style, it requires combining the fields in order to be able to run a custom check. Doing that for more than one field that depends on one value is a little bit tricky using Applicative style.

I'll like to propose adding the following variant of a Monadic Form:

```hs
type WForm m a = MForm (WriterT [FieldView (HandlerSite m)] m) a
```

And the correspondent `req` and `opt` functions like:

```hs
wreq :: (...) => Field m a -> FieldSettings site -> Maybe a -> WForm m (FormResult a)

wopt :: (...) => Field m a -> FieldSettings site -> Maybe (Maybe a) -> WForm m (FormResult (Maybe a))

wFormToAForm ::  WForm  m (FormResult a) -> AForm m a
```

The example above then could be rewritten as follows:

```hs
wFormToAForm $ do
  field1F <- wreq textField MsgField1 Nothing
  field2F <- wreq (checkWith field1F textField) MsgField2 Nothing
  field3F <- wreq (checkWith field1F textField) MsgField3 Nothing
  return $
      MyForm <$> field1F
                    <*> field2F
                    <*> field3F
```

I'll be more than happy to submit a PR if this proposal makes sense to the maintainers.

Issue: #1432

### DIFF
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -13,7 +13,12 @@ module Yesod.Form.Functions
       -- * Applicative/Monadic conversion
     , formToAForm
     , aFormToForm
+    , mFormToWForm
+    , wFormToAForm
+    , wFormToMForm
       -- * Fields to Forms
+    , wreq
+    , wopt
     , mreq
     , mopt
     , areq
@@ -51,8 +56,9 @@ module Yesod.Form.Functions
 import Yesod.Form.Types
 import Data.Text (Text, pack)
 import Control.Arrow (second)
-import Control.Monad.Trans.RWS (ask, get, put, runRWST, tell, evalRWST, local)
 import Control.Monad.Trans.Class
+import Control.Monad.Trans.RWS (ask, get, put, runRWST, tell, evalRWST, local, mapRWST)
+import Control.Monad.Trans.Writer (runWriterT, writer)
 import Control.Monad (liftM, join)
 import Data.Byteable (constEqBytes)
 import Text.Blaze (Markup, toMarkup)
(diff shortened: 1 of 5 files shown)


sestrella:
@snoyberg would you mind taking a look, please?

yesod-form/Yesod/Form/Functions.hs
```diff
@@ -105,6 +111,58 @@ askFiles = do
     (x, _, _) <- ask
     return $ liftM snd x
 
+-- | Converts a form field into monadic form 'WForm'. This field requires a
+-- value and will return 'FormFailure' if left empty.
+--
+-- Since 1.4.14
```
snoyberg: Minor nitpick, but would you be able to convert this to the new style `@since` comments?

yesod-form/yesod-form.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-form
-version:         1.4.13
+version:         1.4.14
```
snoyberg: Bonus points for handling the version bump _and_ changelog, much appreciated!

yesod-test/test/main.hs
```diff
@@ -150,6 +150,18 @@ main = hspec $ do
                     addToken
                 statusIs 200
                 bodyEquals "12345"
+            yit "labels WForm" $ do
```
snoyberg: Triple bonus points for test cases!

sestrella:
@snoyberg  thank you very much for reviewing this PR. All comments have been addressed.

snoyberg:
Awesome, thank you!

PR #1433, opened 2017-08-22
Give FormResult an Alternative instance

jessekempf:
I've got a scenario where I have multiple forms on one page. Since only one will ever be submitted at a time it's convenient having an Alternative instance that surfaces errors first, and if there isn't an error, surfaces the first successful form result.

### DIFF
--- a/yesod-form/Yesod/Form/Types.hs
+++ b/yesod-form/Yesod/Form/Types.hs
@@ -30,7 +30,7 @@ import Text.Blaze (Markup, ToMarkup (toMarkup), ToValue (toValue))
 #define Html Markup
 #define ToHtml ToMarkup
 #define toHtml toMarkup
-import Control.Applicative ((<$>), Applicative (..))
+import Control.Applicative ((<$>), Alternative (..), Applicative (..))
 import Control.Monad (liftM)
 import Control.Monad.Trans.Class
 import Data.String (IsString (..))
@@ -45,6 +45,8 @@ import Data.Foldable
 --
 -- The 'Applicative' instance will concatenate the failure messages in two
 -- 'FormResult's.
+-- The 'Alternative' instance will choose 'FormFailure' before 'FormSuccess',
+-- and 'FormMissing' last of all.
 data FormResult a = FormMissing
                   | FormFailure [Text]
                   | FormSuccess a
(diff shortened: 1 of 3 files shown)


yesod-form/Yesod/Form/Types.hs
```diff
@@ -80,6 +82,17 @@ instance Data.Traversable.Traversable FormResult where
       FormFailure errs -> pure (FormFailure errs)
       FormMissing -> pure FormMissing
 
+instance Alternative FormResult where
```
snoyberg: Can you add a `@since` comment here, bump the minor version in the cabal file, and update the changelog?

yesod-form/Yesod/Form/Types.hs
```diff
@@ -80,6 +82,17 @@ instance Data.Traversable.Traversable FormResult where
...
+    empty = FormMissing
+
+    (FormFailure e1) <|> (FormFailure _)  = FormFailure e1
```
snoyberg: I think this can be more concisely expressed with just three clauses:

```haskell
FormFailure e <|> _ = FormFailure e
FormSuccess s <|> _ = FormSuccess s
FormMissing <|> f = f
```

jessekempf:
@snoyberg: Done. My version was some face-palmingly verbose code.

yesod-form/Yesod/Form/Types.hs
```diff
@@ -45,6 +45,8 @@ import Data.Foldable
 --
 -- The 'Applicative' instance will concatenate the failure messages in two
 -- 'FormResult's.
+-- The 'Alternative' instance will choose 'FormFailure' before 'FormSuccess',
```
snoyberg: Rereading this comment, I'm not sure it's true (though maybe my "cleanup" broke the intended behavior :)). There doesn't seem to be any preference for `FormFailure` over `FormSuccess` in the instance, only `FormMissing` has special handling.

jessekempf: @snoyberg: I think that might be from the cleanup :).

I've pushed a concise version that (I believe) has the correct behavior.

yesod-form/Yesod/Form/Types.hs
```diff
@@ -80,6 +82,16 @@ instance Data.Traversable.Traversable FormResult where
...
 
+-- | @since 1.4.15
+instance Alternative FormResult where
+    empty = FormMissing
+
+    FormFailure e    <|> _             = FormFailure e
+    _                <|> FormFailure e = FormFailure e
+    FormSuccess s    <|> FormSuccess _ = FormSuccess s
+    FormMissing      <|> result        = result
+    result           <|> FormMissing   = result
```
snoyberg: Is this clause necessary? It looks like it's overlapping.

jessekempf: Yes, without it the pattern match is non-exhaustive.

snoyberg: I kept overlooking the `FormSuccess` on the RHS of line 91. OK, let's leave it as-is.

snoyberg:
Thanks!

PR #1435, opened 2017-08-25
Add Korean translation

Ephemera:


yesod-form/Yesod/Form/I18n/Korean.hs
```diff
@@ -0,0 +1,26 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Yesod.Form.I18n.Korean where
```
psibi: You have to also expose this module from the cabal file.

Ephemera: @psibi 

Fixed it!

psibi:
Thanks!

Issue #1436, opened 2017-08-29
Proposal: expose Yesod.PasswordStore


paul-rouse:
The motivation for this is that `yesod-auth-hashdb` still uses `pwstore-fast`, but now that Yesod.PasswordStore has diverged significantly from it (#1399), with different dependencies, I would like to move over to using the same version as the core yesod-auth modules.

I am quite happy to do a PR, of course, but I'd like to ask for some opinions first:
  1. Are there any objections in principle?
  2. Exposing it at the top level of the Yesod hierarchy seems slightly wrong to me - how about moving it to Yesod.Auth.Util.PasswordStore?

snoyberg:
Sounds perfectly reasonable to me. The only other thought I had is: perhaps we should expose it from something like `Yesod.Auth.Internal.PasswordStore`, in case the API isn't really stable yet. But even that last overhaul looks like it was implementation changes only, not an API breakage. In other words: I'm fine with everything you've mentioned, PR welcome, your call on whether `.Internal` is warranted.

paul-rouse:
I had wondered about `.Internal`, but I would prefer to be more positive about encouraging use in third-party auth plugins, and, as you say, even the recent change didn't alter the API.

I'll close this now in favour of the PR https://github.com/yesodweb/yesod/pull/1437, - I'll wait for a review on that.

And thank's for the invitation to the team!

PR #1437, opened 2017-08-29
Expose Yesod.Auth.Util.PasswordStore

paul-rouse:
Implements https://github.com/yesodweb/yesod/issues/1436


yesod-auth/Yesod/Auth/Util/PasswordStore.hs
```diff
@@ -71,7 +66,7 @@
 -- iteration count. This does not have a significant effect on security, but can
 -- be handy for compatibility with other code.
 
-module Yesod.PasswordStore (
+module Yesod.Auth.Util.PasswordStore (
```
snoyberg: Can you add `@since 1.4.18` comments to the module and all identifiers exposed here?

paul-rouse: Of course!  Just pushed../

psibi:
Thanks, a new version has been released to Hackage.

PR #1439, opened 2017-08-31
Correct Yesod-Auth's usage of "log in" vs "login" in English.

jessekempf:
"Log in" (two words) is a verb, indicating the action of, well, logging in. "Login" (one word) is a noun, indicating the credentials used to log in.


snoyberg:
Thanks!

PR #1442, opened 2017-09-05
remove the `--fast` flag from `stack build` executed by `yesod devel`

ncaq:
It is a suggestion to remove the `--fast` flag from `stack build` executed by `yesod devel`.

I am writing an application with the function of uploading files to S3 with yesod.
The size corresponds to about 2GB.

The file is solved to `ByteString`.

Here, when optimization is disabled when building the yesod application, the application crashes using a lot of memory.
Problems will occur with files of about 100MB.

In other words, I went optimize when `yesod devel` but if `stack build --fast` is on,` stack exec -- yesod devel -e --ghc-options="-O2"` can not override `--fast` and optimization will not take effect.

In old yesod, invalidation of optimization at devel was written in cabal files of individual applications, so it was easy to change, but now I have no way to change.

So open the pull request to remove the `--fast` flag.

I think that it is better for the user to be able to select whether to optimize, as in the past.

At least, my application does not work on `yesod devel 'now and I am in trouble.

Please forgive me for reporting the problem now though it is a change 10 months ago.

thanks.

### DIFF
--- a/yesod-bin/ChangeLog.md
+++ b/yesod-bin/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.5.2.6
+
+* Deleted --fast flag of stack build
+
 ## 1.5.2.5
 
 * Support for `add-handler` when modules are in `src/` directory [#1413](https://github.com/yesodweb/yesod/issues/1413)
--- a/yesod-bin/Devel.hs
+++ b/yesod-bin/Devel.hs
@@ -334,7 +334,6 @@ devel opts passThroughArgs = do
                        $ setStderr createSource
                        $ setDelegateCtlc True $ proc "stack" $
                 [ "build"
-                , "--fast"
                 , "--file-watch"
 
                 -- Indicate the component we want
--- a/yesod-bin/yesod-bin.cabal
+++ b/yesod-bin/yesod-bin.cabal
@@ -1,5 +1,5 @@
 name:            yesod-bin
-version:         1.5.2.5
+version:         1.5.2.6
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>


ncaq:
I attach the code we use to figure out the file to `ByteString`.

The following code works if the optimization flag is the default, but will not work if the optimization is disabled (-O0).

~~~hs
-- | Bytestring 'fileSource'
fileSourceBytes :: MonadResource m => FileInfo -> m ByteString
fileSourceBytes fileInfo = fileSource fileInfo $$ foldC
~~~

If you have a policy that you should not write code that will not work if you disable optimization, please tell it.

psibi:
Thanks for the PR. But if the behavior of the code is changing based on the optimization used - then I think it's a bug in the code. In the case, if what you say is true - I suggest you file a bug against conduit.

paul-rouse:
Actually @psibi I am not clear whether there is a bug here - the files being processed seem to be large, so it *may* be reasonable that optimisation is needed to avoid exhausting memory.

However,  @ncaq, I don't understand why `yesod-devel` needs to be changed.  Can't you test on small examples during development, using `yesod-devel`, and then build the production version separately, using `stack build`, taking whatever optimisation settings you have put in the cabal file?

ncaq:
I appreciate the response.

@psibi I have not identified the source of the problem.
However, perhaps, if there is a problem, I think that it is not a bug of conduit but a problem of ByteString.
Before I wrote a program to post lots of large size files with yesod-test.
Only when optimization was disabled (`-O0`), OOM-killer will run without posting memory after posting.
At that time, since `yesod devel` had been built without using stack, we restored the cabal file optimization option to the default and addressed it.

@paul-rouse 
I thought that `ghc-options: -O` in the cabal file would be overwritten by`stack build --fast`.
I thought about checking it, I thought that optimization options on cabal files would be invalidated when` stack build --fast` was done.
I attempted to reproduce
the memory leak to confirm it, but now I do not reproduce it…

Building a pull request based on a problem that is not sure of reproducibility.
I apologize, I close this pull request.

psibi:
Thanks, do let us know if you are able to reproduce the bug.

PR #1444, opened 2017-09-06
Add support to yesod-core for weak etags

iand675:


yesod-core/test/YesodCoreTest/Redirect.hs
```diff
@@ -82,7 +84,7 @@ specs = describe "Redirect" $ do
             { pathInfo = ["etag"]
             , requestHeaders = [("if-none-match", "hello world")]
             }
-        assertStatus 304 res
+        assertStatus 200 res
```
iand675: Since I anticipate this change being asked about, ETags *must* be quoted, so performing a redirect on unquoted ETags doesn't match the spec.

snoyberg: I'm (very) weakly opposed to making such a change, since it's possible that clients out there are, in fact, sending requests contrary to the spec. I don't have an example of such a client right now though. Let me ask: is this change to be more strict required for the rest of the functionality?

iand675: It's not strictly required. I could replace

> `StrongEtag (encodeUtf8 etag) `elem` matches

with a fold that checks if either a matching `StrongEtag` or `InvalidEtag` are in the list, although it pains me (weakly :P) to intentionally reintroduce incorrect behavior.

There are other infelicities as well around the current Etag parsing I believe, such as not supporting commas within them when that ought to be supported– which I'd probably like to implement properly in the future, so I guess the question is under what circumstances doing the wrong thing is preferable to not breaking clients that are doing the wrong thing. Are there other instances where Yesod intentionally breaks from HTTP RFCs?

snoyberg: Yeah, I'm sure we do. IIRC, we allow for HTTP requests to use an LF line ending instead of a CRLF line ending, for instance.

iand675:
@snoyberg I've gone ahead and tweaked the code to support invalid ETags again, but went ahead and added a note to the unit test around the old behavior as a breadcrumb in case it causes confusion to someone in the future.

PR #1446, opened 2017-09-21
Extend `YesodAuthEmail` to support extensible password hashing

bigs:
## Commit Log

This change introduces `hashAndSaltPassword` and `verifyPassword` to the
`YesodAuthEmail` type class, allowing users to implement their own hashing
schemes (i.e. to provide compatibility with an existing database). It also
updates the default handlers to use these new functions when appropriate. The
functions have default implementation such that behavior for legacy applications
should not change.

### Prior Art

The only other discussion about offering extensible password hashing was found in [this email list thread](https://groups.google.com/forum/#!topic/yesodweb/dqZDA18Qjy8). It didn't seem to come to a conclusion, though there have been others asking about this functionality.

### Regressions

There are currently no tests in the `yesod-auth` project so there was nothing to measure a regression against. I am, however, successful in using in a project I'm currently working on that requires integration with a legacy database.

### Steps to Reproduce (stack)

Clone this revision locally and add the `yesod/yesod-auth` folder to your `packages` list in your `stack.yaml`. Build your project using `YesodAuthEmail` and the behavior should be no different from v1.4.19.

Should you choose to implement an alternative password hashing, provide definitions for `hashAndSaltPassword` and `verifyPassword`.

### Miscellanea

Happy to rename or reformat any functions, I'm not married to anything here.

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -4,7 +4,7 @@
 {-# LANGUAGE PatternGuards           #-}
 {-# LANGUAGE QuasiQuotes             #-}
 {-# LANGUAGE Rank2Types              #-}
-{-# LANGUAGE ScopedTypeVariables#-}
+{-# LANGUAGE ScopedTypeVariables     #-}
 {-# LANGUAGE TypeFamilies            #-}
 -- | A Yesod plugin for Authentication via e-mail
 --
@@ -132,7 +132,7 @@ import           Safe                     (readMay)
 import           System.IO.Unsafe         (unsafePerformIO)
 import qualified Text.Email.Validate
 import           Data.Aeson.Types (Parser, Result(..), parseMaybe, withObject, (.:?))
-import           Data.Maybe (isJust, isNothing, fromJust)
+import           Data.Maybe (isJust)
 import Data.ByteArray (convert)
 
 loginR, registerR, forgotPasswordR, setpassR :: AuthRoute
@@ -202,6 +202,22 @@ class ( YesodAuth site
     --
     -- @since 1.1.0
     setVerifyKey :: AuthEmailId site -> VerKey -> HandlerT site IO ()
+
+    -- | Hash and salt a password
+    --
+    -- Default: 'saltPass'.
+    --
+    -- @since 1.4.20
+    hashAndSaltPassword :: Text -> HandlerT site IO SaltedPass
+    hashAndSaltPassword = liftIO . saltPass
+
+    -- | Verify a password matches the stored password for the given account.
+    --
+    -- Default: Fetch a password with 'getPassword' and match using 'Yesod.Auth.Util.PasswordStore.verifyPassword'.
+    --
+    -- @since 1.4.20
+    verifyPassword :: Text -> SaltedPass -> HandlerT site IO Bool
+    verifyPassword plain salted = return $ isValidPass plain salted
 
     -- | Verify the email address on the given account.
     --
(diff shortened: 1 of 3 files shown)


yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -203,6 +203,18 @@ class ( YesodAuth site
     -- @since 1.1.0
     setVerifyKey :: AuthEmailId site -> VerKey -> HandlerT site IO ()
 
+    -- | Hash and salt a password
+    --
+    -- Default: 'saltPass'.
+    hashAndSaltPassword :: Text -> HandlerT site IO SaltedPass
```
psibi: Add @since haddock style documentation for all the newly introduced functions.

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -756,11 +770,14 @@ postPasswordR = do
                 case mrealpass of
                     Nothing ->
                         lift $ loginErrorMessage (tm setpassR) "You do not currently have a password set on your account"
-                    Just realpass
-                        | isNothing current -> loginErrorMessageI LoginR Msg.BadSetPass
-                        | isValidPass (fromJust current) realpass -> confirmPassword aid tm jcreds
-                        | otherwise ->
-                            lift $ loginErrorMessage (tm setpassR) "Invalid current password, please try again"
+                    Just realpass -> do
+                        passValid <- lift $ verifyPassword (fromJust current) realpass
```
psibi: Avoid using partial functions.

bigs:
@psibi thanks! the partial function was lifted from and i knew i should have handled it there! not entirely sure what you mean re: haddock, though. could you explain?

bigs:
@psibi i've removed the partial function and added what seems like an appropriate redirection for the case (which afaict shouldn't occur). if the user didn't supply a password, i redirect them to the login page with an appropriate message.

bigs:
@psibi ohhhhhhhhhhhhh github was just auto-correcting `@since` to a username. one thing i'm not sure about is whether or not i should bump the version number. should i go to `1.4.20`? with that answer in, i'll update the haddocks!

yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -753,14 +771,20 @@ postPasswordR = do
...
-                        | isValidPass (fromJust current) realpass -> confirmPassword aid tm jcreds
-                        | otherwise ->
-                            lift $ loginErrorMessage (tm setpassR) "Invalid current password, please try again"
+                    (_, Nothing) ->
+                        lift $ loginErrorMessage (tm loginR) "You did not submit a password, please try again"
+                    (Just realpass, Just current') -> do
+                        passValid <- lift $ verifyPassword current' realpass
+                        case () of
+                            _
+                                | isNothing current -> loginErrorMessageI LoginR Msg.BadSetPass
```
psibi: Can this check be avoided as we are doing this on top .also ?

psibi:
@bigs Also it should be `1.4.20`

yesod-auth/Yesod/Auth/Email.hs
bigs: ahhh great point. in fact, this is the proper error message to deliver. refactoring & updating version

bigs:
@psibi refactors and version bump added as requested!

bigs:
sounds good! backwards compatible, yes. will squash and update log!

bigs:
@psibi changelog updated and branch squashed :) should be set!

psibi:
Thanks for the PR!

bigs:
@psibi happy to contribute! hoping it's the first of many.

PR #1448, opened 2017-09-30
Relaxation of IO to MonadIO for runDB

schnecki:


snoyberg:
This is a breaking change, as it puts more constraints on the user's definition of `runDB`. Furthermore, it's broken as-is, since you can't use `defaultRunDB` as an implementation of `runDB`. Before considering any kind of relaxing here, I'd like to see a use case to justify it.

schnecki:
Hi, my use case was a database query in onErrorHtml: 
```
instance YesodAuth App where
     onErrorHtml :: MonadResourceBase m => Route master -> Text -> HandlerT master m Html
     onErrorHtml dest msg = do
       mVal <- runDB $ getBy $ UniqueSettingsKey settingRedirectOnLoginFailure
       ...
```
Here `m` must be IO as runDB requieres it, `onErrorHtml :: MonadResourceBase m => Route master -> Text -> HandlerT master m Html`. Thus m must be IO for runDB, but cannot be IO because of the onErroHtml signature (which might be changed instead of runDB). I don't think `liftIO` works here because of `runDB` requiering `m` to be `IO`.

```
    • Couldn't match type ‘m’ with ‘IO’
      ‘m’ is a rigid type variable bound by
        the type signature for:
          onErrorHtml :: forall (m :: * -> *).
                         MonadResourceBase m =>
                         Route App -> Text -> HandlerT App m Html
        at src/Foundation.hs:312:5
      Expected type: HandlerT App m (Maybe (Entity Setting))
        Actual type: HandlerT App IO (Maybe (Entity Setting))
```
However, I agree that a change to `runDB` is breaking. Thus, might it be better to change `onErrorHtml` to require `IO` as well, like `onLogin` and `onLogout` does, instead of the `MonadResourceBase m` constraint?

snoyberg:
It sounds to me like you need a function `liftHandler :: MonadIO m => HandlerT site IO a -> HandlerT site m a`. Is that correct?

schnecki:
True, that would work. I'm not aware of such a function though.

snoyberg:
I think liftHandlerT will do what you're looking for.

schnecki:
Yes that's exactly what I was looking for. Thanks a lot.

schnecki:
Anyhow,. I'm actually curious. Why did you decide to constraint `runDB` to
`IO`?

snoyberg:
Since with a simple call to `liftHandlerT` you can get the more polymorphic
version, and it's easier to write the instances this way.

schnecki:
Thanks for the clearification.

PR #1453, opened 2017-11-08
extend docs of defaultMaybeAuthId

agreif:
make more explicit that on each call a database access is done. This can be of relevance and sometimes redundant with other Handler functionality

### DIFF
--- a/yesod-auth/Yesod/Auth.hs
+++ b/yesod-auth/Yesod/Auth.hs
@@ -243,7 +243,8 @@ credsKey = "_ID"
 -- | Retrieves user credentials from the session, if user is authenticated.
 --
 -- This function does /not/ confirm that the credentials are valid, see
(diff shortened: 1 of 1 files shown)


paul-rouse:
The wording seems a little misleading.  The database lookup is done on the *first* call in a request, but it should be cached for the rest of the request - so I don't understand your remark about it being "redundant with other Handler functionality".  If you are seeing more than one DB access resulting from `maybeAuth`, `requireAuth` and relatives during a single request, I think you should raise it as an issue.

agreif:
you are right, it is called only once per request. 
My point was only to state that a DB access is done, because the docs does not mention it and the name of the 'cachedAuth' function also does not indicate that any database access is done.

agreif:
thanks, I applied your suggestions

paul-rouse:
Thanks!

PR #1455, opened 2017-11-17
Give better error messages when CSRF validation fails

MaxGabriel:
This PR modifies the AJAX CSRF checking code to provide a clearer error message, based on my experience integrating a frontend with a Yesod app with @devonzuegel. The error message now says:

> A valid CSRF token wasn't present. Because the request could have been forged, it's been rejected altogether. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection.
> Tip: By default, the CSRF token is sent to the client in a cookie named XSRF-TOKEN.
> Tip: The server is looking for the token in the following locations:
> - An HTTP header named X-XSRF-TOKEN (which is not currently set)
> - A POST parameter named _token (which has the current, incorrect value: 'PpjJoRnZPmfoo') 

compared to the previous, static message:

> A valid CSRF token wasn't present in HTTP headers or POST parameters. Because the request could have been forged, it's been rejected altogether. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection.

<hr> 

* This is important because historically these errors have tripped people up
* Making security as easy as possible is important so that it doesn't just get turned off
* Giving clear directions about where to get the CSRF token (a cookie) and where to send it (a header/param) is especially helpful to frontend developers not necessarily familiar with the backend codebase

<hr>

- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html) (No new APIs)
- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock (No new APIs)

After submitting your PR:

- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts) (Passed on most versions, just timed out on GHC 8.0.2)

### DIFF
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -242,7 +242,7 @@ import           Yesod.Core.Types
 import           Yesod.Routes.Class            (Route)
 import           Blaze.ByteString.Builder (Builder)
 import           Safe (headMay)
-import           Data.CaseInsensitive (CI)
+import           Data.CaseInsensitive (CI, original)
 import qualified Data.Conduit.List as CL
 import           Control.Monad.Trans.Resource  (MonadResource, InternalState, runResourceT, withInternalState, getInternalState, liftResourceT, resourceForkIO)
 import qualified System.PosixCompat.Files as PC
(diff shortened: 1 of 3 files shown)


yesod-core/Yesod/Core/Handler.hs
```diff
@@ -1550,17 +1558,31 @@ checkCsrfHeaderOrParam :: (MonadHandler m, MonadLogger m)
...
 validCsrf (Just _token)     Nothing = False
 
-csrfErrorMessage :: Text
-csrfErrorMessage = "A valid CSRF token wasn't present in HTTP headers or POST parameters. Because the request could have been forged, it's been rejected altogether. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection."
+data CSRFExpectation = CSRFHeader Text (Maybe Text) -- Key/Value
+                     | CSRFParam Text (Maybe Text) -- Key/Value
+
+csrfErrorMessage :: [CSRFExpectation]
+                  -> Text -- ^ Error message
+csrfErrorMessage expectedLocations = "A valid CSRF token wasn't present. Because the request could have been forged, it's been rejected altogether. Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF protection.\nTip: By default, the CSRF token is sent to the client in a cookie named " `mappend` (decodeUtf8 defaultCsrfCookieName) `mappend` ".\nTip: The server is looking for the token in the following locations:\n" `mappend` T.intercalate "\n" (map csrfLocation expectedLocations)
```
sbditto85: Question: is there any reason to have `Check the Yesod.Core.Handler docs of the yesod-core package for details on CSRF ` in the message. Its not something I would want being shown to anyone viewing my site. Is there some other way to get that message to developers and not viewers?

MaxGabriel: I don't think that `yesod-core` has a way of knowing if it's being used for development or production, so I don't think so.

I agree that it's not an ideal error message for a user, but if your users encounter a CSRF error your site is probably broken, which is also a pretty bad user experience. Potential compromise: the sentence could be prefaced with: "If you're a developer of this site"

sbditto85: or possibly a link to a wiki page. Then the information is available and "normal" people will probably just ignore it.

Issue #1454, opened 2017-11-12
`byLabel` fails if multiple label names start with the given label name


pythonissam:
ex.) Call `byLabel "hobby" "fishing"` and if the form has `<label for="hobby">hobby</label>` and `<label for="hobby2">hobby2</label>`. Then, it fails with `"More than one label contained "` **ALTHOUGH THEIR NAMES ARE DIFFERENT**.

`Yesod.Test.byLabel` calls `nameFromLabel` in its definition. `nameFromLabel` looks up the name of a field based on the contents of the label pointing to it. In its definition, label names which start with `label` are assigned into `mlabel`

```haskell
mlabel = parseHTML body
                $// C.element "label"
                >=> contentContains label

contentContains x c
        | x `T.isInfixOf` T.concat (c $// content) = [c]
        | otherwise = []
```

Then, the definition follows:

```haskell
mfor = mlabel >>= attribute "for"

case mfor of
    for:[] -> do
      let mname = parseHTML body
                    $// attributeIs "id" for
                    >=> attribute "name"
      case mname of
        "":_ -> failure $ T.concat
            [ "Label "
            , label
            , " resolved to id "
            , for
            , " which was not found. "
            ]
        name:_ -> return name
        [] -> failure $ "No input with id " <> for
    [] ->
      case filter (/= "") $ mlabel >>= (child >=> C.element "input" >=> attribute "name") of
        [] -> failure $ "No label contained: " <> label
        name:_ -> return name
    _ -> failure $ "More than one label contained " <> label
```

Here, the first example fails. Apparently, the cause is that we used `T.isInfixOf` in `contentContains` and not only `"hobby"`, but also `"hobby2"` is taken because it contains "hobby" at the beginning.

snoyberg:
It would be great if you could submit a PR adding a failing test case for this. Even better if it includes a fix too of course :) Would you be able to take a stab at the tests?

snoyberg:
Since it's a breaking behavioral change, I'd suggest making a new function, `byLabelExact`. Some people may be relying on the current behavior.

MaxGabriel:
I found the current behavior pretty confusing. Possibly deprecating byLabel, and adding new functions (e.g. byLabelContains/byLabelExact, or something like addParamForLabel/addParamForLabelContains) could be better?

snoyberg:
Sounds reasonable to me.

On Wed, Nov 22, 2017 at 3:07 PM, Maximilian Tagher <notifications@github.com

[...]

PR #1459, opened 2017-12-02
Bylabel exact

pythonissam:

- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -680,7 +680,7 @@ byLabelExact = byLabelWithMatch (==)
 -- >   <label>Please submit an image <input type="file" name="f1"> </label>
 -- > </form>
 --
--- Warning: There exists the same issue of 'byLabel'. Please use 'fileByLabelExact' instead.
+-- Warning: This function has the same issue as 'byLabel'. Please use 'fileByLabelExact' instead.
 fileByLabel :: T.Text -- ^ The text contained in the @\<label>@.
             -> FilePath -- ^ The path to the file.
             -> T.Text -- ^ The MIME type of the file, e.g. "image/png".
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -75,6 +75,7 @@ module Yesod.Test
     , byLabel
     , byLabelExact
     , fileByLabel
+    , fileByLabelExact
 
     -- *** CSRF Tokens
     -- | In order to prevent CSRF exploits, yesod-form adds a hidden input
@@ -678,12 +679,46 @@ byLabelExact = byLabelWithMatch (==)
 -- > <form method="POST">
 -- >   <label>Please submit an image <input type="file" name="f1"> </label>
 -- > </form>
+--
+-- Warning: There exists the same issue of 'byLabel'. Please use 'fileByLabelExact' instead.
 fileByLabel :: T.Text -- ^ The text contained in the @\<label>@.
             -> FilePath -- ^ The path to the file.
             -> T.Text -- ^ The MIME type of the file, e.g. "image/png".
             -> RequestBuilder site ()
 fileByLabel label path mime = do
   name <- genericNameFromLabel T.isInfixOf label
+  addFile name path mime
+
+  -- | Finds the @\<label>@ with the given value, finds its corresponding @\<input>@, then adds a file for that input to the request body.
+--
+-- ==== __Examples__
+--
+-- Given this HTML, we want to submit a file with the parameter name @f1@ to the server:
+--
+-- > <form method="POST">
+-- >   <label for="imageInput">Please submit an image</label>
+-- >   <input id="imageInput" type="file" name="f1" accept="image/*">
+-- > </form>
+--
+-- You can set this parameter like so:
+--
+-- > request $ do
+-- >   fileByLabel "Please submit an image" "static/img/picture.png" "img/png"
+--
+-- This function also supports the implicit label syntax, in which
+-- the @\<input>@ is nested inside the @\<label>@ rather than specified with @for@:
+--
+-- > <form method="POST">
+-- >   <label>Please submit an image <input type="file" name="f1"> </label>
+-- > </form>
+--
+-- @since 1.5.9
+fileByLabelExact :: T.Text -- ^ The text contained in the @\<label>@.
+                 -> FilePath -- ^ The path to the file.
+                 -> T.Text -- ^ The MIME type of the file, e.g. "image/png".
+                 -> RequestBuilder site ()
+fileByLabelExact label path mime = do
+  name <- genericNameFromLabel (==) label
   addFile name path mime
 
 -- | Lookups the hidden input named "_token" and adds its value to the params.
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -525,46 +525,6 @@ addFile name path mimetype = do
             MultipleItemsPostData $ ReqFilePart name path contents mimetype : posts
 
 -- This looks up the name of a field based on the contents of the label pointing to it.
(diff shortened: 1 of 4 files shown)


pythonissam:
Issue: https://github.com/yesodweb/yesod/issues/1454

I have no confidence if my English is right. Please correct any sentences you feel unconfortable.
Thank you.

pythonissam:
Well, `fileByLabel` also uses `T.isInfixOf`. Should I create the exact version for this, too?

snoyberg:
Sounds good to me

yesod-test/Yesod/Test.hs
```diff
@@ -523,23 +524,26 @@ addFile name path mimetype = do
           addPostData (MultipleItemsPostData posts) contents =
             MultipleItemsPostData $ ReqFilePart name path contents mimetype : posts
 
+-- |
 -- This looks up the name of a field based on the contents of the label pointing to it.
-nameFromLabel :: T.Text -> RequestBuilder site T.Text
-nameFromLabel label = do
+--
+-- @since 1.5.9
```
MaxGabriel: You don't need to give `@since` annotations for non-exported functions

yesod-test/Yesod/Test.hs
```diff
@@ -566,6 +570,16 @@ nameFromLabel label = do
 (<>) :: T.Text -> T.Text -> T.Text
 (<>) = T.append
 
+-- |
+-- @since 1.5.9
+byLabelWithMatch :: (T.Text -> T.Text -> Bool) -- ^ The matching method which is used to find labels (i.e. exact, contains)
```
MaxGabriel: Is this meant to be exported? It seems really useful

yesod-test/Yesod/Test.hs
```diff
@@ -591,12 +605,57 @@ nameFromLabel label = do
...
+--
+-- >   <label for="nickname2">Nickname2</label>
+-- >   <input id="nickname2" name="f2" />
+-- > </form>
+--
+-- > request $ do
+-- >   byLabel "Nickname" "Snoyberger"
+--
+-- Therefore, this function is deprecated. Please consider using 'byLabelExact',
+-- which performs the exact match over the given word.
```
MaxGabriel: I think we should use a GHC [`DEPRECATED` pragma](https://downloads.haskell.org/~ghc/7.0.3/docs/html/users_guide/pragmas.html), and recommend `byLabelWithMatch Data.Text.contains` for the old behavior and `byLabelExact` for the new behavior

yesod-test/Yesod/Test.hs
```diff
@@ -591,12 +605,57 @@ nameFromLabel label = do
 -- > <form method="POST">
 -- >   <label>Username <input name="f1"> </label>
 -- > </form>
+--
+-- Note that this function finds the labels in which contain the given word.
+-- It might choice labels unexpectedly or just fail in the circumstances like below:
```
MaxGabriel: Can you make this more explicit about how the function would fail? Something like the following:

> Warning: This function looks for any label that contains a subset of the provided text. If multiple labels contain that subset, this function will throw an error, as in the example below:

Site note since it sounds like you're working on your English skills:

"labels in which contain" should be "labels which contain"

"It might choice labels" should be "It might choose labels"

yesod-test/Yesod/Test.hs
pythonissam: No. I understand the rule. I'll do so:)

yesod-test/Yesod/Test.hs
pythonissam: No, neither.

yesod-test/Yesod/Test.hs
pythonissam: > recommend byLabelWithMatch Data.Text.contains for the old behavior and byLabelExact for the new behavior

Actually, `byLabelWithMatch` is the abstraction of the matching methods and `byLabel` performs the old behavior (contain) as before, unless users will be forced to rewrite their tests to use `byLabelContain` or whatever. And yes, `byLabelExact` is for the new behavior.

So,

`byLabel` -> deprecated
`byLabelExact` -> recommanded

Is this okay?

> I think we should use a GHC DEPRECATED pragma

NiceXD. I even didn't know that. I'll modify the code.

yesod-test/Yesod/Test.hs
pythonissam: > Can you make this more explicit about how the function would fail?

Sure.

And thank you for your advice about my English. It helps me a lot.

pythonissam:
@snoyberg Should I create another PR?

snoyberg:
I don't think a new PR is necessary, I'd say just update this one.

pythonissam:
@MaxGabriel Following your advices, I updated haddocks. What do you think?

yesod-test/Yesod/Test.hs
```diff
@@ -620,12 +679,46 @@ byLabel label value = do
 -- > <form method="POST">
 -- >   <label>Please submit an image <input type="file" name="f1"> </label>
 -- > </form>
+--
+-- Warning: There exists the same issue of 'byLabel'. Please use 'fileByLabelExact' instead.
```
MaxGabriel: It would sound better to say "This function has the same issue as 'byLabel'"

MaxGabriel:
I made one small comment. Otherwise LGTM!

pythonissam:
@MaxGabriel Noted with thanks!

pythonissam:
@snoyberg @MaxGabriel I guess I've done. Would you review changes?

MaxGabriel:
@pythonissam what’s the other bug in byLabel?

pythonissam:
@MaxGabriel Please consider the following situation

```html
<html>
  <form>
    <label>foo2<input type=text name="bar2"></label>
    <label>foo <input type=text name="bar" ></label>
  </form>
</html>
```

Then `byLabel "foo" "baz"` sets **baz** into **bar2**, not **bar**.

The cause is around here:

```haskell
-- In genericNameFromLabel

case mfor of
  for:[] -> ...

  [] ->
    case filter (/= "") $ mlabel >>= (child >=> C.element "input" >=> attribute "name") of
      [] -> failure $ "No label contained: " <> label
      name:_ -> return name

  _ -> ...
```

Using `isInfixOf`, `mlabel` will be `["foo2","foo"]`. The above form doesn't contain any `for` attributes, so `mfor` will be an empty list.

Finally, `filter (/= "") $ mlabel >>= (child >=> C.element "input" >=> attribute "name")` becomes `["bar2", "bar"]` and its head is taken.

pythonissam:
Same here, the fundamental cause is that we decided to use `isInfixOf`. So I feel like there's nothing we can do without changing their behaviors.

pythonissam:
@MaxGabriel @snoyberg

Hi. What's the status of this? CI looks like it failed. 
Is there anything I can do with this?

yesod-test/Yesod/Test.hs
```diff
@@ -162,6 +164,8 @@ import GHC.Exts (Constraint)
 type HasCallStack = (() :: Constraint)
 #endif
 
+{-# DEPRECATED byLabel "This function seems to have multiple bugs. Use byLabelExact instead" #-}
```
snoyberg: How about including a link to this PR in the deprecation notices?

pythonissam: It sounds great.

pythonissam:
@snoyberg Is this okay?

snoyberg:
Looks great, thanks!

MaxGabriel:
Thanks @pyhtonissam!

pythonissam:
@snoyberg @MaxGabriel Yay! Thank you, too!

PR #1460, opened 2017-12-06
Switch CoC to Contributor Covenant

snoyberg:
If anyone has concerns with this switch, please let me know. Otherwise I'll merge to master some time next week.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,13 +1,74 @@
-# Contributor Code of Conduct
+# Contributor Covenant Code of Conduct
 
-Always be nice.
+## Our Pledge
 
-When communicating online treat people the way you would if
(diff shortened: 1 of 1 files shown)


bitemyapp:
Think this is a great idea, somebody already went to the trouble of hammering out a sound CoC, might as well use it. The enforcement documentation linked on the website seemed good too, e.g. https://www.djangoproject.com/conduct/enforcement-manual/

afcady:
This was already merged, and I'm not even working on this project (though I do use it) but I will still say I found this passage a bit disturbing:

    The project team is obligated to maintain confidentiality with regard to the reporter of an incident.

The legal principle of  the right to confront one's accuser was adopted for a reason.  The anonymity of the "reporter" means that "reports" can be outright invented and can never be disputed when false.

PR #1462, opened 2017-12-08
Fixed exposed version number and also haddock syntax for Yesod.Auth module

psibi:

- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

### DIFF
--- a/yesod-auth/Yesod/Auth.hs
+++ b/yesod-auth/Yesod/Auth.hs
@@ -85,7 +85,7 @@ type Piece = Text
 
 -- | The result of an authentication based on credentials
 --
--- Since 1.4.4
+-- @since 1.4.4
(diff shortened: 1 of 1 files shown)


yesod-auth/Yesod/Auth.hs
```diff
@@ -184,7 +184,8 @@ class (Yesod master, PathPiece (AuthId master), RenderMessage master FormMessage
 
     -- | When being redirected to the login page should the current page
     -- be set to redirect back to. Default is 'True'.
-    -- @since 1.4.18
+    --                      
+    -- @since 1.4.21
```
snoyberg: Thanks for catching this, my mistake :(

Issue #1418, opened 2017-07-13
Header type equality comparison


psibi:
The equality comparison is case insensitive according to the HTTP specs. Right now, we don't make that distinction.

snoyberg:
Header keys are wrapped in the `CI` type constructor, which ensures a case insensitive comparison. Is there a specific counterexample?

psibi:
@snoyberg I think this `Header` is not wrapped in `CI` type constructor: https://www.stackage.org/haddock/lts-8.22/yesod-core-1.4.35/Yesod-Core.html#t:Header

psibi:
I think it would be better to just wrap it in `CI` type constructor instead of doing `Eq` instance fix.

paul-rouse:
At the moment the code in #1417 would still need to be changed to use Eq on Header - does this actually go any wider than that code, or would it be best just to use CI for case-insensitivity there, for that one comparison?

psibi:
I would vote for using `CI` for case-insensitivity there as that's the most easiest thing to do. :)

snoyberg:
@psibi Up until now, there has been no comparison performed there that I'm aware of. 20/20 hindsight, it would make sense to use `CI ByteString`, but I'd rather not break the API for this.

PR #1464, opened 2017-12-12
Better monads

snoyberg:


MaxGabriel:
Should the docs (“While this is simply a @WriterT@”) be updated too?

snoyberg:
Hah, I didn't expect anyone to be reviewing this. Good catch. I'm not quite sure which (if any) of these experiments will end up making it onto master, so feedback on whether this is worth it at all is certainly welcome :). I'm currently experimenting with a pretty crazy subsite rewrite based on all of this, but it's not ready to be pushed, and I probably won't finish it until at least next week due to travel.

psibi:
While you are doing some breaking changes, I think this is a good time to fix this also: https://github.com/yesodweb/yesod/issues/1418#issuecomment-315051642

snoyberg:
Good call @psibi, added in 915d9e2f

PR #1466, opened 2017-12-18
Drop the transformers

snoyberg:
This includes the additional change mentioned at the end of https://www.yesodweb.com/blog/2018/01/upcoming-yesod-breaking-changes, swapping `HandlerT` and `WidgetT` for `HandlerFor` and `WidgetFor`.


snoyberg:
Pinging @yuras, this is what I was telling you about yesterday.

PR #1475, opened 2018-01-16
remove postLogoutR

TobyGoodwin:
A colleague was checking a yesod-based website and asked:
> What's the odd interstitial page with a "continue" button when I log out?

A good question! After peering at it for a while, I can't see any reason for the use of `redirectToPost` here, nor even the existence of `postLoginR`.

Checking the history of these lines, they date from the 2010 commit that introduced the `yesod-auth` package. So I suspect this was just fluff in the initial implementation, rather than anything to solve a real world problem.

### DIFF
--- a/yesod-auth/ChangeLog.md
+++ b/yesod-auth/ChangeLog.md
@@ -1,6 +1,6 @@
 ## unreleased
 
-* Avoid `redirectToPost` on logout
+* Avoid `redirectToPost` on logout [#1475](https://github.com/yesodweb/yesod/pull/1475)
 
 ## 1.4.21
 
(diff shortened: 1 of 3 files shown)


snoyberg:
This is quite intentional as-is. We don't want GET requests mutating state. It's best for logout buttons to actually be POST forms, but since many people want to make them as normal links, have `redirectToPost` handle it is a middle ground.

TobyGoodwin:
Ah. Good point. Sorry for the noise.

[ Hazards of being a not-yet-good-enough "full stack" developer... I've spent so long styling links as buttons that I'm forgetting they're supposed to different! :) ]

snoyberg:
No worries, it's a subtle point.

PR #1476, opened 2018-01-18
[yesod] Fix comment for contentTypeTypes & simpler implementation

JaSpa:
In the implementation of `contentTypeTypes` there is no need to repeat the logic from `simpleContentType` which is defined just above it.

----

After submitting your PR:


snoyberg:
Thanks, I've merged this into #1464.

Issue #1365, opened 2017-03-24
Extending mkYesodSubData


jprider63:
For background, please see [this post](https://groups.google.com/forum/#!topic/yesodweb/DTDcDPT7Wps) on the google group. To solve the type variable not being in scope problem, I've deviated from my original plan of using a `&{}` syntax. Instead I've:

- [Updated](https://github.com/jprider63/yesod/commit/80622c2a7fe664a83eac38fe96780b9880865abd) `mkYesodGeneral` so that it can accept type variables as well. For example, you can now do: `mkYesodSubData "MySubsite master"`. This also fixed a non-exhaustive pattern matching bug. 
- [Updated](https://github.com/jprider63/yesod/commit/deb72b30db18e429c60f69314746f71456efa5dd) `ttToType` so that type variables can be parsed in routes. This now can look something like: `/#{MyDataId master} MyHandlerR GET`
- [Added contexts](https://github.com/jprider63/yesod/commit/f5376ef7ca4c8fa6ddded1e6f7a1466d2d92a804) as an argument to `mkYesodGeneral'`. These contexts are added to the instances generated. When there are contexts, I switched to use standalone deriving instead of regular deriving since I was running into [this issue](http://stackoverflow.com/questions/13144740/deriving-instances-with-typefamilies). 

So far, it seems like everything is working as I'd like. You can see what I currently have on my vars [branch](https://github.com/yesodweb/yesod/compare/master...jprider63:vars). I don't think there are any breaking changes (I'm not sure if every template-haskell version works properly though).

I'd like to make a few more changes.

- Implement a context parser so that we can just write `mkYesodSubData "(MyClass master) => MySubsite master"` instead of passing in a manually constructed `Cxt`. 
- Change the `[Either String [String]]` to `[String]` in `mkYesodGeneral` since the context parser would subsume this functionality. 

Do you have any thoughts before I continue?

jprider63:
I just created PR #1366. This implements the context parser and everything mentioned above it. There shouldn't be any breaking changes. 

I didn't change `mkYesodGeneral` since this would be a breaking change with `mkYesodWith` exported. I can create a separate pull request if you'd like to save this change for the next major version bump.

jezen:
@jprider63 This is awesome. Thank you for the implementation, and for documenting an example in the cookbook.

I know you did this rather a long time ago, but do you happen to remember how you wrote the master site side of this? There isn't an example of that part in the cookbook.

[...]

PR #1478, opened 2018-01-22
Update `mkYesodWith` and refactor so that `mkYesod` uses the context parser

jprider63:
Since it looks like you're planning on introducing breaking changes, it seems like a good time for this. This PR is breaking since it updates the type of `mkYesodWith` to separate the list of contexts from the list of type arguments. This makes `mkYesodWith` more general since you can now have multiple arguments for contexts. For example:

```
data Foo a b = Foo a b

class Bar a b where
  ...

mkYesodWith [["Bar", "a", "b"]] "Foo" ["a", "b"] ...
```

This should generate something like:

```
instance Bar a b => YesodDispatch (Foo a b) where
    ...
```

I also updated `mkYesod` to use the context parser and deprecated `mkYesodWith` since you should be able to get the same functionality from `mkYesod` now. Let me know if you don't want to deprecate this. 

`mkYesodWith` was originally introduced in #1055. @Daniel-Diaz, does this look good to you? 

This should finish #1365.

After submitting your PR:

### DIFF
--- a/yesod-core/Yesod/Core/Internal/TH.hs
+++ b/yesod-core/Yesod/Core/Internal/TH.hs
@@ -44,11 +44,13 @@ mkYesod :: String -- ^ name of the argument datatype
         -> Q [Dec]
 mkYesod name = fmap (uncurry (++)) . mkYesodGeneral name [] False return
 
-mkYesodWith :: String
-            -> [Either String [String]]
+{-# DEPRECATED mkYesodWith "Contexts and type variables are now parsed from the name. (https://github.com/yesodweb/yesod/pull/1366)" #-}
+mkYesodWith :: [[String]]
+            -> String
+            -> [String]
             -> [ResourceTree String]
             -> Q [Dec]
-mkYesodWith name args = fmap (uncurry (++)) . mkYesodGeneral name args False return
+mkYesodWith cxts name args = fmap (uncurry (++)) . mkYesodGeneral' cxts name args False return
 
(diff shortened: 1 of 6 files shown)


jprider63:
I've changed the target to better-monads. There aren't any conflicts, but I don't think the tests reran.

Daniel-Diaz:
@jprider63 I'm OK with more flexibility, but maybe it would be good to have a note on how to migrate current code? Or is that not necessary?

jprider63:
I doubt it's necessary since this shouldn't impact the majority of users, but I added some more documentation to `mkYesod` and `mkYesodWith`. There's also a more detailed example in the [cookbook](https://github.com/yesodweb/yesod-cookbook/blob/master/cookbook/Subsite-with-typeclass-contexts-and-type-variables.md).

It looks like `mkYesodSub` was removed, so I changed `mkYesod`'s documentation to point to `mkYesodSubData` and `mkYesodSubDispatch` instead.

Daniel-Diaz:
@jprider63 Well, I do use this function, and I would not assume how many people use it. And even a minority deserves a pleasant migration. :) Anyway, thanks for extending the docs! It looks good to me.

snoyberg:
@Daniel-Diaz Just to confirm: are you OK with this change making it into yesod-core 1.6?

Daniel-Diaz:
@snoyberg Yes.

snoyberg:
Cool, thanks!

jprider63:
Thanks!

PR #1485, opened 2018-02-06
yesodSpecWithSiteGenerator with an argument

NorfairKing:

After submitting your PR:

^ I'll complete the checkboxes if you approve of the change.

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -34,7 +34,7 @@ module Yesod.Test
       yesodSpec
     , YesodSpec
     , yesodSpecWithSiteGenerator
-    , yesodSpecWithSiteGenerator'
+    , yesodSpecWithSiteGeneratorAndArgument
     , yesodSpecApp
     , YesodExample
     , YesodExampleData(..)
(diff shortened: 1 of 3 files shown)


yesod-test/Yesod/Test.hs
```diff
@@ -276,6 +277,24 @@ yesodSpecWithSiteGenerator getSiteAction yspecs =
             , yedResponse = Nothing
             }
 
+yesodSpecWithSiteGenerator' :: YesodDispatch site
```
snoyberg: What's the purpose of this function? Why would I use this function instead of `yesodSpecWithSiteGenerator`? We definitely need a Haddock for this. (Also, when written, we need a `@since` comment, ChangeLog entry, and cabal file version bump.)

NorfairKing:
Let's first see if we can agree that this addition is useful.

> What's the purpose of this function?
> Why would I use this function instead of yesodSpecWithSiteGenerator?

The use-case is as follows:

I have a `Yesod` site that contains the information necessary to call an external web service: Url and port, for example.
I am also the author of the external web service and would like to test how these two interact.
I would like to use [`testWithApplicationSettings`](https://hackage.haskell.org/package/warp-3.2.12/docs/Network-Wai-Handler-Warp.html#v:testWithApplicationSettings) to start the web service around the test.
That means I need something like `SpecWith Wai.Application`.
The port automatically gets chosen when this application is started, so the Yesod site needs to be made with the `IO Site` function only after this port is available.
This is why I need an argument to the `IO Site` function. (It is the yesod equilavent of [`beforeWith`](https://hackage.haskell.org/package/hspec-2.5.0/docs/Test-Hspec.html#v:beforeWith)

I hope the above clears up the need and the purpose of this new function.
I am very open to changing the specifics as long as my use-case can be supported.

> We definitely need a Haddock for this. (Also, when written, we need a @since comment, ChangeLog entry, and cabal file version bump.)

Will do that if you agree that this is a useful change.

snoyberg:
I wasn't objecting to the inclusion of the function, something which takes an argument like this seems fine. But it needs a more descriptive name and Haddocks.

Also, while we're at it: can't `yesodSpecWithSiteGenerator` be rewritten to use the new function, and avoid some code duplication?

NorfairKing:
@snoyberg Ready for more review.

snoyberg:
Thanks!

PR #1492, opened 2018-03-04
Derive Show instances for route data structures

MaxGabriel:
* It's very helpful to have a Show instance for debugging and development
* Currently third party packages are deriving this instance themselves which is not ideal.
    * http://hackage.haskell.org/package/yesod-routes-flow-2.0/docs/src/Yesod-Routes-Flow-Generator.html
    * http://hackage.haskell.org/package/yesod-routes-typescript-0.3.0.0/docs/src/Yesod-Routes-Typescript-Generator.html
    * This change would break those packages, which isn't great
         * At least the typescript one is broken anyway

I haven't looked into this much, but if someone knows offhand, what is the use case for `Piece` and `Dispatch` being parametrized data types? It seems like they would most always be `String`?

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,8 @@
+## 1.6.2
+
+* Derive a `Show` instance for `ResourceTree` and `FlatResource` [#1492](https://github.com/yesodweb/yesod/pull/1492)
+	* Some third party packages, like `yesod-routes-flow` derive their own `Show` instance, and this will break those packages.
+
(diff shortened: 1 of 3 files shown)


MaxGabriel:
Since `yesod-routes-flow` is being actively developed, I think a good solution is to have a preprocessor macro that says that if yesod-core is >= this version, then don't derive the show instances.

MaxGabriel:
Honestly seems like I should also basic typeclasses like `Eq` and `Ord` as well.

PR #1494, opened 2018-03-12
Add JSONResponse type

### LINKED ISSUE #1481: JSON return type for Handlers?
So I was thinking about this blog post recently http://tech.frontrowed.com/2015/12/21/servant-style-handlers-for-yesod/, that uses `UndecidableInstances` to let you specify the return type of your Handlers that return JSON as `Handler <RecordName>` rather than `Handler Value`

Why not just have something like this?

```haskell
data JSONResponse a = ToJSON a => JSONResponse a

instance ToContent (JSONResponse a) where
  toContent (JSONResponse a) = toContent $ toEncoding a

instance ToTypedContent (JSONResponse a) where
  toTypedContent = TypedContent typeJson . toContent
```

Then handlers are typed as:

```haskell
postSignupR :: Handler (JSONResponse CreateUserResponse)
```

and the function returns: 
```haskell
return $ JSONResponse $ CreateUserResponse userId
```

Slightly more overhead than the blog post version. I don’t know if `UndecidableInstances` is actually worth avoiding though. Thoughts? I think endorsing one or the other of these options would be good for Yesod to do—it seems much better than returning `Value`.

MaxGabriel:
Closes #1481

After submitting your PR:

### DIFF
--- a/yesod-core/Yesod/Core/Types.hs
+++ b/yesod-core/Yesod/Core/Types.hs
@@ -308,6 +308,18 @@ newtype RepXml = RepXml Content
 
 type ContentType = ByteString -- FIXME Text?
 
+-- | Wrapper around types so that Handlers can be typed them, encoded as JSON.
+-- Example usage in a type signature:
+-- 
+-- > postSignupR :: Handler (JSONResponse CreateUserResponse)
+--
+-- And in the implementation:
+--
+-- > return $ JSONResponse $ CreateUserResponse userId
+--
+-- @since 1.6.3 
+data JSONResponse a = ToJSON a => JSONResponse a
+
 -- | Prevents a response body from being fully evaluated before sending the
 -- request.
 --
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.6.3
+
+* Add a type `JSONResponse a` that can be used to type a Handler as returning some value `a`, which should be encoded as JSON. [#1481](https://github.com/yesodweb/yesod/issues/1481)
+
 ## 1.6.2
 
 * Derive a `Show` instance for `ResourceTree` and `FlatResource` [#1492](https://github.com/yesodweb/yesod/pull/1492)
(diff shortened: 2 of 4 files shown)


yesod-core/Yesod/Core/Types.hs
```diff
@@ -308,6 +308,18 @@ newtype RepXml = RepXml Content
...
+-- Example usage in a type signature:
+-- 
+-- > postSignupR :: Handler (JSONResponse CreateUserResponse)
+--
+-- And in the implementation:
+--
+-- > return $ JSONResponse $ CreateUserResponse userId
+--
+-- @since 1.6.3 
+data JSONResponse a = ToJSON a => JSONResponse a
```
snoyberg: I think it would be better without the datatype context `ToJSON a =>` here. I don't remember all of the arguments, but it's typically considered a bad extension.

MaxGabriel: Hm, this gives some reasons https://wiki.haskell.org/Data_declaration_with_constraint

Haven’t tried this, but the wiki recommends replacing with a GADT

snoyberg: I don't think a GADT is necessary here. I'd replace with just:

```haskell
newtype JSONResponse a = JSONResponse a
```

Is anything lost in such a change?

snoyberg: Does the GADT approach solve that?

StevenXL:
The version with GADTs works pretty well. The image below shows the declaration of the `JSONResponse` type constructor using GADTs, the type signature of the `JSONResponse` data constructor, and the error when you try apply `JSONResponse` to a value that is not an instance of `ToJSON`. I think it works out very well. 

@MaxGabriel Given that it has been a long time since your PR, I'll make another PR with GADTs and reference this as well as #1481. (I can always back off if you say so. Thank you). 

Screen Shot 2019-04-12 at 3 18 54 PM

StevenXL:
Closed by #1592

PR #1499, opened 2018-03-28
1496 add failure

pythonissam:

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -102,6 +102,7 @@ module Yesod.Test
     , assertNotEq
     , assertEqualNoShow
     , assertEq
+    , assertFailure
 
     , assertHeader
     , assertNoHeader
@@ -382,6 +383,12 @@ assertEqual = assertEqualNoShow
 -- @since 1.5.2
 assertEqualNoShow :: (HasCallStack, Eq a) => String -> a -> a -> YesodExample site ()
 assertEqualNoShow msg a b = liftIO $ HUnit.assertBool msg (a == b)
+
+-- | Unconditionally signals that a failure has occured.
+--
+-- @since 1.6.3
+assertFailure :: HasCallStack => String -> YesodExample site ()
+assertFailure = liftIO . HUnit.assertFailure
 
 -- | Assert the last response status is as expected.
 statusIs :: HasCallStack => Int -> YesodExample site ()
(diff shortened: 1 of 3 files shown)


pythonissam:
ref: 1496

yesod-test/Yesod/Test.hs
```diff
@@ -383,6 +384,12 @@ assertEqual = assertEqualNoShow
 assertEqualNoShow :: (HasCallStack, Eq a) => String -> a -> a -> YesodExample site ()
 assertEqualNoShow msg a b = liftIO $ HUnit.assertBool msg (a == b)
 
+-- | Unconditionally signals that a failure has occured.
+--
+-- @since 1.6.3
```
snoyberg: Thank you for including the `@since`! Can you also add a `ChangeLog` entry?

pythonissam: Is this okay??

pythonissam:
Is this okay?

snoyberg:
Sorry, there actually is a problem with the PR, CI is failing:

https://travis-ci.org/yesodweb/yesod/jobs/359647053

pythonissam:
@snoyberg Well, I understand...

pythonissam:
```
test/EmbedProductionTest.hs:41:28: error:
    Ambiguous occurrence ‘assertFailure’
    It could refer to either ‘Test.HUnit.assertFailure’,
                             imported from ‘Test.HUnit’ at test/EmbedProductionTest.hs:11:20-32
                             (and originally defined in ‘Test.HUnit.Lang’)
                          or ‘Yesod.Test.assertFailure’,
                             imported from ‘Yesod.Test’ at test/EmbedProductionTest.hs:15:1-17
```

In this case, which should we modify?  The name of our custom `assertFailure` or the test which uses the original `assertFailure`.

snoyberg:
I think we need to rename this function, in order to avoid a more general baking conflict with a dependency.

pythonissam:
@snoyberg How about this one: `assertFlr`

snoyberg:
I'm not a fan of the abbreviated naming. Let me ask a more basic question: why is `assertFailure` better than just using `fail` or `error`?

pythonissam:
> I'm not a fan of the abbreviated naming

I understand.

> why is assertFailure better than just using fail or error?

Using `assertFailure`, we can obtain more detailed information. For example,

`error` and `fail`:

``````

snoyberg:
That looks like the same information displayed in different formats to me.

pythonissam:
Hmm, come to think of it, you're right... I'm closing this issue.

PR #1503, opened 2018-04-17
added: fileSourceByteString

ncaq:
This function is to get `FileInfo` raw body.

Conduit type is hard for begginer user.
* [haskell - Persisting an uploaded file in the database (Couldn't match type ‘ConduitM () ByteString (ResourceT IO) ()’ with ‘HandlerT App IO ByteString’) - Stack Overflow](https://stackoverflow.com/questions/34969335/persisting-an-uploaded-file-in-the-database-couldnt-match-type-conduitm-by)
* [Processing an uploaded file with and embedding it in a template - Google グループ](https://groups.google.com/forum/?hl=ja#!searchin/yesodweb/fileSource|sort:date/yesodweb/zeC_ZzDeg6E/EkXVISTdhAIJ)

Actually I was not sure when I was a Haskell beginner.

ByteString raw body of `FileInfo` is need to upload file to S3.
This function running our product.
I send it to upstream.

After submitting your PR:

### DIFF
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -253,7 +253,8 @@ import           Data.CaseInsensitive (CI, original)
 import qualified Data.Conduit.List as CL
 import           Control.Monad.Trans.Resource  (MonadResource, InternalState, runResourceT, withInternalState, getInternalState, liftResourceT, resourceForkIO)
 import qualified System.PosixCompat.Files as PC
-import           Data.Conduit (ConduitT, connect, transPipe, Flush (Flush), yield, Void)
+import           Conduit ((.|), runConduit)
+import           Data.Conduit (ConduitT, transPipe, Flush (Flush), yield, Void)
 import qualified Yesod.Core.TypeCache as Cache
 import qualified Data.Word8 as W8
 import qualified Data.Foldable as Fold
@@ -1369,7 +1370,7 @@ fileSource = transPipe liftResourceT . fileSourceRaw
 --
 -- @since 1.6.4
 fileSourceByteString :: MonadResource m => FileInfo -> m S.ByteString
-fileSourceByteString fileInfo = fileSource fileInfo `connect` CL.foldMap id
+fileSourceByteString fileInfo = runConduit $ fileSource fileInfo .| CL.foldMap id
 
 -- | Provide a pure value for the response body.
 --
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -47,6 +47,7 @@ module Yesod.Core.Handler
     , fileName
     , fileContentType
     , fileSource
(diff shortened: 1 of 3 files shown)


yesod-core/Yesod/Core/Handler.hs
```diff
@@ -1360,6 +1361,16 @@ rawRequestBody = do
...
 
+-- | Strict `ByteString` body from `FileInfo`.
+-- This function blocking while read file.
+--
+-- > do
+-- >     fileByteString <- fileSourceByteString fileInfo
+--
+-- @since 1.6.4
+fileSourceByteString :: MonadResource m => FileInfo -> m S.ByteString
+fileSourceByteString fileInfo = fileSource fileInfo `connect` CL.foldMap id
```
snoyberg: There are a few problems with this implementation:

1. `CL.foldMap id` has quadratic complexity. It would be better to use something like [`sinkLazy`](https://www.stackage.org/haddock/lts-11.5/conduit-1.3.0.2/Conduit.html#v:sinkLazy)
2. It's more common to use `runConduit $ src .| sink`

How about:

```haskell
runConduit $ fileSource fileInfo .| Conduit.

File: yesod-core/Yesod/Core/Handler.hs [hunk 2]
```diff
@@ -1360,6 +1361,16 @@ rawRequestBody = do
 fileSource :: MonadResource m => FileInfo -> ConduitT () S.ByteString m ()
 fileSource = transPipe liftResourceT . fileSourceRaw
 
+-- | Strict `ByteString` body from `FileInfo`.
+-- This function blocking while read file.
```
snoyberg: Can you reword this comment to something like:

```
Extract a strict `ByteString` body from a `FileInfo`.

This function will block while reading the file.
```

yesod-core/Yesod/Core/Handler.hs
ncaq: I use `runConduit`.
But `fileSource` function result strict `ByteString`.
So I think can't use `sinkLazy`.
Do you think?

snoyberg: You'd have to use `Data.ByteString.toStrict`, but that's more efficient than using `foldMap`.

ncaq:
@snoyberg I think that `sinkLazy` can't use this case.
Do you think?

ncaq:
@snoyberg 
I use `sinkLazy`.
However, Maybe `fileSourceByteString` return Lazy `ByteString` is great than Strict?

snoyberg:
`foldC` has the same quadratic behavior issues as `foldMapC`.

ncaq:
@snoyberg 
Sorry, What are "complexity" mean code clearing or performance?
I understood to code clearing.
So, I try to write clear code.
But, Are you speak performance?

snoyberg:
It refers to the performance.

ncaq:
@snoyberg I rewrite.

PR #1504, opened 2018-04-17
added: addContentDispositionFileName

ncaq:

After submitting your PR:

I pushed other pull request, I checkbox after complete other pull request.

This function is setting attchment file name.
Setting is hard that allow UTF-8 multibyte character.
This function running our product.
I want to push upstream.

### DIFF
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -118,6 +118,7 @@ module Yesod.Core.Handler
     , setHeader
     , replaceOrAddHeader
     , setLanguage
+    , addContentDispositionFileName
       -- ** Content caching and expiration
     , cacheSeconds
     , neverExpires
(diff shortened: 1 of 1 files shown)


yesod-core/Yesod/Core/Handler.hs
```diff
@@ -780,6 +781,14 @@ deleteCookie a = addHeaderInternal . DeleteCookie (encodeUtf8 a) . encodeUtf8
 setLanguage :: MonadHandler m => Text -> m ()
 setLanguage = setSession langKey
 
+-- | Set attachment file name.
+--
+-- allow UTF-8 character.
```
snoyberg: This should probably say something like: "Allows Unicode characters by encoding to UTF-8"

yesod-core/Yesod/Core/Handler.hs
```diff
@@ -780,6 +781,14 @@ deleteCookie a = addHeaderInternal . DeleteCookie (encodeUtf8 a) . encodeUtf8
 setLanguage :: MonadHandler m => Text -> m ()
 setLanguage = setSession langKey
 
+-- | Set attachment file name.
```
snoyberg: Can you add a ChangeLog entry, cabal version bump, and a `@since` comment?

yesod-core/Yesod/Core/Handler.hs
```diff
@@ -780,6 +781,14 @@ deleteCookie a = addHeaderInternal . DeleteCookie (encodeUtf8 a) . encodeUtf8
...
+--
+-- allow UTF-8 character.
+addContentDispositionFileName :: MonadHandler m => T.Text -> m ()
+addContentDispositionFileName name
+    = addHeader "Content-Disposition" $
+      "attachment; filename*=UTF-8''" `mappend` decodeUtf8 (H.urlEncode True (encodeUtf8 name))
```
snoyberg: I haven't seen this syntax before. Can you reference the relevant spec that details how this is supposed to work in a comment?

yesod-core/Yesod/Core/Handler.hs
ncaq: I thinked that wait other pull request.
However, If I should write before it, I write.
So I write.

ncaq:
@snoyberg I wrote comment, changelog.

PR #1505, opened 2018-04-17
updated: stack: resolver: lts-8.12 -> lts-11.5

ncaq:
[th-lift-instances](https://github.com/bennofs/th-lift-instances) cause build error on appveyor on Windows.
This bug fixed [230839c](https://github.com/bennofs/th-lift-instances/commit/230839cb3460f30e61812f125e1ef9c3f26f9fc4).
This commit goal is to fix to appveyor build.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-8.12
+resolver: lts-11.5
 packages:
   - ./yesod-core
   - ./yesod-static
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks! I have a feeling that removing the extra-deps is going to break Travis, so perhaps a simpler solution would be to just add `tf-lift-instances` as its own extra-dep.

ncaq:
I understand simpler solution.
I close this pull request.

PR #1507, opened 2018-04-29
Cleaned up the stack.yaml

bitemyapp:
Don't need all the extra-dep specification if we use a current LTS snapshot. Should lend anyone that's compiled something recently better caching if they work on the repo too.


snoyberg:
I didn't realize you were referring to the extra-deps. Removing those breaks the CI build. This could be worked around by having a separate stack.yaml with the extra-deps for CI.

bitemyapp:
@snoyberg I usually make explicitly marked Stack yamls for the GHC version intended, e.g.

https://github.com/bitemyapp/bloodhound/blob/master/stack-8.2.yaml

https://github.com/bitemyapp/bloodhound/blob/master/stack-8.0.yaml

Then `stack.yaml` is a symlink to the latest stable. How does that catch your fancy?

snoyberg:
I don't usually use that approach, but I'm not opposed to it. I'll accept it if it works with CI.

bitemyapp:
@snoyberg https://github.com/bitemyapp/bloodhound/blob/master/.travis.yml this is what our CI looks like and probably demonstrates why we do it like that.

psibi:
Another approach which I like recently is using packcheck for these things: https://github.com/harendra-kumar/packcheck

bitemyapp:
@psibi I've reviewed packcheck and I am unlikely to use it.

PR #1510, opened 2018-05-11
Move 'addClass' to Yesod.Form.Functions and add 'removeClass'

bsima:
'addClass' is more general than just Bootstrap forms. In particular, it is
copied into the yesod-form-bootstrap4 project and I found myself using it in my
custom forms. It would be useful to have it exported for use elsewhere.

I added 'removeClass' because I needed it while creating a custom 'readonly'
input in a form and thought it might be generally useful.

---

After submitting your PR:

### DIFF
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -623,9 +623,8 @@ convertField to from (Field fParse fView fEnctype) = let
 --
 -- ==== __Examples__
 --
--- > removeFormControl :: FieldSettings site -> FieldSettings site
--- > removeFormControl fs = fs { fsAttrs = newAttrs }
--- >   where newAttrs = removeClass "form-control" (fsAttrs fs)
+-- >>> removeClass "form-control" [("class","form-control login-form"),("id","home-login")]
+-- [("class","  login-form"),("id","home-login")]
 --
 -- @since 1.6.2
 removeClass :: Text -- ^ The class to remove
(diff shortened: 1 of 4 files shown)


yesod-form/Yesod/Form/Functions.hs
```diff
@@ -615,3 +618,23 @@ convertField to from (Field fParse fView fEnctype) = let
   fParse' ts = fmap (fmap (fmap to)) . fParse ts
   fView' ti tn at ei = fView ti tn at (fmap from ei)
   in Field fParse' fView' fEnctype
+
+-- | Removes a CSS class from the 'fsAttrs' in a 'FieldSettings'.
+--
```
snoyberg: I think this would be a great place for a concrete usage example, it's unclear from the docs or signature exactly what the behavior of this is supposed to be.

yesod-form/Yesod/Form/Functions.hs
```diff
@@ -615,3 +618,23 @@ convertField to from (Field fParse fView fEnctype) = let
...
+-- @since 1.6.2
+removeClass :: Text -- ^ The class to remove
+            -> [(Text, Text)] -- ^ List of existing 'fsAttrs'
+            -> [(Text, Text)]
+removeClass _     []                    = []
+removeClass klass (("class", old):rest) = ("class"::Text, T.replace klass " " old) : rest
+removeClass klass (other         :rest) = other : removeClass klass rest
+
+-- | Adds a CSS class to the 'fsAttrs' in a 'FieldSettings'.
+--
```
snoyberg: Same comment about a usage example.

bsima:
Done! Do you think what I added is a good example? It's not too
complicated, just list manipulation anyways.

snoyberg:
I'd actually been hoping to see the input and output of the function calls. It's certainly helpful to see how to use it, but it's still not clear from the docs what the expected result would be.

bsima:
Oh okay, I'll add ghci examples

Issue #1495, opened 2018-03-25
Show actual next to expected in yesod tests


Qwertie-:
In my test I have
` bodyEquals "[{\"name\":\"a\",\"id\":1,\"desc\":null}]"`
and while running the test it fails with the error:
`Expected body to equal [{"name":"a","id":1,"desc":null}] `
That's all fine but after seeing that the next thing I want to know is what was the body if it wasn't that.
Other test systems show the actual next to the expected which saves you having to go back and add a print in to the tests to see what went wrong.

Qwertie-:
For `bodyContains` there might be large responses but I think if you are using `bodyEquals` you are likely testing an API with a fairly small response because you have to fit the body in your test.

snoyberg:
Fair enough. In that case, PR to change the implementation is welcome.

StevenXL:
Closed due to merge of #1525.

PR #1525, opened 2018-06-14
bodyEquals prints actual body.

StevenXL:

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -138,6 +138,7 @@ import qualified Data.ByteString.Char8 as BS8
 import Data.ByteString (ByteString)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as TE
+import qualified Data.Text.Encoding.Error as TErr
 import qualified Data.ByteString.Lazy.Char8 as BSL8
 import qualified Test.HUnit as HUnit
 import qualified Network.HTTP.Types as H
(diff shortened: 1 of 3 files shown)


StevenXL:
Hoping to address #1495.

yesod-test/Yesod/Test.hs
```diff
@@ -439,9 +439,10 @@ assertNoHeader header = withResponse $ \ SResponse { simpleHeaders = h } ->
 -- | Assert the last response is exactly equal to the given text. This is
 -- useful for testing API responses.
 bodyEquals :: HasCallStack => String -> YesodExample site ()
-bodyEquals text = withResponse $ \ res ->
-  liftIO $ HUnit.assertBool ("Expected body to equal " ++ text) $
-    (simpleBody res) == encodeUtf8 (TL.pack text)
+bodyEquals text = withResponse $ \ res -> do
+  let actual = simpleBody res
+      msg    = concat ["Expected body to equal ", text, ". Actual is: ", BSL8.unpack actual]
```
snoyberg: Would it make sense for display purposes to put `text` and `actual` on their own lines? Also, I think it would be better to use UTF8 decoding (maybe via `decodeUtf8With lenientDecode`) instead of Char8.

StevenXL: Changes made. Let me know what you think.

PR #1527, opened 2018-06-19
cleaned: appveyor conf

### LINKED ISSUE #1523: Broken Windows build
Carried over from https://github.com/yesodweb/yesod/pull/1520#issuecomment-396116235. CC @ncaq @StevenXL.

The AppVeyor build is broken, and I don't see an indication of what's breaking it. @ncaq can you share thoughts on what might be causing it?

ncaq:
* use network-2.6.3.3, network-2.6.3.4 include only fix to OpenBSD and **bug** and document
* appveyor.yml conf file to simple from offcial reference
* AppVeyor split build and test, So we can easy detect build or test error
* AppVeyor test job to single thread, So we can easy find bug of test

↓ I fix only CI, I think to do not need bump version.

After submitting your PR:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,19 +1,15 @@
+# cache:
+#   - '%AppData%\stack'
+
+install:
+  - curl -sS -ostack.zip -L https://get.haskellstack.org/stable/windows-i386.zip
+  - 7z x stack.zip stack.exe
+  - stack setup > nul
+
 build: off
 
-before_test:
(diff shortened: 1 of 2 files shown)


ncaq:
Oh, I wasted d38d00f11400f3c41a528c11315ed2278da385c7

ncaq:
In the case of Windows Since processing to ignore the test entered,
There was no need to roll back the network package.
However, I rewrote appveyor's configuration file seriously.
I think this difference is useful, so leave a pull request.
I change the title.

PR #1528, opened 2018-06-19
cleaned: appveyor: Easier to detect bugs

ncaq:
* appveyor.yml conf file to simple from offcial reference
* Make the cache easier to use by commenting out
* fixed clone directory, from stack to yesod
* curl check https
* stack URL
* split build and test, So we can easy detect build or test error
* test job to single thread, So we can easy find bug of test

↓ I change only CI script, So I do not bump version.

After submitting your PR:


ncaq:
relation #1523

ncaq:
The reason why I think this difference is useful is because it makes it easier to understand the cause when similar things happen again.

snoyberg:
Could you rebase this to minimize the diff between the current and new versions? It's not easy to determine from looking at the diff what actual changes are being made here versus rearranging of the file.

ncaq:
Do you mean that you want rebase?
OK, I rebased.
Diff divides CI into build and test and rewrites almost all the lines of the file, so it seems difficult to reduce the difference more than this.

PR #1534, opened 2018-07-03
Reduce deps

snoyberg:


bitemyapp:
screenshot from 2018-07-03 11-12-56
Always a pleasing sight

PR #1537, opened 2018-07-09
AuthEmail: Fix forgot-password endpoint

kk-hainq:
Currently, defining `emailPreviouslyRegisteredResponse` will prevent `sendConfirmationEmail` from happening, even when `registerHelper` is called from `postForgotPasswordR`; which effectively disable the forgot password function.
Since this is just a quick fix for https://github.com/yesodweb/yesod/pull/1524, I don't think we need to bump version?

### DIFF
--- a/yesod-auth/ChangeLog.md
+++ b/yesod-auth/ChangeLog.md
@@ -1,5 +1,8 @@
 # ChangeLog for yesod-auth
 
+## 1.6.4.1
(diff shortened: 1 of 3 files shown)


snoyberg:
Since I already released this to Hackage, we will need a patch version bump.

kk-hainq:
I see, just bumped to `1.6.4.1`.
Should have done this before the register-password one, sorry.

PR #1540, opened 2018-07-13
Fix Improper 406 Responses

StevenXL:

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,9 @@
+## 1.6.7
+
+* If no matches are found, `selectRep` chooses first representation regardless
+    of the presence or absence of a `Content-Type` header in the request
+    [#1540](https://github.com/yesodweb/yesod/pull/1540)
+
 ## 1.6.6
 
 * `defaultErrorHandler` handles text/plain requests [#1522](https://github.com/yesodweb/yesod/pull/1520)
--- a/yesod-core/Yesod/Core/Handler.hs
+++ b/yesod-core/Yesod/Core/Handler.hs
@@ -1289,15 +1289,9 @@ selectRep w = do
         [] ->
             case reps of
                 [] -> sendResponseStatus H.status500 ("No reps provided to selectRep" :: Text)
-                rep:_ ->
-                  if null cts
-                    then returnRep rep
-                    else sendResponseStatus H.status406 explainUnaccepted
+                rep:_ -> returnRep rep
         rep:_ -> returnRep rep
   where
-    explainUnaccepted :: Text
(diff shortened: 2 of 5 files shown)


StevenXL:
Folks,

I wanted to present this PR as a discussion on fixing the issue with the `defaultErrorHandler` function identified by #1402.

There is a lot to clean up in the code. For example, if we keep `selectRepE`, then `selectRep` can be implemented in terms of `selectRepE`, and I'd also style the code so that it is more in keeping with conventions - specifically, no code going all the way to column 158.

However, is the general solution to the problem adequate? The original problem that the tests surface is that we have an error handler (`defaultErrorHandler`) that can itself fail, because `selectRep` can fail. The way I see it, the problem is that the error handling behavior of`selectRep` is baked in. The new function `selectRepE` is `selectRep`, but with explicit failure so that failure by `selectRep` can be handled by the client. 

If this gets a thumbs up I can clean up this code and, hopefully, get all the other tests to pass.

snoyberg:
Playing devil's advocate: what about instead including a fall-back representation for when there is no matching content type? This could be handled exclusively in the `errorHandler` method without other changes.

StevenXL:
@snoyberg Please take a look at the current solution.

It seems to me that the problem is that `selectRep` **does not** [select the first provided representation if no representations match](https://github.com/yesodweb/yesod/blob/ea182bb46484b620f2421211c933ddbcc5123aa1/yesod-core/Yesod/Core/Handler.hs#L1274). That's a necessary but insufficient condition - the `Content-Type` header of the request must also be empty. 

It seems to me that this behavior was added in [this commit](https://github.com/yesodweb/yesod/commit/18be9bc206174e5c2c5e6d202b56dc86beacf93e) very deliberately. 

The reason I felt comfortable changing that work is:

1. The [relevant RFC](https://tools.ietf.org/html/rfc7231#section-6.5.6) seems to allow the server to choose a default representation, so we're not violating the standard as I interpret it.
2. Based on your first comment - paraphrasing as "add a default representation" - it seems to me that you also thought that the `selectRep` function would choose the first representation. 

Tagging @gregwebs to get his thoughts on this as well. 

Happy to take directions on this from the team, but this seemed like a correct, clean change.

snoyberg:
I'm OK with heading in this direction. It seems like perhaps we should have a separate function like `selectRepNoDefault` (bad name) that provides the current behavior, but personally I've never had that need. And it seems like a client can always simply check the `content-type` header and decide it doesn't like what's been delivered.

But a change like this certainly needs to be called out clearly in the ChangeLog.

StevenXL:
@snoyberg take a look now please. Feedback welcomed.

Issue #1541, opened 2018-07-19
selectField selected option is lost when the field validation fails


nmk:
When using `selectField` with either `optionsPairs` or `optionsEnum` and a `check`-validation, if the validation for the field fails, the selected option is lost when the form is reloaded (no option has the `selected` attribute). This is a small demonstration:

https://gist.github.com/ffe31e6a0af7784112f4532e31d9ab99

When `2` is selected from the first dropdown or `Banana` is selected from the second dropdown and the form is submitted both selects have no option selected in the rerendered form. I would expect the same behaviour as exhibited by the number field in the form - if one enters 2 there, the field validation also fails but the 2 is retained in the rerendered form.

Is this expected behaviour and I am just setting up the fiels wrong?

snoyberg:
Firstly, wonderful repro. 💯 

This may be an inherent difference in how selection fields handle their processing versus text-like fields. I'd be open to some changes to make this work, but unfortunately I'm not going to be able to dive in on this right now. I'd be happy to advise if someone else (yourself included!) want to look into this further.

nmk:
I read through the code of `selectFieldHelper` and found [this](https://github.com/nmk/yesod/commit/470bdcf613c1ae23bfe6fd4d1d4bbaffdac57c65) to resolve the problem. The yesod-form tests (`stack test yesod-form`) pass, but I am not sure if there are other implications.

If it looks good I would be happy to add a note to the changelog and submit a PR.

snoyberg:
I think that should work, PR will be greatly appreciated :)

nmk:
I have added a PR: https://github.com/yesodweb/yesod/pull/1544

jezen:
This appears to have been resolved by the merged pull request.

PR #1544, opened 2018-07-26
Do not lose selected value in `selectFieldHelper` when validation fails

nmk:
This PR addresses https://github.com/yesodweb/yesod/issues/1541. I have not bumped the version number, but instead added an entry to the changelog under unreleased. Please advise if anything else needs to be done.

### DIFF
--- a/yesod-form/ChangeLog.md
+++ b/yesod-form/ChangeLog.md
@@ -1,5 +1,10 @@
 # ChangeLog for yesod-form
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


snoyberg:
I _think_ this behavior was introduced in https://github.com/nmk/yesod/commit/1421bacd49d1ee09c7cd3069da2f1a6ff97540b3 from doing a `git blame`. I don't see a problem with the change here.

Issue #1543, opened 2018-07-22
Setting X-XSS-Protection: 1 in response header by default?


pythonissam:
I suggest setting [X-XSS-Protection: 1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection) in HTTP response header by default. With this setting, Yesod can force browsers to enable XSS filtering. I think it allows us to enforce Yesod's security.

Putting `addHeader` into `defaultYesodMiddleware` or modifying stack-template would work.

MaxGabriel:
My understanding is that `1` is the default for all browsers that support it. Though there's a case to be made for using `1; mode=block` instead 

https://blog.innerht.ml/the-misunderstood-x-xss-protection/
https://markitzeroday.com/headers/content-security-policy/2018/02/10/x-xss-protection.html

pythonissam:
Ah, you're right! Yes, `1; mode=block` is what i meant. haha

StevenXL:
Addressed by #1550.

PR #1550, opened 2018-08-03
Set X-XSS-Protection to 1; mode=block.

StevenXL:

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.6.8
+
+* Sets the `X-XSS-Protection` header to `1; mode=block` [#1550](https://github.com/yesodweb/yesod/pull/1550)
+
 ## 1.6.7
 
 * If no matches are found, `selectRep` chooses first representation regardless
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -337,12 +337,14 @@ defaultShouldLogIO :: LogSource -> LogLevel -> IO Bool
 defaultShouldLogIO _ level = return $ level >= LevelInfo
 
 -- | Default implementation of 'yesodMiddleware'. Adds the response header
--- \"Vary: Accept, Accept-Language\" and performs authorization checks.
+-- \"Vary: Accept, Accept-Language\", \"X-XSS-Protection: 1; mode=block\", and
+-- performs authorization checks.
 --
 -- Since 1.2.0
 defaultYesodMiddleware :: Yesod site => HandlerFor site res -> HandlerFor site res
 defaultYesodMiddleware handler = do
     addHeader "Vary" "Accept, Accept-Language"
+    addHeader "X-XSS-Protection" "1; mode=block"
     authorizationCheck
     handler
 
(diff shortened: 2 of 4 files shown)


StevenXL:
Address #1543. This is a very small PR, so if this is deemed unnecessary, happy to discard.

snoyberg:
Pinging @maxgabriel. I haven't been following this header in general.

@stevenxl if you feel comfortable with this, you should also feel free to merge.

MaxGabriel:
I think that this is unlikely to cause anyone any problems, and improves the situation over the browser default by making the error more visible. I'd say it's good to merge.

StevenXL:
@MaxGabriel thanks.

MaxGabriel:
As a side note, there are other security headers that people should maybe add: 
https://securityheaders.com/?q=alpha.mercury.co&followRedirects=on

They require more custom configuration compared to this one, so there isn't much a library would really be adding, but it would be cool if Yesod developers were encouraged to use them. Not sure the best way to do that—maybe some commented out code in the scaffolding?

bitemyapp:
>Not sure the best way to do that—maybe some commented out code in the scaffolding?

Be a good start but I'd rather good defaults that are on, if possible.

screenshot from 2018-08-06 16-04-45

Here's what was missing on my relatively vanilla Yesod app that is otherwise A+ on qualys ssl labs.

MaxGabriel:
Content-Security-Policy is a huge win for security, because you can whitelist where things like Javascript can be loaded from, and prevent XSS from harming you. This is our CSP used for a single-page app (with some additions coming from headers):

```
  <meta http-equiv="Content-Security-Policy" content="
    base-uri 'self';
    block-all-mixed-content;
    default-src 'self';
    connect-src
      'self'
      ws://localhost:3001
      ws://app.mercury.local
      ws://lh.mercury.co
      wss://app.mercury.local
      https://maps.googleapis.com
      http://localhost:3000
      https://api.mercury.local
      https://app.mercury.local
      https://s3.amazonaws.com/mercury-technologies-user-uploads-dev
      https://api.mixpanel.com
      https://sessions.bugsnag.com
      https://notify.bugsnag.com
      https://backend.mercury.co;
    font-src
      'self'
      https://fonts.gstatic.com;
    img-src
      https://csi.gstatic.com
      https://maps.googleapis.com
      'self';
    media-src
      'none';
    script-src
      'self'
      'sha256-srzUm2bR52Dg2g2SqWiGuo719+FqQulMWAYRYJ8hkhc='
      https://maps.googleapis.com
      https://cdn.plaid.com/link/v2/stable/link-initialize.js;
    style-src
      'self'
      https://hello.myfonts.net/count/3694b0
      https://fonts.googleapis.com
      https://maps.googleapis.com
      'unsafe-inline';
    frame-src
      https://app.mercury.local
      https://cdn.plaid.com/;
    object-src
      'none';
    form-action
      'none';
    worker-src
      'self';
  " />
```

I think the best default you could give is to set the `default-src` to self, to only allow the domain serving the Yesod app to send any javascript, css, etc. The UX downside is that developers wouldn't be able to load e.g. JQuery from cdnjs until they went to the CSP and whitelisted it. The error messages in the browser are pretty informative about this, but it would definitely trip a lot of people up. Github has a write-up of their use of it: 
https://githubengineering.com/githubs-csp-journey/ Thoughts?

Referrer-Policy seems fine to enable by default. My understanding is that it improves privacy by concealing where users came from.

Feature-Policy is brand spanking new, like it was on Hacker News last week, so I don't have comments on it really. But it lets you do things like prevent the site from vibrating the phone 
https://wicg.github.io/feature-policy/

Issue #1221, opened 2016-04-21
Yesod.Auth.Email.verifyR leaks verification key ids.


chreekat:
E.g. http://localhost:3000/auth/page/email/verify/3/f22uWzaa7s2aqxV8Id1Hz3cs

Here, `/3/` is the verification key id.

I would expect the verification url to _only_ use the key itself.

I note that `verifyR` already has a `FIXME` comment. :) Maybe this is what it refers to?

I guess fixing this will require adding a new method to the YesodAuthEmail typeclass, in order to work backwards from just the key itself.

snoyberg:
I'm ambivalent about this one. Leaking the key ID is not a major issue IMO.
That said, if someone wants to take a stab at this, I'll happily review.

chreekat:
Actually, it might be particularly harmless once keys are deleted in #1222.
I'll think about how much i really care about this issue...
On Apr 21, 2016 11:15 PM, "Michael Snoyman" <notifications@github.com>
wrote:

I'm ambivalent about this one. Leaking the key ID is not a major issue IMO.
That said, if someone wants to take a stab at this, I'll happily review.

On Thu, Apr 21, 2016 at 11:25 PM Bryan Richter <notifications@github.com>
wrote:

> E.g.
> http://localhost:3000/auth/page/email/verify/3/f22uWzaa7s2aqxV8Id1Hz3cs
>
> Here, /3/ is the verification key id.
>
> I would expect the verification url to *only* use the key itself.
>
> I note that verifyR already has a FIXME comment. :) Maybe this is what it
> refers to?
>
> I guess fixing this will require adding a new method to the YesodAuthEmail
> typeclass, in order to work backwards from just the key itself.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/yesodweb/yesod/issues/1221>
>

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/yesodweb/yesod/issues/1221#issuecomment-213277826>

chreekat:
I still think it's a good idea to fix this some day, but I don't think it's high enough priority for me to do it now.

I can't tell how/if these issues are organized, so I'm not sure if I should keep this open, or how you'd like to deal with it. I'm fine with closing it though.

felixSchl:
The way it could be done is encrypting the verification id in the verification key - is that a reasonable approach?

StevenXL:
[...]

PR #1557, opened 2018-10-06
Do not expose AuthEmailId in verify URL.

StevenXL:

After submitting your PR:

### DIFF
--- a/yesod-auth/ChangeLog.md
+++ b/yesod-auth/ChangeLog.md
@@ -1,5 +1,8 @@
 # ChangeLog for yesod-auth
 
+## 1.6.5
+* Do not expose AuthEmailId in verify URL [#1557](https://github.com/yesodweb/yesod/pull/1557)
+
 ## 1.6.4.1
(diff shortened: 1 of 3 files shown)


StevenXL:
This hopefully addresses #1221.

StevenXL:
@snoyberg You are correct - the last version I uploaded wouldn't work. I have fixed the dispatch function so that it no longer expects the `eid`. I try to keep the changes necessary to a minimum, so that instead of changing the signature of multiple methods in the `YesodAuthEmail` typeclass to something that utilizes `VerKey`, I added the method `getAuthEmailId :: VerKey -> AuthHandler site (Maybe (AuthEmailId site))`. 

Once I get an OK on the changes to the source code, I will update `auth2.hs` so the example stays up to do date with the actual code. However, I'd like to do this as its dependent on locking down the changes to `Yesod.Auth.Email`.

StevenXL:
I agree with your points above @snoyberg. I'm going to close this PR, but I'd also like to close #1221. Thoughts on that?

PR #1560, opened 2018-10-11
Relax constraint, allow persistent-2.9.*

DanBurton:
In anticipation of a new release of `persistent`. See: https://github.com/yesodweb/persistent/blob/04d1c5169e6174fcfd8836c64739168c9e975d68/persistent/persistent.cabal#L2

```
```

I have not bumped the version number, as this change can be published as a revision, and the actual "code change" can be uploaded later.


DanBurton:
Just noticed that #1516 is already open for this purpose.

PR #1571, opened 2019-01-18
Make FormResult instances of Eq and Monad

jlamothe:
I don't know why this wasn't already the case. Perhaps because it doesn't accumulate errors, but this doesn't really make sense in a monadic context.

Feedback is welcome.


psibi:
Related: https://github.com/yesodweb/yesod/pull/1559

jlamothe:
Removed the Monad part.  Eq still applies.

psibi:
Thanks, Can you update changelog too: https://github.com/yesodweb/yesod/blob/master/yesod-form/ChangeLog.md ?

Issue #1575, opened 2019-01-22
requireJsonBody is vulnerable to XSS, is the wrong default and should be deprecated


eborden:
`requireJsonBody` is open to XSS attacks through `text` and `form` mime types and can lead to privilege escalation. This is the wrong default for `yesod` to have and leads users into the lion's den. This function should be deprecated in favor of the safer alternative, `requireCheckJsonBody`. The insecure version can still be supported as `insecureRequireJsonBody` for users that need it.

snoyberg:
I'm not opposed to a deprecation, the original functionality was clearly a
mistake on my part. However, I'm not seeing the attack vector you mention,
could you provide some more details?

MaxGabriel:
There is a potential attack vector in that the browser [does not trigger CORS requests for "simple requests"](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests), which includes POST requests that are form or text content-types. An attacker can craft [a form whose body is valid JSON](https://docs.spring.io/spring-security/site/docs/4.2.x/reference/html/csrf.html#csrf-protection-and-json), and when a user visits attacker.com and submits that form, it can be submitted to bank.com and bypass CORS.

Checking the content-type is `application/json` prevents this, because if the content-type was set to `application/json`, then the browser would send a CORS request—a preflight OPTIONS request to the server asking if the current domain (and some other values) are whitelisted to send requests to that server. If the server doesn't say attacker.com is whitelisted, the browser will not send the real request to the server.

I would describe this as a CSRF attack, not XSS, though.

It's worth noting that this attack vector does not apply to an API service not intended to serve browsers—if your customers are hitting an API from say curl or a Ruby script, those clients don't do CORS checks (not being in a browser, they don't have origin domains). This might impact how the API should be named/what a potential deprecation message should be.

snoyberg:
Thanks for clarifying! Would one of you be able to PR a fix?

MaxGabriel:
I'll let @eborden take it if he would like, otherwise I'll do it

eborden:
Thanks for jumping in with the details @MaxGabriel. I'll put a PR together today.

PR #1576, opened 2019-01-23
Deprecate insecure JSON body functions

eborden:
`parseJsonBody` and `requireJsonBody` do not require a mime type when
parsing `JSON` content. This leaves them open to CSRF. They are now
deprecated and `insecure` versions are added in their place. Consumers
are now given a proper choice between secure and insecure functions.

There is a potential attack vector in that the browser does not trigger
CORS requests for "simple requests", which includes POST requests that
are form or text content-types. An attacker can craft a form whose body
is valid JSON, and when a user visits attacker.com and submits that
form, it can be submitted to bank.com and bypass CORS.

Checking the content-type is application/json prevents this, because if
the content-type was set to application/json, then the browser would
send a CORS request—a preflight OPTIONS request to the server asking if
the current domain (and some other values) are whitelisted to send
requests to that server. If the server doesn't say attacker.com is
whitelisted, the browser will not send the real request to the server.

After submitting your PR:

### DIFF
--- a/yesod-core/Yesod/Core/Json.hs
+++ b/yesod-core/Yesod/Core/Json.hs
@@ -10,11 +10,14 @@ module Yesod.Core.Json
     , provideJson
 
       -- * Convert to a JSON value
-    , parseJsonBody
     , parseCheckJsonBody
+    , parseInsecureJsonBody
+    , requireCheckJsonBody
+    , requireInsecureJsonBody
+      -- ** Deprecated JSON conversion
+    , parseJsonBody
     , parseJsonBody_
     , requireJsonBody
-    , requireCheckJsonBody
 
       -- * Produce JSON values
     , J.Value (..)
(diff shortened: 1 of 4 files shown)


eborden:
This addresses https://github.com/yesodweb/yesod/issues/1575

@MaxGabriel, I'm not sure what `yesod`'s versioning policy is. I would consider a deprecation to be a major bump.

yesod-core/Yesod/Core/Json.hs
```diff
@@ -31,26 +35,27 @@ module Yesod.Core.Json
...
+import Data.Text (pack)
+import qualified Data.Vector as V
+import Yesod.Core.Class.Handler
+import Yesod.Core.Class.Yesod (Yesod, defaultLayout)
+import Yesod.Core.Content (TypedContent)
+import Yesod.Core.Handler
+  (HandlerFor, ProvidedRep, getRequest, invalidArgs, lookupHeader, provideRep, rawRequestBody, redirect, selectRep)
+import Yesod.Core.Types (reqAccept)
+import Yesod.Core.Widget (WidgetFor)
+import Yesod.Routes.Class
```
eborden: Woops, looks like my `stylish-haskell` got in here.

snoyberg: Would you be able to rebase without these unrelated changes?

eborden: Yup, I'll pull them out.

yesod-core/Yesod/Core/Json.hs
```diff
@@ -92,51 +95,70 @@ returnJsonEncoding = return . J.toEncoding
...
+--
+-- @since 0.3.0
+parseJsonBody :: (MonadHandler m, J.FromJSON a) => m (J.Result a)
+parseJsonBody = parseInsecureJsonBody
+{-# DEPRECATED parseJsonBody "Use parseCheckJsonBody or parseInsecureJsonBody instead" #-}
+
+-- | Same as 'parseCheckJsonBody', but does not check that the mime type
+-- indicates JSON content.
+--
+-- Note: This function is vulnerable to CSRF attacks.
```
snoyberg: Can you add `@since` comments, a changelog entry, and a minor version bump to 1.6.11?

eborden: Can do!

Issue #1578, opened 2019-01-26
Add method of changing where in the output the julius bundle is placed.


Fizzixnerd:
Right now when you render a widget, the julius bundle is tacked onto the end of the scripts.  I would like to be able to choose where exactly it goes, relative to other scripts added via `addScript` and `addScriptRemote`.  The reason I want to do this is so that I can pass information from the julius bundle to later scripts on the page.  For example, I might have something like:

small_file.julius
```javascript
var config_options = {
  myYesodVariable = #{myYesodVariable}
};
```

big_bundle.js
```javascript
// pretend this is an enormous 
// React bundle that slows compilation
// if used directly as a templated file
// ...
var myObject = new MyObject(config_options.myYesodVariable);

// ...
```

As it stands, I can find no way (besides dynamically loading the big_bundle.js script after the page has loaded) to make sure the julius file's contents are available to big_bundle.js.  I would like to be able to do something like:
```haskell
myWidget = do
  addScriptRemote "https://www.example.com/my_remote_script_perhaps_used_in_a_julius_file.min.js"
  anchorJulius
  addScript $ StaticR js_big_bundle_js
```
And have the resulting page put the autogen bundle between the two scripts.

Fizzixnerd:
I implemented the change to do this in my spare time.  I can submit a pull request if you'd like.

snoyberg:
When I've needed something like this in the past, I've embedded the Javascript inside a Hamlet `<script>` tag. If there's something more elegant, I'm open to a PR.

PR #1580, opened 2019-01-30
Add `addScriptAnchor` to control where in the page the autogen bundle is placed (#1578).

Fizzixnerd:
Addresses #1578.

When creating a `Widget`, `addScriptAnchor` can be used to anchor the autogen bundle to a particular location in the page.  If omitted, the default behavior is to put the script at the end of the bundle as before.

Usage:

```haskell
myWidget = do
  addScript $ StaticR js_before_bundle_js
  addScriptAnchor
  addScript $ StaticR js_after_bundle_js
```

After submitting your PR:

### DIFF
--- a/yesod-core/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/Yesod/Core/Class/Yesod.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes       #-}
 {-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE LambdaCase #-}
 module Yesod.Core.Class.Yesod where
 
 import           Yesod.Core.Content
@@ -53,6 +54,7 @@ import           Yesod.Core.Widget
 import Data.CaseInsensitive (CI)
 import qualified Network.Wai.Request
 import Data.IORef
+import Data.Maybe
 
 -- | Define settings for a Yesod applications. All methods have intelligent
 -- defaults, and therefore no implementation is required.
@@ -515,7 +517,8 @@ widgetToPageContent w = HandlerFor $ \hd -> do
     }
   GWData (Body body) (Last mTitle) scripts' stylesheets' style jscript (Head head') <- readIORef ref
   let title = maybe mempty unTitle mTitle
-      scripts = runUniqueList scripts'
+      toUnique = UniqueList . (:)
+      scripts = runUniqueList (scripts' <> toUnique SOAAnchor)
       stylesheets = runUniqueList stylesheets'
 
   flip unHandlerFor hd $ do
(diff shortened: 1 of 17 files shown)


Fizzixnerd:
So the CI failed on one of the Linux GHCs.  It seems like it couldn't find fpcomplete.com?

yesod-core/Yesod/Core/Class/Yesod.hs
```diff
@@ -541,18 +546,24 @@ widgetToPageContent w = HandlerFor $ \hd -> do
                    $ encodeUtf8 $ renderJavascriptUrl render s
                 return $ renderLoc x
 
-    -- modernizr should be at the end of the <head> http://www.modernizr.com/docs/#installing
-    -- the asynchronous loader means your page doesn't have to wait for all the js to load
-    let (mcomplete, asyncScripts) = asyncHelper render scripts jscript jsLoc
+    let scripts'' = catMaybes $ (\case
```
snoyberg: This looks like a usage of `mapMaybe`. It's probably even more elegant as a list comprehension `[s | SOAScript s <- scripts]`.

Fizzixnerd:
I have added tests to the suite to demonstrate it working.

The mechanism is basically this: in the code below in `regularScriptLoad`, it used to make a script tag for each script, and then at the end put the autogen bundle.  Now it places the bundle at the location of any SOAAnchors it finds.
```haskell
    let scripts'' = [s | SOAScript s <- scripts]
        -- modernizr should be at the end of the <head> http://www.modernizr.com/docs/#installing
        -- the asynchronous loader means your page doesn't have to wait for all the js to load
        (mcomplete, asyncScripts) = asyncHelper render scripts'' jscript jsLoc
        regularScriptLoad = [hamlet|
            $newline never
            $forall s <- scripts
                $case s
                    $of SOAScript s
                        ^{mkScriptTag s}
                    $of SOAAnchor
                        $maybe j <- jscript
                            $maybe s <- jsLoc
                                <script src="#{s}" *{jsAttributes master}>
                            $nothing
                                <script>^{jelper j}
        |]
```

The reason this is okay is because just above we have:
```haskell
      scripts = runUniqueList (scripts' <> toUnique SOAAnchor)
```
which makes sure that:

1. There is at most one `SOAAnchor` so the bundle is not included multiple times.
2. There is at least one `SOAAnchor` at the end of the bundle, in case `addScriptAnchor` was never used.  This preserves the old behavior for old code.

The only thing I'm just now wondering about is if there would be any difference for async scripts.

snoyberg:
I may be dense here, but is there a reason not to go in the direction I mentioned, of generating `script` tags and adding them as HTML to the body?

Fizzixnerd:
One issue is that there is no instance of `Blaze.ToMarkup` for JSON `Value`s.  For instance, if I wanted do something like:
```haskell
data MyObj = MyObj
  { s :: String,
    n :: Int
  } deriving (Eq, Show)

instance ToJSON MyObj where
  toJSON MyObj {..} = object
    [ "s" .= s
    , "n" .= n
    ]

embeddedScript :: Handler Html
embeddedScript = defaultLayout $ do
  let myObj = MyObj "string and stuff" 4
  [whamlet|
          <script>
            myObj = #{toJSON myObj};
          |]
```
then I get a type error, while

```haskell
normalScript :: Handler Html
normalScript = defaultLayout $ do
  let myObj = MyObj "string and stuff" 4
  toWidget [julius|myObj = #{toJSON myObj};|]
```
works just fine.  It is, of course, possible to serialize this value manually, but more complicated examples would be more annoying to deal with (such as an array, which is something I actually have to pass in my application).

jezen:
Closing due to inactivity.

PR #1581, opened 2019-02-10
Use at most one valid session cookie per request

### LINKED ISSUE #994: Possible minor security issue with client session handling
I was reviewing the client session code and [this caught my eye](https://github.com/yesodweb/yesod/blob/882956255a2f5e6141bb34c50461d743843c42ee/yesod-core/Yesod/Core/Class/Yesod.hs#L698):

``` haskell
    sess date = Map.unions $ do
      raw <- [v | (k, v) <- W.requestHeaders req, k == "Cookie"]
      val <- [v | (k, v) <- parseCookies raw, k == sessionName]
      let host = "" -- fixme, properly lock sessions to client address
      maybe [] return $ decodeClientSession key date host val
```

If I'm understanding the code correctly, we're:
- Taking all cookie headers.
- Taking all cookies with the target name (default `_SESSION`).
- Decoding them all separately.
- Merging the contents of all of them.

As the code has almost no comments, it's not clear if the merging is intended behavior or just a convenient way that works for the common cases (namely, zero and one).  However, it's certainly unexpected for me to find this behavior, and I have already looked at the snippet before.  

The `clientsession` library guarantees that the session data was constructed by the server at a point in the past, and that the server may trust it.  The behavior above allows one to take unions of any two or more valid sessions.

On most applications this should be harmless as the last session is going to override most of the earlier ones.  However, if some application in the wild maintained some invariant between session variables, or conditionally added some session variables, they may not be prepared to handle these unions.

As I cannot imagine how someone would be using this feature, I assume it's a misfeature.  I propose that we force at most one session cookie to be present.  In other words, if more than one session cookie is found, ignore them all.

nytopop:
Makes `loadClientSession` ignore all sessions in a request if more than a single session cookie decodes successfully. The prior behavior was to merge all valid session cookies' values.

Bumps version to 1.6.11.1.

Not sure if this counts as a bugfix or breaking change that would require 1.7.

Closes #994, I figured after 3 years it's probably fair game :slightly_smiling_face:. Feel free to close if you don't think this change is warranted anymore.

After submitting your PR:

### DIFF
--- a/yesod-core/yesod-core.cabal
+++ b/yesod-core/yesod-core.cabal
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.11
+version:         1.6.12
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-core
 
+## 1.6.12
+
+* Use at most one valid session cookie per request [#1581](https://github.com/yesodweb/yesod/pull/1581)
+
 ## 1.6.11
 
 * Deprecate insecure JSON parsing functions [#1576](https://github.com/yesodweb/yesod/pull/1576)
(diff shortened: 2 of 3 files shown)


yesod-core/yesod-core.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.11
+version:         1.6.11.1
```
snoyberg: This should be a minor version bump, to 1.6.12, since it does affect behavior.

nytopop: Fixed :+1: and rebased to fix the commit message.

snoyberg:
Thanks!

PR #1586, opened 2019-03-06
Document runDB

MaxGabriel:
My coworker who is new to Haskell was pointing out that for such an important function to Yesod, this one is lacking any documentation. It's slightly hard to document because people could provide various implementations for it, but I think this description captures the essence pretty well, and notes the important implicit behavior of opening a transaction.

image

(Ignoring the following since this is a documentation only PR)

After submitting your PR:


bitemyapp:
This is a good idea. It's an important faculty.

Issue #1588, opened 2019-03-13
maximumContentLength can do IO


geekingfrog:
The current type signature for `maximumContentLength` is `site -> Maybe (Route site) -> Maybe Word64`.
I'd like to have the ability to restrict the content length based on a path parameter of the route, so I would need a db lookup.
For example: `/upload/foo` -> 2MB max, but `/upload/bar` -> 10MB max, these mappings are stored in a db.
Is there any workaround I can use for now? Is it something which could be added to the next release?

snoyberg:
We had a similar situation with `shouldLog` in the past, and added a `shouldLogIO` method to the `Yesod` type class to address this. I think we could do the same thing in this case. Would you be interested in taking a stab at such a PR?

For comparison, here's the prior art: https://github.com/yesodweb/yesod/commit/aa5781d4e42f8797ebe2974a6f3cfabe43af4c50

PR #1589, opened 2019-03-15
maximumContentLengthIO

geekingfrog:
As raised with #1588 here's an attempt to add this feature. I mostly replicated what was done for `shouldLogIO`.

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/src/Yesod/Core/Class/Yesod.hs
@@ -195,6 +195,7 @@ class RenderRoute site => Yesod site where
     addStaticContent _ _ _ = return Nothing
 
     -- | Maximum allowed length of the request body, in bytes.
+    -- This method may be ignored if 'maximumContentLengthIO' is overridden.
     --
     -- If @Nothing@, no maximum is applied.
     --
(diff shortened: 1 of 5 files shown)


yesod-core/src/Yesod/Core/Class/Yesod.hs
```diff
@@ -202,6 +202,18 @@ class RenderRoute site => Yesod site where
     maximumContentLength :: site -> Maybe (Route site) -> Maybe Word64
     maximumContentLength _ _ = Just $ 2 * 1024 * 1024 -- 2 megabytes
 
+    -- | Maximum allowed length of the request body, in bytes. This is similar
+    -- to 'maximumContentLength', but the result lives in @IO@. This allows
+    -- you to dynamically change the maximum file size based on some external
+    -- source like a database or an @IORef@.
+    --
+    -- The default implementation uses 'maximumContentLength'. Future version of yesod will
```
snoyberg: How about adding a comment to `maximumContentLength` as well that overriding this method will cause `maximumContentLength` to be ignored?

geekingfrog: I've pushed f1374c91409f2e22a0b1cc8890e87fbbd8d7d880 but I'm not sure this is what you wanted.

PR #1592, opened 2019-04-12
Introduce JSONResponse.

StevenXL:
This data type allows us to return a domain type in our handlers, even
if we eventually want to send JSON to the client.

See: https://tech.freckle.com/2015/12/21/servant-style-handlers-for-yesod/

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Types.hs
+++ b/yesod-core/src/Yesod/Core/Types.hs
@@ -8,8 +8,10 @@
 {-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE GADTs #-}
 module Yesod.Core.Types where
 
+import Data.Aeson (ToJSON)
 import qualified Data.ByteString.Builder            as BB
 import           Control.Arrow                      (first)
 import           Control.Exception                  (Exception)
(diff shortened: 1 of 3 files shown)


StevenXL:
Tagging #1481  and #1494

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -303,6 +305,20 @@ newtype RepXml = RepXml Content
...
+-- Example usage in a type signature:
+--
+-- > postSignupR :: Handler (JSONResponse CreateUserResponse)
+--
+-- And in the implementation:
+--
+-- > return $ JSONResponse $ CreateUserResponse userId
+--
+-- @since 1.6.14
+data JSONResponse a where
```
snoyberg: It's an interesting design, but I would have expected a simpler `newtype` and then putting the `ToJSON` constraints on the instances instead. That removes any runtime overhead (however negligible) from an extra data constructor, but more importantly sticks to simpler constructs. Is there a strong motivation for the GADT approach?

StevenXL: Hi Michael - adding the constraints onto the instance definitions is possible, but @MaxGabriel mentioned in #1494 that the errors using the newtype approach are a bit cryptic, and to answer your question from that PR, GADTs does make the errors simpler. 

At the end of the day though, the error between the newtype approach and GADTs approach are both completely understandable. I'm happy to implement it with newtype; let me know which way you'd like to proceed, as I do think it is a nice feature to add.

snoyberg: Better error messages sounds like a great reason. I rescind my comment, let's go this way. Thanks for clarifying!

snoyberg:
Do you want to merge and release, or would you prefer I do?

StevenXL:
> Do you want to merge and release, or would you prefer I do?

If you could do so that would be great; I'll close the old PR and Issue.

snoyberg:
Cool, doing so now, thanks!

StevenXL:
> Cool, doing so now, thanks!

Always a pleasure!

PR #1593, opened 2019-04-19
I18n placeholder

StevenXL:

After submitting your PR:

### DIFF
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -53,6 +53,7 @@ module Yesod.Form.Functions
     , convertField
     , addClass
     , removeClass
+    , fullAttrs
     ) where
 
 import Yesod.Form.Types
@@ -618,6 +619,23 @@ convertField to from (Field fParse fView fEnctype) = let
   fParse' ts = fmap (fmap (fmap to)) . fParse ts
   fView' ti tn at ei = fView ti tn at (fmap from ei)
   in Field fParse' fView' fEnctype
+
+-- | Combine placeholder and fsAttrs into a full set of attributes.
+--
+-- @since 1.6.5
+fullAttrs :: (site ~ HandlerSite m, MonadHandler m) => FieldSettings site -> m [(Text, Text)]
+fullAttrs fs = do
+    site <- getYesod
+    langs <- languages
+    let renderer = renderMessage site langs
+    return $ combineAttrs renderer (fsAttrs fs) (fsPlaceholder fs)
+
+combineAttrs :: (message -> Text) -- ^ Function to internationalize a message
+             -> [(Text, Text)] -- ^ List of existing 'fsAttrs'
+             -> Maybe message  -- ^ An optional placeholder message
+             -> [(Text, Text)] -- ^ Full list of attributes
+combineAttrs _ attrs Nothing  = attrs
+combineAttrs f attrs (Just t) = attrs ++ [("placeholder", f t)]
 
 -- | Removes a CSS class from the 'fsAttrs' in a 'FieldSettings'.
 --
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -569,7 +569,7 @@ customErrorMessage msg field = field
 
 -- | Generate a 'FieldSettings' from the given label.
 fieldSettingsLabel :: RenderMessage site msg => msg -> FieldSettings site
-fieldSettingsLabel msg = FieldSettings (SomeMessage msg) Nothing Nothing Nothing []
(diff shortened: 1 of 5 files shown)


StevenXL:
Hi. I would like some feedback on this PR (hence why the PR template is blank). 

This PR is meant to address #686. The original discussion centered around creating a new field `fsPlaceholder` on the `FieldSettings` data type. I still believe this is the right way to go, but the downside is that anyone that is dealing directly with `fsAttrs` now has to know that they need to combine that with the `fsPlaceholder`. There's a helper function to do so `fullAttrs` and `combineAttrs`, but again, you have to know to use that.

I didn't think it made sense to add `fvPlaceholder` to the `FieldView` data type, since by this point the placeholder will be "baked in" to the input. (Even if it does make sense to do so, it might be better to leave that to a second PR). 

I believe this is a breaking change for other packages, and if the PR makes it to merge status, we should bump the version number to 1.7. Thoughts?

yesod-form/Yesod/Form/Functions.hs
```diff
@@ -619,6 +621,23 @@ convertField to from (Field fParse fView fEnctype) = let
   fView' ti tn at ei = fView ti tn at (fmap from ei)
   in Field fParse' fView' fEnctype
 
+-- | Combine placeholder and fsAttrs into a full set of attributes.
+--
+-- @since 1.6.5
```
StevenXL: Not to self. Update this once we fill out PR template.

yesod-form/Yesod/Form/Functions.hs
```diff
@@ -619,6 +621,23 @@ convertField to from (Field fParse fView fEnctype) = let
...
+fullAttrs :: (site ~ HandlerSite m, MonadHandler m) => FieldSettings site -> m [(Text, Text)]
+fullAttrs fs = do
+    site <- getYesod
+    langs <- languages
+    let renderer = renderMessage site langs
```
StevenXL: Can replace this with `getMessageRender`.

snoyberg:
As an alternative: what about writing a function that renders the message and inserts the placeholder value into the existing `Attrs` field? What's the advantage of deferring the rendering until later in the process?

StevenXL:
Hi Michael - thank you for the reply.

The problem with that alternative is that the only way to actually perform the rendering is within a `MonadHandler` monad, since we need to know both the master site and the languages that the user will accept. A function which does the rendering and adds the internationalized placeholder to the `fsAttrs` field would have to return `MonadHandler m -> m (FieldSettings site)`, or it would have to accept a function `msg -> Text` which, again, can only be retrieved within a `MonadHandler`. 

All of this to say that, such a function would not allow us to separate what we wish to internationalize from the actual process of internationalization. To quote the original #686, it would only be a slightly better than using `getMessageRender` manual

The only wart with the current PR that I can see is that if you are processing a `FieldSettings` manually, you have to remember to insert the placeholder in the right place. We can address this by changing `fsAttrs :: [(Text, Text)]` to `fsAttrs :: [(Text, SomeMessage msg)]`, but I think this might be getting too cute by half. 

```
 placeHolder :: (MonadHandler m, RenderMessage (HandlerSite m) msg) => msg -> FieldSettings site -> m (FieldSettings site)
 placeHolder msg fs = do
   renderer <- getMessageRender
   let newAttrs = fsAttrs fs ++ [("placeholder", renderer msg)]
   return (fs {fsAttrs = newAttrs})
 
 placeHolder' :: msg -> (msg -> Text) -> FieldSettings site -> FieldSettings site
 placeHolder' msg render fs = fs {fsAttrs = newAttrs}
   where newAttrs = fsAttrs fs ++ [("placeholder", render msg)]
```

snoyberg:
My bigger concern is that this is a non-scalable solution. If next week, someone needs a special message rendered for a `data-*` attribute, for instance, we'll be back at square one. I don't see any reason to believe things will stop at placeholder.

StevenXL:
> My bigger concern is that this is a non-scalable solution. If next week, someone needs a special message rendered for a `data-*` attribute, for instance, we'll be back at square one. I don't see any reason to believe things will stop at placeholder.

I think this is a fair argument, but if that is the case, then I say the solution would be to change:

```
data FieldSettings master = FieldSettings
    { fsLabel :: SomeMessage master
    , fsTooltip :: Maybe (SomeMessage master)
    , fsId :: Maybe Text
    , fsName :: Maybe Text
    , fsAttrs :: [(Text, Text)]
    }
```

to 

```
data FieldSettings master = FieldSettings
    { fsLabel :: SomeMessage master
    , fsTooltip :: Maybe (SomeMessage master)
    , fsId :: Maybe Text
    , fsName :: Maybe Text
    , fsAttrs :: [(Text, SomeMessage master)]
    }
```

This would allow *any* attribute to be internationalized. I'm going to proceed in that direction, and will open up a new PR, so that we can see where that leads us. 

Thank you for the help!

PR #1597, opened 2019-05-10
Internationalize

StevenXL:

After submitting your PR:

### DIFF
--- a/yesod-form/ChangeLog.md
+++ b/yesod-form/ChangeLog.md
@@ -1,5 +1,10 @@
 # ChangeLog for yesod-form
 
+## 1.6.5
+
+* Internationalize fsAttrs [#1597](https://github.com/yesodweb/yesod/pull/1597)
+* Change the type of the `fsAttrs` field from `[(Text, Text]` to `[(Text, [SomeMessage master])]`
+
 ## 1.6.4
 
 * Make FormResult an instance of Eq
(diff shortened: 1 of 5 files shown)


StevenXL:
Please see #686 and #1593.

The type of `fsAttrs` (in the `FieldSettings` data type) does not reflect the full semantics of HTML attributes. This pull request changes the type of `fsAttrs` from `[(Text, Text)]` to `[(Text, [SomeMessage master])]`. 

The change of the second element of the tuple to a list allows us to express the idea that an attribute can have multiple values. For example: `<input class="rounded orange"/>`. 

The use of `SomeMessage master` allows the user to provide to use some value that we can, later on, internationalize.

StevenXL:
Also, I am wary about `removeClass`, given the definition of [SomeMessage](https://hackage.haskell.org/package/shakespeare-2.0.20/docs/src/Text.Shakespeare.I18N.html#SomeMessage). We can't compare values of type `SomeMessage` for equality, because the type system has lost information about what data type the `SomeMessage` data constructor holds. So `removeClass` does not make sense.

Should we update `SomeMessage master` to:

```
data SomeMessage master msg where
    SomeMessage :: (Eq msg, RenderMessage master msg) => msg -> SomeMessage master msg

fromSomeMessage :: SomeMessage master msg
fromSomeMessage (SomeMessage msg) = msg
```

I'm not sure what can of worms I'd be opening up at that point.

snoyberg:
I'm not really bought in on this. You've identified at least one issue above. I still don't see the major issue with forcing the caller to render to `Text` before including in the `fsAttrs`.

StevenXL:
@snoyberg I agree with you. I think the changes required to make this change are not worth it. I will close down the PRs as well as the issue. We can revisit this problem if the issue comes up again.

PR #1604, opened 2019-06-10
Optional labels

jprider63:

After submitting your PR:

### DIFF
--- a/yesod-form/ChangeLog.md
+++ b/yesod-form/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-form
 
+## 1.6.5
+
+* Omit labels in `renderBootstrap3` when they are null.
+
 ## 1.6.4
 
 * Make FormResult an instance of Eq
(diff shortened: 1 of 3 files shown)


jprider63:
Actually, I've only tested this with `BootstrapBasicForm`. It might not make sense for `BootstrapHorizontalForm`.

snoyberg:
Did you want to investigate that before I review?

jprider63:
Sure. I should be able to later this afternoon.

jprider63:
According to the [documentation](https://getbootstrap.com/docs/3.3/css/#forms-inline), it looks like we should use `.sr-only` to hide labels so I've switched to do this. Inline forms currently always have `.sr-only` so I left it as is.

yesod-form/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-form
 
+## 1.6.5
+
+* Omit labels in `renderBootstrap3` when they are null.
```
snoyberg: Perhaps update this to clarify we use `.sr-only`?

PR #1606, opened 2019-06-25
urlParamRenderOverride includes prerequest session info and resovled approot

sbditto85:
Submitting the PR for RFC not for merge really.

Basically I have a somewhat odd need in that when a URL is rendered if the session contains a specific value then I need put something in the query string so that the value persists even if the link is followed from an email to another computer etc. Since I only want to change the query string I need the resolved app root and since I need access to the session the only way i found to do that is by passing it in, but that means the middleware and handler can't affect the way the url is rendered so a full request must happen to set the session THEN the urls would have the needed session information passed on.

I tried to allow some kind of affects in urlParamRenderOverride but that just caused a huge rabbit hole of changing types everywhere which seemed like a bad option.

Is there a better way to do this? I'm not really a fan of this particular change but its the closest I've come to making what I need done. The only alternative I can think of is to change our architecture a bit and use sub domains to represent the data so that the resolved app root will "contain" it.

Thanks for the help! 

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Class/Yesod.hs
+++ b/yesod-core/src/Yesod/Core/Class/Yesod.hs
@@ -103,7 +103,7 @@ class RenderRoute site => Yesod site where
     --
     -- Since 1.4.23
     urlParamRenderOverride :: site
-                           -> SessionMap -- ^ Session information
+                           -> SessionMap -- ^ Session information available before handler or middleware is processed
                            -> ResolvedApproot -- ^ Approot resolved to Text
                            -> Route site
                            -> [(T.Text, T.Text)] -- ^ query string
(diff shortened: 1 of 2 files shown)


sbditto85:
thoughts @snoyberg? I'm guessing its best to just encode the data in sub domains.

snoyberg:
I haven't had time to think about this deeply, but this feels rather special-cased. That said, I don't have a great idea on better ways to do it in general. There may be some approach of allowing you to override the URL-rendering function manually inside `Handler` and letting all of the Shakespearean templates use that one, but I'm not certain.

sbditto85:
That's ok, it seems the best solution is to just encode the meaning into subdomains. So that's what we'll do.

PR #1610, opened 2019-07-08
Make Routes.Parse use rio

leifmetcalf:
This replaces #1609 

One thing I noticed doing this is that `lineContinuations` will keep empty strings in its input, is that ok? i.e.

`lineContinuations [] rest = []:rest`, not `lineContinuations [] rest = rest`

### DIFF
--- a/yesod-core/src/Yesod/Routes/Parse.hs
+++ b/yesod-core/src/Yesod/Routes/Parse.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE PatternGuards #-}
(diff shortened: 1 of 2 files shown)


snoyberg:
This is introducing a lot of CI breakages. Before we go deeper on this, let me just ask: what's the motivation behind this? I haven't seen a reason why the current code is actually a problem.

leifmetcalf:
I'm trying to understand the yesod source (still new to Haskell), and I thought I might try improve some parts when I understand them. I'll stop if it isn't helpful.

snoyberg:
Reviewing large PRs like this and working through how it's breaking CI ends up taking a significant amount of my time. If there's not a clear advantage to the new code, I'd rather leave it as-is. Small doc fixes, typo corrections, or changes that fix issues are definitely still appreciated!

leifmetcalf:
Ok, I'll close the PR.

PR #1613, opened 2019-07-18
Add mreqMsg

jezen:
When a form has multiple `mreq` fields and the user submits a form with more than one of these values missing, they see an ambiguous collection of "Value is required" messages. The `mreqMsg` function allows you to provide your own MsgValueRequired message for a better UX.

- [✓] Bumped the version number
- [✓] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [✓] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddocks for new, public APIs

After submitting your PR:

- [✓] Update the Changelog.md file with a link to your PR

### DIFF
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -20,6 +20,7 @@ module Yesod.Form.Functions
     , wreq
     , wopt
     , mreq
+    , mreqMsg
     , mopt
     , areq
     , aopt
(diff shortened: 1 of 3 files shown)


yesod-form/Yesod/Form/Functions.hs
```diff
@@ -175,6 +176,23 @@ mreq :: (RenderMessage site FormMessage, HandlerSite m ~ site, MonadHandler m)
      -> MForm m (FormResult a, FieldView site)
 mreq field fs mdef = mhelper field fs mdef (\m l -> FormFailure [renderMessage m l MsgValueRequired]) FormSuccess True
```
snoyberg: What do you think about rewriting this function to use `mreqMsg`?

jezen: I think that sounds sensible. I'll try that.

jezen: Fixed in https://github.com/yesodweb/yesod/pull/1613/commits/e5e39f353d3e60ea767f46d3b2491a95ef464dd2.

Issue #1616, opened 2019-08-11
addPostParam in yesod-test doesn't URI-encode


league:
I'm a little surprised to find that `addPostParam` doesn't URI-encode its value. For example, if I try to test a login form with:

```hs
  request $ do
    setUrl LoginR
    setMethod "POST"
    addPostParam "email" "x@y.z"
    addPostParam "password" "6TW0#B+3/N"
```

It comes through to the handler as:

```
POST /auth/login
  Params: [("password","6TW0#B 3/N"),("email","x@y.z")]
```

(Note the `+` converted to a space.) Even worse, if the value contains `&`, such as in `"G&&E>Q$1MU"` – I'm getting:

```
POST /auth/login
  Params: [("password","G"),("",""),("E>Q$1MU",""),("email","x@y.z")]
```

Should we apply URI-encoding to the key-value pairs when constructing the request body? Or at least maybe the haddock for `addPostParam` should state that the key-value are required to be URI-safe already?

Thanks!

snoyberg:
Seems like this is a bug. Would you be interested in sending a PR? Bonus points if it includes a regression test :)

PR #1617, opened 2019-08-13
URL-encode POST parameters in yesod-test

league:
Addresses #1616. Includes tests that fail without the corresponding change to the code. Potentially this change could cause somebody's tests to fail, if they were already working around the issue (because URL-encoding is not idempotent). 

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -1242,9 +1242,8 @@ request reqBuilder = do
       BS8.concat $ separator : [BS8.concat [multipartPart p, separator] | p <- parts]
     multipartPart (ReqKvPart k v) = BS8.concat
       [ "Content-Disposition: form-data; "
-      , "name=\"", enc k, "\"\r\n\r\n"
-      , enc v, "\r\n"]
-      where enc = H.urlEncode False . TE.encodeUtf8
+      , "name=\"", TE.encodeUtf8 k, "\"\r\n\r\n"
+      , TE.encodeUtf8 v, "\r\n"]
     multipartPart (ReqFilePart k v bytes mime) = BS8.concat
       [ "Content-Disposition: form-data; "
       , "name=\"", TE.encodeUtf8 k, "\"; "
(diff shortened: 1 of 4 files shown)


league:
**Wait –** sorry, I got something wrong in the way it works for multipart/form-data. (I only ran into the bug with x-www-form-urlencoded.) Will push something new when I'm more confident!

league:
**Okay,*** I think this is right (08a9632). multipart/form-data didn't need the fix, only x-www-form-urlencoded. (I fooled myself into thinking it did because of a confounding difference in one test returning text/plain and the other text/html.) Please squash when merging!

yesod-test/Yesod/Test.hs
```diff
@@ -1266,14 +1266,14 @@ request reqBuilder = do
                           ([ ("Cookie", cookieValue) ] ++ headersForPostData rbdPostData)
                           method extraHeaders urlPath urlQuery)
         simpleRequestBody' (MultipleItemsPostData x) =
-          BSL8.fromChunks $ return $ TE.encodeUtf8 $ T.intercalate "&"
-          $ map singlepartPart x
+          BSL8.fromChunks $ return $ H.renderSimpleQuery False
```
snoyberg: I like this change.

league:
> Please add a ChangeLog comment. Also, it doesn't look like the modified test case is testing both url-encoded and form data request types, am I correct?

I believe it does test both types. The `byLabel` on line 160 tests passing through special characters, and that example also uses `fileByLabel`, which necessitates encoding using multipart/form-data.

yesod-test/test/main.hs
```diff
@@ -145,21 +145,25 @@ main = hspec $ do
...
             yit "labels" $ do
                 get ("/form" :: Text)
                 statusIs 200
 
                 request $ do
                     setMethod "POST"
                     setUrl ("/form" :: Text)
-                    byLabel "Some Label" "12345"
+                    byLabel "Some Label" "foo+bar%41<&baz"
                     fileByLabel "Some File" "test/main.hs" "text/plain"
```
league: Here's the test case that passes through special characters with multipart/form-data. (This test did not fail previously, so the bug I'm fixing here is only with form-url-encoded. I thought it failed at one point, because the endpoint this test is accessing returns HTML instead fo text, so `&` comes through as `&amp;` etc.

snoyberg: Got it, thanks for clarifying.

PR #1619, opened 2019-08-19
Dummy: Add support for JSON submissions

### LINKED ISSUE #1618: I'm willing to open a PR to add JSON support to Yesod.Auth.Dummy
From what I understand `Yesod.Auth.Dummy` does not support JSON parameters. In fact it uses `runInputPost` which I believe only takes care of POST parameters.

I've been playing with Haskell and Yesod for a few weeks while writing a series of blog posts. I wanted to have `Yesod.Auth.Dummy` play nice with both form submissions and API requests. That's why I patched `Yesod.Auth.Dummy` to something like this:

```hs
parser :: Value -> Parser Text
parser = withObject "ident" (\obj -> do
                ident <- obj .: "ident"
                return ident)

authDummy' :: YesodAuth m => AuthPlugin m
authDummy' =
    AuthPlugin "dummy" dispatch login
  where
    dispatch "POST" [] = do
        result <- runInputPostResult $ ireq textField "ident"
        case result of
          FormSuccess ident ->
            setCredsRedirect $ Creds "dummy" ident []
          _ -> do
            (result :: Result Value) <- parseCheckJsonBody
            case result of
              Success val -> do
                let mIdent = parseEither parser val
                case mIdent of
                  Right ident ->
                    setCredsRedirect $ Creds "dummy" ident []
                  Left err ->
                    invalidArgs [T.pack err]
              Error err ->
                invalidArgs [T.pack err]
    dispatch _ _ = notFound
    url = PluginR "dummy" []
    login authToMaster = do
        request <- getRequest
        toWidget [hamlet|
$newline never
<form method="post" action="@{authToMaster url}">
    $maybe t <- reqToken request
        <input type=hidden name=#{defaultCsrfParamName} value=#{t}>
    Your new identifier is: #
    <input type="text" name="ident">
    <input type="submit" value="Dummy Login">
|]
```

If anything I said above makes any sense, would you be willing to accept a PR to have `Yesod.Auth.Dummy` accepting JSON parameters? I'm ready to dig deeper and improve the code. Keep in mind, this would be my first PR in Haskell and Yesod, thus I'll prolly need some help.

Yesod is awesome, thanks for all your work!

3v0k4:
Fixes #1618 

- [x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddocks for new, public APIs (**see below**)

After submitting your PR:

Extra stuff:

- `setCredsRedirect` internally does [`provideJsonMessage "Login Successful"`](https://github.com/yesodweb/yesod/blob/master/yesod-auth/Yesod/Auth.hs#L370); the message doesn't make much sense whenever using the plugin without making use of the session (i.e. cookie). I guess it's not a biggie? In my case, I have `authenticate` setup to add the user to the database so logging in actually means registering.
- It seems like Haddock does not support `@since` for part of a paragraph. That's why I've added `This plugin supports form submissions via JSON (since 1.6.8).`. Please let me know if there's a better way.
- It's my first PR in Haskell: I'm super happy to get feedback and iterate on this.

### DIFF
--- a/yesod-auth/Yesod/Auth/Dummy.hs
+++ b/yesod-auth/Yesod/Auth/Dummy.hs
@@ -2,24 +2,67 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 -- | Provides a dummy authentication module that simply lets a user specify
--- his/her identifier. This is not intended for real world use, just for
--- testing.
+-- their identifier. This is not intended for real world use, just for
+-- testing. This plugin supports form submissions via JSON (since 1.6.8).
+--
+-- = Using the JSON Login Endpoint
+--
+-- We are assuming that you have declared `authRoute` as follows
+--
+-- @
+--       Just $ AuthR LoginR
+-- @
+--
+-- If you are using a different one, then you have to adjust the
+-- endpoint accordingly.
+--
+-- @
+--       Endpoint: \/auth\/page\/dummy
+--       Method: POST
+--       JSON Data: {
+--                      "ident": "my identifier"
+--                  }
+-- @
+--
+-- Remember to add the following headers:
+--
+--     - Accept: application\/json
+--     - Content-Type: application\/json
+
 module Yesod.Auth.Dummy
     ( authDummy
     ) where
 
 import Yesod.Auth
-import Yesod.Form (runInputPost, textField, ireq)
+import Yesod.Form (FormResult(..), runInputPostResult, textField, ireq)
 import Yesod.Core
+import Data.Text (Text)
+import qualified Data.Text as T
+import Data.Aeson.Types (Result(..), Parser)
+import qualified Data.Aeson.Types as A (parseEither, withObject)
+
+identParser :: Value -> Parser Text
+identParser = A.withObject "Ident" (.: "ident")
 
 authDummy :: YesodAuth m => AuthPlugin m
 authDummy =
     AuthPlugin "dummy" dispatch login
   where
     dispatch "POST" [] = do
-        ident <- runInputPost $ ireq textField "ident"
-        setCredsRedirect $ Creds "dummy" ident []
+        formResult <- runInputPostResult $ ireq textField "ident"
+        eIdent <- case formResult of
+          FormSuccess ident ->
+            return $ Right ident
+          _ -> do
+            (jsonResult :: Result Value) <- parseCheckJsonBody
+            case jsonResult of
+              Success val -> return $ A.parseEither identParser val
+              Error   err -> return $ Left err
+        case eIdent of
+          Right ident -> setCredsRedirect $ Creds "dummy" ident []
+          Left  err   -> invalidArgs [T.pack err]
     dispatch _ _ = notFound
     url = PluginR "dummy" []
     login authToMaster = do
--- a/yesod-auth/Yesod/Auth/Dummy.hs
+++ b/yesod-auth/Yesod/Auth/Dummy.hs
@@ -36,10 +36,9 @@ module Yesod.Auth.Dummy
     ) where
 
 import Yesod.Auth
-import Yesod.Form (FormResult(..), runInputPostResult, textField, ireq)
+import Yesod.Form (runInputPost, textField, ireq)
 import Yesod.Core
 import Data.Text (Text)
-import qualified Data.Text as T
 import Data.Aeson.Types (Result(..), Parser)
 import qualified Data.Aeson.Types as A (parseEither, withObject)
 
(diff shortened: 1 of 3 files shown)


3v0k4:
@snoyberg this is ready to review. Unfortunately, as said in the PR description I can't find a solution to the `@since`. 

Please feel free to share all the feedback you have!

yesod-auth/Yesod/Auth/Dummy.hs
```diff
@@ -2,24 +2,67 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 -- | Provides a dummy authentication module that simply lets a user specify
--- his/her identifier. This is not intended for real world use, just for
--- testing.
+-- their identifier. This is not intended for real world use, just for
+-- testing. This plugin supports form submissions via JSON (since 1.6.8).
```
snoyberg: For this kind of comment, the `@since` syntax wouldn't be appropriate anyway, this is the right way to do it.

yesod-auth/Yesod/Auth/Dummy.hs
```diff
@@ -2,24 +2,67 @@
...
+identParser = A.withObject "Ident" (.: "ident")
 
 authDummy :: YesodAuth m => AuthPlugin m
 authDummy =
     AuthPlugin "dummy" dispatch login
   where
     dispatch "POST" [] = do
-        ident <- runInputPost $ ireq textField "ident"
-        setCredsRedirect $ Creds "dummy" ident []
+        formResult <- runInputPostResult $ ireq textField "ident"
```
snoyberg: This changes the behavior regarding error messages slightly. I'd recommend one of the following:

* Try using `parseCheckJsonBody` first and if that fails, try `runInputPost`
* Check the mime type manually and then branch on which code path to take

3v0k4: You mean that we don't want to show the error that `parseCheckJsonBody` generates (i.e. `Non-JSON content type: Just "application/x-www-form-urlencoded"`) when the mime type was `application/x-www-form-urlencoded` but `runInputPostResult` returned something else than `FormSuccess`, correct?

For checking the mime type I guess I could do what Yesod.Core.Json does with `lookupHeader` (https://github.com/yesodweb/yesod/blob/master/yesod-core/src/Yesod/Core/Json.hs#L136)
```hs
    mct <- lookupHeader "content-type"
    case fmap (B8.takeWhile (/= ';')) mct of
        Just "application/json" -> parseInsecureJsonBody
        _ -> return $ J.Error $ "Non-JSON content type: " ++ show mct
```

When it comes to the first suggestion (Try using `parseCheckJsonBody` first and if that fails, try `runInputPost`), wouldn't it create the same problem as described at the beginning of this comment. I mean, if we had

```hs
    dispatch "POST" [] = do
      (jsonResult :: Result Value) <- parseCheckJsonBody
      eIdent <- case jsonResult of
        Success val -> return $ A.parseEither identParser val
        Error   err -> return $ Left err
      case eIdent of
        Right ident -> setCredsRedirect $ Creds "dummy" ident []
        Left   _        -> do
          ident <- runInputPost $ ireq textField "ident"
          setCredsRedirect $ Creds "dummy" ident []
```

Then when `A.parseEither identParser val` returns Left because the JSON is of the wrong shape in a `application/json` request, we would try to run `runInputPost` which I feel is wrong. In fact, `runInputPost` would be the one calling `invalidArgs`. And prolly that error is misleading because a JSON request should never get the error that `runInputPost` generates.

snoyberg: > When it comes to the first suggestion (Try using parseCheckJsonBody first and if that fails, try runInputPost), wouldn't it create the same problem as described at the beginning of this comment.

Yes, but:

1. That would be consistent with the current behavior, so at least it's not a change
2. The error message in the JSON case is less important, since it's less likely to be shown to a user

The header check that you mention would work, and is in line with what I was thinking. But it also makes me think that we should have a function in `yesod-core` to check if the content type is JSON-ish, so we don't need to duplicate that code here.

3v0k4: Thanks for the explanation! I changed the code as you suggested:
> Try using `parseCheckJsonBody` first and if that fails, try `runInputPost`

Please review whenever convenient.

I'd like to take a look at
> makes me think that we should have a function in yesod-core to check if the content type is JSON-ish

Are you willing to accept a PR for that? Any recommendations before I start working on it?

snoyberg: Yes, I'd definitely be up for a PR for that. I didn't have any specific recommendations in mind. What do you think about pausing this PR for the moment, waiting for that new function to be available, and then updating this PR to use it? Since it's all in the same repo, I'd also be OK with simply updating this PR to add the new function to yesod-core, whichever you prefer.

3v0k4: If this PR is fine in your opinion, would you be willing to merge? If I manage to take care of the other one I'm happy to come back and change the code here once again

snoyberg: Sure, makes sense.

snoyberg:
Thanks!

PR #1620, opened 2019-08-19
Move JavaScript form submission to script block

jezen:
If someone wants their website to score a good grade on a security vulnerability scanner like Mozilla Observatory, they will need to enable the Content Security Policy header. When using CSP, it is possible to explicitly allow inline JavaScript in `<script>` tags by specifying the sha256 of the snippet. However the same is _not_ true of any JavaScript included in a HTML attribute like `onload`.

This changes moves the JavaScript form submission out of the `onload` HTML attribute and into a `<script>` tag so the user can add the hash of this script to their explicitly-allowed `script-src` list, and they can avoid using undesirable CSP rules like `unsafe-inline`.

Without explicitly allowing this script when using CSP, the script would fail and the user would have to click the button to continue.


snoyberg:
Thanks!

Issue #1621, opened 2019-08-21
Dynamic jsAttributes


jezen:
I need to add a `nonce` attribute to the `<script>` tag that's generated by `widgetFile`.

I have noticed that by using the Yesod typeclass method [`jsAttributes`](http://hackage.haskell.org/package/yesod-core-1.6.15/docs/Yesod-Core.html#v:jsAttributes), I can add arbitrary attributes to that `<script>` tag, however I don't see a way to add a randomised per-request value. A nonce (for CSP) _must_ be randomised on every request.

I am happy to do the work of implementing this and sending a PR, but I actually don't really know where to start with this. Ideally I would like to be able to run a `Handler` for every request (like a middleware). In my case, this `Handler` would generate a nonce, store it somewhere (like a TVar, or per-request cache, or anything else), and then return the nonce.

For additional context, I will share how I am currently handling nonces for static files:

```
-- …
import qualified Data.ByteString.Base64        as B64
import           Data.UUID                     (toASCIIBytes)
-- …

generateRequestNonce :: MonadHandler m => m Text
generateRequestNonce = do
  nonces <- cacheGet >>= return . fromMaybe Set.empty
  nonce <- decodeUtf8 . B64.encode . toASCIIBytes <$> liftIO nextRandom
  cacheSet $ Set.insert nonce nonces
  return nonce

addScriptWithNonce :: MonadWidget m => Route (HandlerSite m) -> m ()
addScriptWithNonce route = generateRequestNonce >>= \n ->
  addScriptAttrs route [("nonce", n)]

addScriptRemoteWithNonce :: MonadWidget m => Text -> m ()
addScriptRemoteWithNonce uri = generateRequestNonce >>= \n ->
  addScriptRemoteAttrs uri [("nonce", n)]

```

The added code I am envisioning would look something like

```
instance Yesod App where

  jsAttributesHandler = do
    nonce <- generateRequestNonce
    return [("nonce", nonce)]
```

Does this seem feasible? Or do I have the wrong approach entirely?

snoyberg:
I don't have an opinion on this. I've never heard of this nonce requirement in CSP, and I have no idea what its purpose would be. That said, I don't think `jsAttributes` is intended for truly static files, but rather for widget-generated Javascript files.

jezen:
> I don't have an opinion on this. I've never heard of this nonce requirement in CSP, and I have no idea what its purpose would be.

[...]

PR #1622, opened 2019-08-23
Add jsAttributesHandler

jezen:
This adds `jsAttributesHandler` to run arbitrary Handler code before building the attributes map for the script tag generated by `widgetFile`. This is useful if you need to add a randomised nonce value to that tag.

Closes https://github.com/yesodweb/yesod/issues/1621

After submitting your PR:


yesod-core/src/Yesod/Core/Class/Yesod.hs
```diff
@@ -249,6 +249,17 @@ class RenderRoute site => Yesod site where
...
 
+    -- | Same as @jsAttributes@ but allows you to run arbitrary Handler code
+    --
+    -- This is useful if you need to add a randomised nonce value to the script
+    -- tag generated by @widgetFile@. If you use both this function and
+    -- @jsAttributes@, the two attribute maps will be merged, favouring the
+    -- keys from @jsAttributes@ in case of a collision.
+    --
+    -- @since 1.6.16
+    jsAttributesHandler :: HandlerFor site [(Text, Text)]
```
snoyberg: The way we've handled this for other. similar cases is to have a default implementation of `jsAttributesHandler` which uses `jsAttributes`. Then we document that if `jsAttributesHandler` is overridden, `jsAttributes` is ignored.

jezen: Ok, I think that is the better approach. I'll make that change.

yesod-core/src/Yesod/Core/Class/Yesod.hs
```diff
@@ -249,6 +249,16 @@ class RenderRoute site => Yesod site where
...
 
+    -- | Same as @jsAttributes@ but allows you to run arbitrary Handler code
+    --
+    -- This is useful if you need to add a randomised nonce value to the script
+    -- tag generated by @widgetFile@. If this function is overridden then
+    -- @jsAttributes@ is ignored.
+    --
+    -- @since 1.6.16
+    jsAttributesHandler :: HandlerFor site [(Text, Text)]
+    jsAttributesHandler = getYesod >>= return . jsAttributes
```
snoyberg: This can also be written as `jsAttributes <$> getYesod`, but there difference isn't that improtant.

jezen: I'm happy to make that change, and also some more hlint-driven cleanup along the way.

Issue #1614, opened 2019-07-26
parseCheckJsonBody rejects +json content-types


whittle:
Using yesod-core-1.6.14, it looks like `parseCheckJsonBody` is checking to see if the content-type is exactly "application/json". This check fails on other valid JSON-derived content-types, e.g. "application/json-patch+json". 

For my PATCH endpoints, I currently use an alternative to `parseCheckJsonBody` that checks for the json-patch content-type. Would it be acceptable if I submitted a PR that liberalizes `parseCheckJsonBody` to accept content-types of the form `/^application\/([\w-]*\+)?json$/`?

snoyberg:
Is there an RFC or similar that documents valid JSON mine types?

snoyberg:
OK, cool, thanks. PR welcome then for this addition.

PR #1625, opened 2019-09-06
Content type check

3v0k4:
This is still a work-in-progress. We've discussed in a [previous PR](https://github.com/yesodweb/yesod/pull/1619#discussion_r316180937) having something in core to check if the content-type is JSON-ish.

I thought of creating a function like `hasJsonMime :: MonadHandler m => m (Bool)`. But saying if a mime is either JSON or anything else does not seem too useful (i.e. if it's not JSON it could be really anything else). 

Another way would be to have something like

```hs
data Mime
  = Html
  | Json
  | ...
```
And a function `lookupContentType :: MonadHandler m => m (Mime)`. But I'm not sure it's a good approach.

At the moment I've added a `lookupContentType :: MonadHandler m => m (Maybe S8.ByteString)`. But I'm not sure having the caller `case`ing on a `Maybe S8.ByteString` is a good idea.

Also, I see in the code both the terms "content type" and "mime", do you have a preference on any of the two? Would you use them both but in different contexts?

How would you approach all of this?

---

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Handler.hs
+++ b/yesod-core/src/Yesod/Core/Handler.hs
@@ -191,6 +191,7 @@ module Yesod.Core.Handler
     , defaultCsrfParamName
       -- ** Checking CSRF Headers or POST Parameters
     , checkCsrfHeaderOrParam
+    , lookupContentType
     ) where
 
 import           Data.Time                     (UTCTime, addUTCTime,
@@ -1686,3 +1687,8 @@ getRouteToParent = liftSubHandler $ SubHandlerFor $ return . rheRouteToMaster . 
 
 getSubCurrentRoute :: MonadHandler m => m (Maybe (Route (SubHandlerSite m)))
 getSubCurrentRoute = liftSubHandler $ SubHandlerFor $ return . rheRoute . handlerEnv
+
+lookupContentType :: MonadHandler m => m (Maybe S8.ByteString)
+lookupContentType = do
+    mct <- lookupHeader "content-type"
+    return $ fmap (S8.takeWhile (/= ';')) mct
--- a/yesod-auth/Yesod/Auth/Dummy.hs
+++ b/yesod-auth/Yesod/Auth/Dummy.hs
@@ -39,8 +39,9 @@ import Yesod.Auth
 import Yesod.Form (runInputPost, textField, ireq)
 import Yesod.Core
 import Data.Text (Text)
+import qualified Data.Text as T
 import Data.Aeson.Types (Result(..), Parser)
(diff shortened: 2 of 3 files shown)


snoyberg:
In #1614, we're discussing making the JSON content type detection a bit more sophisticated. I think it would make more sense to provide an explicit function that says "is the content type some variation of JSON", instead of putting the `application/json` checking in multiple places.

3v0k4:
I've pushed another commit. It still is very work-in-progress. Please let me know if we should iterate more or if I can clean up and ask you for a final review. 

As always, I'd appreciate as much feedback as you can provide. I'm learning a ton by working on Yesod!

yesod-core/src/Yesod/Core/Handler.hs
```diff
@@ -1686,3 +1687,11 @@ getRouteToParent = liftSubHandler $ SubHandlerFor $ return . rheRouteToMaster .
 
 getSubCurrentRoute :: MonadHandler m => m (Maybe (Route (SubHandlerSite m)))
 getSubCurrentRoute = liftSubHandler $ SubHandlerFor $ return . rheRoute . handlerEnv
+
+hasJsonContentType :: MonadHandler m => m (Either String ())
```
snoyberg: I think I'd go in the direction of returning a `Bool`, I don't see an advantage to the `Either` return value. Also, when ready, please include a Haddock comment including a `@since` declaration.

3v0k4: I went with Either cause parseCheckJsonBody in the error case returns the content-type (you can see it in the changes for this PR). I guess we should not change that. Do you see a better way than using a Either?

snoyberg: It's possible at the call site to ask a second time for the mime type. Another option would be a return value such as:

```haskell
data IsJsonMime = JsonMime | NonJsonMime !ByteString
```

3v0k4:
@snoyberg thanks a lot for the support in this PR, I appreciate it!

I'm not sure I like either of the solutions [we discussed](https://github.com/yesodweb/yesod/pull/1625#discussion_r322006376). In particular, returning a bool and asking again for the mime means checking the content-type header twice, which I don't like much. Using a data structure like `data IsJsonMime = JsonMime | NonJsonMime !ByteString` seems to be putting too much emphasis on JSON vs everything else. Also, I agree with you that `Either String ()` doesn't look that good.

I'm leaving this PR to you or somebody else to take care of. Feel free to close. Thanks again for the opportunity of contributing to Yesod for the second time!

jezen:
Closing due to inactivity.

PR #1630, opened 2019-09-13
Use original type var names when generating type synonyms

### LINKED ISSUE #1629: Can't use mkYesod with type arguments
Info:
* Debian bullseye
* `stack 2.1.3`
* `lts-13.20`

Repro file:
```haskell
{-# Language QuasiQuotes #-}
{-# Language TemplateHaskell #-}
{-# Language TypeFamilies #-}
{-# Language OverloadedStrings #-}

import Yesod.Core

data App a

mkYesod "App a" [parseRoutes|
  /up UpR GET
|]

getUpR :: Handler ()
getUpR = pure ()
```
Errors out with 
```
/home/googleson78/test.hs:8:1: error:
    Not in scope: type variable ‘a’
  |
8 | mkYesod "App a" [parseRoutes|
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...

/home/googleson78/test.hs:8:1: error:
    Not in scope: type variable ‘a’
  |
8 | mkYesod "App a" [parseRoutes|
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

I believe this is because of the two type synonyms that are generated:
```haskell
type Handler = HandlerFor (App a)
type Widget = WidgetFor (App a) ()
```
which are not valid, because they also need the type parameter.

googleson78:
(Potentially) fixes #1629 

This makes the synonyms valid when the using type params, e.g.
```haskell
mkYesod "App a" ...
```

However I am not sure if this is the "right" fix, i.e. if I've accidentally fixed it by doing something wrong that breaks other stuff. I've added my thoughts as a comment on the review.

I would love to add a test for this, if someone could give me a hint on how to do so.

Additionally I am not sure **which** version to bump, and **how much**, so I've left a bit of the template below as a `TODO`.

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Internal/TH.hs
+++ b/yesod-core/src/Yesod/Core/Internal/TH.hs
@@ -160,7 +160,7 @@ mkYesodGeneral appCxt' namestr mtys isSub f resS = do
             , renderRouteDec
             , [routeAttrsDec]
             , resourcesDec
-            , if isSub then [] else masterTypeSyns vns site
+            , if isSub then [] else masterTypeSyns (map mkName mtys) site
             ]
     return (dataDec, dispatchDec)
 


yesod-core/src/Yesod/Core/Internal/TH.hs
```diff
@@ -160,7 +160,7 @@ mkYesodGeneral appCxt' namestr mtys isSub f resS = do
             , renderRouteDec
             , [routeAttrsDec]
             , resourcesDec
-            , if isSub then [] else masterTypeSyns vns site
+            , if isSub then [] else masterTypeSyns (map mkName mtys) site
```
googleson78: It feels like `vns` above is always of length 0? The arity of the data constructor should be the same as the amount of arguments that got parsed from `mkYesodWithParser` (modulo some compliated type/data family stuff). We also need to use the exact same names, because those are placed in all the places the generated type is used. Alternatively we could figure out how to place the generated `t1, .. tn`s in both places, because it isn't being done right now.

snoyberg:
For a PR like this, including a test case that fails with the original code and is fixed by this change would be great. It's difficult to be certain it's doing what we think it's doing.

googleson78:
Is there a test for TH generation I could look at? (or golden tests?) I can't seem to find them - maybe I'm looking for the wrong thing.

snoyberg:
There are lots of tests in the `test` directory. I think using your terminology I'm recommending a golden test.

jezen:
Since the attached issue is actually fixed by https://github.com/yesodweb/yesod/pull/1697, I think we can close this PR.

PR #1631, opened 2019-09-29
Support categories for Atom and RSS newsfeed

yogsototh:
Add support for `<categories>` in RSS and Atom feeds:

- RSS: http://www.rssboard.org/rss-specification#ltcategorygtSubelementOfLtitemgt
- Atom: https://tools.ietf.org/html/rfc4287#section-4.2.2

Note I'm not using `url` type for domains as taxonomies URI will almost always be external URL.

---

After submitting your PR:

### DIFF
--- a/yesod-newsfeed/Yesod/FeedTypes.hs
+++ b/yesod-newsfeed/Yesod/FeedTypes.hs
@@ -2,6 +2,7 @@ module Yesod.FeedTypes
     ( Feed (..)
     , FeedEntry (..)
     , EntryEnclosure (..)
+    , EntryCategory (..)
     ) where
 
 import Text.Hamlet      (Html)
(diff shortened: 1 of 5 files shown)


yesod-newsfeed/yesod-newsfeed.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-newsfeed
-version:         1.6.1.0
+version:         1.7.0.0
```
snoyberg: Please also include a ChangeLog entry.

yesod-newsfeed/Yesod/FeedTypes.hs
```diff
@@ -51,4 +64,5 @@ data FeedEntry url = FeedEntry
       -- rel=enclosure>
       --
       -- @since 1.5
+    , feedEntryCategories :: [EntryCategory]
```
snoyberg: This should also include a `@since` entry.

snoyberg:
Ignoring the cabal failures, they seem unrelated.

PR #1642, opened 2019-11-17
WIP - Add testModifySite

MaxGabriel:
~*This PR is a draft to propose a new function `testModifySite`*~

In most Yesod apps, there is an `App` datatype at the core of the application, holding all configuration and state for the app (e.g. settings, connection pools, `IORef`s, etc.). In the tests, this is setup on a per-test basis. This PR adds a new function, `testModifySite`, to modify that site in the middle of a test. This is the signature:

```haskell
testModifySite :: YesodDispatch site => (site -> site) -> Middleware -> YesodExample site ()
```

Here are some use cases I'm imagining:

### Testing how modifying settings changes behavior

The yesod scaffold comes with an `AppSettings` type you use to configure your app. An app may want to have a test like so:

```
it "only fires missiles if config allows it" $ do
  -- Many
  -- common
  -- lines
  -- of
  -- database
  -- configuration
  post FireMissilesR
  -- Assert nothing happens
  testModifySite (\site -> site { appSettings = (appSettings site) { appSettingsFireMissiles = True } }) id
  post FireMissilesR
  -- Assert missiles are fired
```

### Changing out stubbed functions

The use case we have at work is we have a `Stubs` datatype in our app, and we use it to stub out functions. This looks roughly like so:

```
data App = App
  { appTestStubs :: Stubs
  ...
  }

data Stubs = Stubs 
  { stubMailerHandlerSendEmail :: forall m. (MonadIO m) => Maybe (m ())
  ...
  }

withStub :: (Monad m, HasApp m) => (Stubs -> Maybe (m a)) -> m a -> m a
withStub f realImpl = do
  stubs <- appTestStubs <$> getApp
  fromMaybe realImpl (f stubs)

handlerSendEmail message = withStub stubMailerHandlerSendEmail $ do
  -- Real implementation here
```

And in production, we set `stubMailerHandlerSendEmail` to `Nothing`, in which case our codebase uses the production implementation, but in tests we can use the stubbed function to customize behavior.

`testModifySite` would allow us to modify that stub in the middle of a test, so we could have tests like this:

```
it "records sent emails in the database if sending succeeds" $ do
  -- Lots
  -- of 
  -- test 
  -- setup
  post SendEmailR
  -- Assert that SentEmail record exists in database
  testModifySite (\site -> site { appTestStubs = (appTestStubs site) { stubMailerHandlerSendEmail = Just . liftIO $ throwIO EmailSendException } }) id
  post SendEmailR
  -- Assert that new email record was not created
```

<hr>

A gotcha about this function is that the `yedSite` and `yedApp` parameters are tied together. So, once the site is modified, a new `yedApp` must be created from it using `toWaiAppPlain` then calling a new middleware function. I think this is OK though?

One could maybe drop the `Middleware` parameter of `testModifySite`, but the `Middleware` initially provided would have to be stored in `YesodExampleData` for later, which would be a breaking change. Also, it adds complexity to `YesodExampleData` just for the convenience of this function.

<hr>

If this function sounds good, I'll add another function called `testSetSite` that outright replaces the site value.

### DIFF
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -2,7 +2,7 @@
 
 ## 1.6.8
 
-Add `testModifySite` function [#](https://github.com/yesodweb/yesod/pull/)
+Add `testModifySite` function [#1642](https://github.com/yesodweb/yesod/pull/1642)
 
 ## 1.6.7
 
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-test
 
+## 1.6.8
+
+Add `testModifySite` function [#](https://github.com/yesodweb/yesod/pull/)
+
 ## 1.6.7
 
 Add `addBasicAuthHeader` function [#1632](https://github.com/yesodweb/yesod/pull/1632)
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -45,6 +45,9 @@ module Yesod.Test
     , ydescribe
     , yit
 
+    -- * Modify test site
+    , testModifySite
+
     -- * Modify test state
     , testSetCookie
     , testDeleteCookie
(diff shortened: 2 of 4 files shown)


MaxGabriel:
After looking at usage a little more, the middleware parameter should probably be:

```
App -> IO Middleware
```

given that that's how the yesod scaffold generates it: https://github.com/yesodweb/yesod-scaffold/blob/3378e25e54472b7ea9f691b460ff46e9f9e8382c/test/TestImport.hs#L44

snoyberg:
Seems reasonable :+1:

MaxGabriel:
Awesome, ready for review 👍

MaxGabriel:
I can take it

Issue #1643, opened 2019-11-18
Function to parse JSON in yesod-test


MaxGabriel:
At work we have a function to parse a response in yesod-test to a JSON value. This seemed pretty widely helpful to me, would you want such a function in yesod-test?

```
-- | Parses the response body from JSON and returns it.
requireJsonParse :: FromJSON a => YesodExample App a
requireJsonParse = do
    YT.withResponse $ \ (SResponse _status headers body) -> do
        unless
            ((hContentType, "application/json; charset=utf-8") `elem` headers)
            (failure $ T.pack $ "Expected `Content-Type: application/json; charset=utf-8` in the headers, got: " ++ show headers)
        case eitherDecode' body of
            Left err -> failure $ T.concat ["Failed to parse JSON response; error: ", T.pack err]
            Right v -> return v
```

It might fit more into the existing naming schemes of yesod-test if it was named `assertJsonParse` (or `assertJSONParse`) and took a string parameter for the assertion name.

snoyberg:
Seems good to me, no objection at all. In general, I have a pretty high level of trust that if you think a function should be added, it should be added :)

jezen:
This has all been merged, so we can close this issue. Thanks @MaxGabriel!

PR #1646, opened 2019-11-24
[yesod-test] Adds requireJSONResponse function

MaxGabriel:
This function checks that a response body is JSON, and parses it into a Haskell value. Having something like this function is pretty essential to using Yesod as a JSON API server, so I think it's a good addition. You can use it to parse a Haskell record directly (usually by adding FromJSON classes to your response types), or parse a Value and pull out individual fields, maybe using something like `aeson-lens` (though probably a testing-specific library would be better).

I debated over these things:

1. The name. I was thinking of something like [assert/require/decode/parse]JSON[Response/Body]. I ultimately went with requireJSONResponse:
	- decode/parse sound like the aeson functions that return Either or Maybe, and I wanted this function to throw an error if it failed
	- I'm open to using `assertJSONResponse`—it matches the other functions (`assertEq`) better—but I think it reads less like English.
	- I chose Response over Body because (a) It also checks the content-type header, which is not in the body (b) "Body" felt slightly in-the-weeds of HTTP; I think "response" is more approachable.
	- Edit: Oh, just realized the function on the serve is called `requireJsonBody`. That kind of rules out that name to avoid duplicate identifier issues.      
2. Should it require the JSON content type? You can definitely have a server that returns JSON without JSON content types, but I think that's a such a bad idea, it's more likely requiring it helps people if they accidentally don't add the header.
3. Should it take a String parameter to add to the error message? This would match `assertEq`, but other functions like `statusIs` don't take a message. Ultimately I went without it, because the messages felt like I was repeating myself: `(comment :: Comment) <- requireJSONResponse "the response has a comment"`

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -125,6 +125,7 @@ module Yesod.Test
     , htmlAnyContain
     , htmlNoneContain
     , htmlCount
+    , requireJSONResponse
 
     -- * Grab information
     , getTestYesod
@@ -195,6 +196,9 @@ import GHC.Exts (Constraint)
 type HasCallStack = (() :: Constraint)
 #endif
 import Data.ByteArray.Encoding (convertToBase, Base(..))
+import Network.HTTP.Types.Header (hContentType)
+import Data.Aeson (FromJSON, eitherDecode')
+import Control.Monad (unless)
 
 {-# DEPRECATED byLabel "This function seems to have multiple bugs (ref: https://github.com/yesodweb/yesod/pull/1459). Use byLabelExact, byLabelContain, byLabelPrefix or byLabelSuffix instead" #-}
 {-# DEPRECATED fileByLabel "This function seems to have multiple bugs (ref: https://github.com/yesodweb/yesod/pull/1459). Use fileByLabelExact, fileByLabelContain, fileByLabelPrefix or fileByLabelSuffix instead" #-}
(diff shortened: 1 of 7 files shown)


MaxGabriel:
This closes #1643

yesod-test/Yesod/Test.hs
```diff
@@ -598,6 +602,29 @@ htmlCount query count = do
...
+requireJSONResponse :: (HasCallStack, FromJSON a) => YesodExample site a
+requireJSONResponse = do
+  withResponse $ \(SResponse _status headers body) -> do
+    let mContentType = lookup hContentType headers
+        isJSONContentType = maybe False (\contentType -> BS8.takeWhile (/= ';') contentType == "application/json") mContentType
+    unless
+        isJSONContentType
+        (failure $ T.pack $ "Expected `Content-Type: application/json` in the headers, got: " ++ show headers)
+    case eitherDecode' body of
+      -- TODO: include full body in error message?
```
MaxGabriel: Could do this if there's a failure. I think it would mostly be quite useful, but would be large and sometimes extremely large (e.g. Base64 encoded images). Thoughts?

yesod-test/Yesod/Test.hs
```diff
@@ -598,6 +602,29 @@ htmlCount query count = do
...
+--
+-- > get CommentR
+-- > (comment :: Comment) <- requireJSONResponse
+--
+-- @since 1.6.9
+requireJSONResponse :: (HasCallStack, FromJSON a) => YesodExample site a
+requireJSONResponse = do
+  withResponse $ \(SResponse _status headers body) -> do
+    let mContentType = lookup hContentType headers
+        isJSONContentType = maybe False (\contentType -> BS8.takeWhile (/= ';') contentType == "application/json") mContentType
```
MaxGabriel: Copied from `acceptsJson`, mostly

snoyberg: Would it work to use [`acceptsJson`](https://www.stackage.org/haddock/lts-14.16/yesod-core-1.6.16.1/Yesod-Core-Json.html#v:acceptsJson) here? That would be preferable, since the set of acceptable JSON mime types may expand in the future.

yesod-test/Yesod/Test.hs
snoyberg: How about including the first 256 (arbitrary number) bytes?

yesod-test/Yesod/Test.hs
MaxGabriel: `acceptsJson` is `MonadHandler`, so we couldn't just drop it in here I don't think, but I could create a new function exposed from yesod-core that takes a list of headers (`[(CI ByteString, ByteString)]`, aka [`RequestHeaders`](https://www.stackage.org/haddock/lts-14.16/http-types-0.12.3/Network-HTTP-Types-Header.html#t:RequestHeaders)) and returned if one of the was a JSON content type?

yesod-test/Yesod/Test.hs
MaxGabriel: This seems pretty reasonable

yesod-test/Yesod/Test.hs
snoyberg: That sounds good

yesod-test/Yesod/Test.hs
MaxGabriel: Cool, implemented the body preview

yesod-test/Yesod/Test.hs
MaxGabriel: Realized that the `Accept` header is slightly different, since it has a list of preferred MIME types. I added a `contentTypeIsJson` function for general use on Content-Type.

MaxGabriel:
Ok I think this is ready for final review. One last thing is should the function be renamed `requireJsonResponse`? This matches Yesod.Core.Json's naming scheme, but not aeson's functions (parseJSON)

MaxGabriel:
I'll take the release 👍

MaxGabriel:
Done 👍

MaxGabriel:
(note: this PR was missing a requirement from yesod-test that it was getting the latest yesod-core for the new content type check. I pushed that directly to master)

PR #1658, opened 2020-01-27
Replace decodeFile with decodeFileEither

v0d1ch:
The title says it all...

### DIFF
--- a/yesod/Yesod/Default/Config.hs
+++ b/yesod/Yesod/Default/Config.hs
@@ -179,8 +179,8 @@ loadConfig :: ConfigSettings environment extra
            -> IO (AppConfig environment extra)
 loadConfig (ConfigSettings env parseExtra getFile getObject) = do
     fp <- getFile env
-    mtopObj <- decodeFile fp
-    topObj <- maybe (fail "Invalid YAML file") return mtopObj
+    etopObj <- decodeFileEither fp
+    topObj <- either (const $ fail "Invalid YAML file") return etopObj
     obj <- getObject env topObj
     m <-
         case obj of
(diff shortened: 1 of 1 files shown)


yesod/Yesod/Default/Config.hs
```diff
@@ -232,9 +232,10 @@ withYamlEnvironment :: Show e
                     -> (Value -> Parser a) -- ^ what to do with the mapping
                     -> IO a
 withYamlEnvironment fp env f = do
-    mval <- decodeFile fp
+    mval <- decodeFileEither fp
     case mval of
-        Nothing -> fail $ "Invalid YAML file: " ++ show fp
-        Just (Object obj)
+        Left err ->
+          fail $ "Invalid YAML file: " <> show fp <> " " <> prettyPrintParseException err
```
MaxGabriel: It looks like tests are failing on old versions of GHC/base, where `<>` isn't in scope

```
yesod                            > /home/vsts/work/1/s/yesod/Yesod/Default/Config.hs:238:40: error:
yesod                            >     • Variable not in scope: (<>) :: m0 a0 -> String -> t1
yesod                            >     • Perhaps you meant one of these:
yesod                            >         ‘<$>’ (imported from Prelude), ‘*>’ (imported from Prelude),
yesod                            >         ‘<$’ (imported from Prelude)
yesod                            >     |
yesod                            > 238 |           fail $ "Invalid YAML file: " <> show fp <> " " <> prettyPrintParseException err

```

v0d1ch: Oh my bad. I fixed it but how did you run those tests? By running `stack test` I didn't see any failures.

MaxGabriel: I didn't run them, that was continuous integration. It tests against a bunch of different configurations. I don't have exact instructions, but you could copy what CI does to test against different configurations (probably just changing the resolver in the stack file)

Issue #948, opened 2015-03-05
Yesod.Auth.Email RegisterR sends verification email if the email is already registered


sevcsik:
I found out when I try to register an email on RegisterR which already exists, instead of displaying an error message, that the user already exists, it sends an an email confirmation request to the given email address, just like if I asked for a password reset.

After going through the sources, I found out that `registerHelper` checks if the email exists, and if it does, it sends a verify email even if it's already verified.

``` haskell
    -- [...]
        mecreds <- lift $ getEmailCreds identifier                           
            registerCreds <-                                                 
                case mecreds of                                              
                    Just (EmailCreds lid _ _ (Just key) email) -> return $ Just (lid, key, email)
                    Just (EmailCreds lid _ _ Nothing email) -> do            
                        key <- liftIO $ randomKey y                          
                        lift $ setVerifyKey lid key                          
                        return $ Just (lid, key, email)                      
                    Nothing                                                  
                        | allowUsername -> return Nothing                    
                        | otherwise -> do                                    
                            key <- liftIO $ randomKey y                      
                            lid <- lift $ addUnverified identifier key       
                            return $ Just (lid, key, identifier)

            case registerCreds of
                Nothing -> loginErrorMessageI dest (Msg.IdentifierNotFound identifier)
                Just (lid, verKey, email) -> do
                    render <- getUrlRender
                    let verUrl = render $ verify (toPathPiece lid) verKey
                    lift $ sendVerifyEmail email verKey verUrl
                    lift $ confirmationEmailSentResponse identifier                                         
    -- [...]
```

Source: https://hackage.haskell.org/package/yesod-auth-1.4.3.1/docs/src/Yesod-Auth-Email.html#registerHelper

Since `registerCreds` is a `Just` when the email exists and it has a key, it sends the verify email again to the user, whiche he/she can use as a password reset link. 

I think the first two cases of `mecreds` should be something like this:

``` haskell
Just (EmailCreds lid _ False (Just key) email) -> return $ Just (lid, key, email)
Just (EmailCreds _ _ True _ _) -> return Nothing
```

So if the email is not yet verified, it just gets the verification key sent again, but if it is verified already, it displays an error message instead of acting like a password reset, which is a bit confusing.

Also, it would be nice to put `registerHelper` under the `YesodAuthEmail` typeclass, so this behavior could be customized.

geraldus:
good catch!

Laquendi:
[...]

PR #1660, opened 2020-02-09
Add methods to YesodAuthEmail that allow custom flow after registration/password reset

charukiewicz:
### Summary

This change moves `registerHelper` (previously a private function in `Yesod.Auth.Email`) into the `YesodAuthEmail` typeclass to allow for customization by end users of the library. It also adds a `passwordResetHelper`, allowing for different flow between registration and password reset behavior.

### Details

The issue was first mentioned in #948 almost five years ago, wherein using `Yesod.Auth.Email` currently has very limited customization capabilities in terms of what happens upon submission of a registration or password reset request.

Moreover, overriding `registerHelper` is necessary if the developer wants to handle additional fields that are sent from the registration form. For example, implementing a honeypot field that attempts to thwart bots requires examining the contents of the additional field upon submission, which was previously impossible, since the developer had no control over what happened after the form was POSTed to the registration action.

This change preserves the default behavior (by adding a `defaultRegisterHelper`) and so it should be fully backwards compatible with previous versions of `yesod-auth`.

I'm not sure what the typical versioning scheme used in the `yesod-*` plugins is, but since this isn't a breaking change, I've currently bumped the version of `yesod-auth` from `1.6.8.1` to `1.6.9`.

### Feedback

I'm happy to receive any feedback about this change. I know `Yesod.Auth.Email` is not the most active module in terms of development, but since it ships with `yesod-auth`, I figure exposing some of its internals and letting developers override some more of its defaults should improve its utility.

-----------

After submitting your PR:

### DIFF
--- a/yesod-auth/Yesod/Auth/Email.hs
+++ b/yesod-auth/Yesod/Auth/Email.hs
@@ -385,6 +385,10 @@ class ( YesodAuth site
       -> AuthHandler site TypedContent
     setPasswordHandler = defaultSetPasswordHandler
 
+
+    registerHelper :: Bool -> Bool -> Route Auth -> AuthHandler site TypedContent
+    registerHelper = defaultRegisterHelper
+
 authEmail :: (YesodAuthEmail m) => AuthPlugin m
 authEmail =
     AuthPlugin "email" dispatch emailLoginHandler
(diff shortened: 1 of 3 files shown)


yesod-auth/Yesod/Auth/Email.hs
```diff
@@ -549,7 +580,7 @@ registerHelper allowUsername forgotPassword dest = do
                     _ -> Nothing
 
     case eidentifier of
-      Left route -> loginErrorMessageI dest route
+      Left failMsg -> loginErrorMessageI dest failMsg
```
charukiewicz: This is just a renaming of the `route` variable to `failMsg` in the local scope of this case statement, since the value is in fact a `Msg` and not an `AuthRoute` as the old name implies.

snoyberg: Thanks!

Issue #1659, opened 2020-02-06
Add meta tag helpers


jezen:
In a few of my projects I maintain a module called `Util.Meta` which just contains a bunch of functions [like these][0]. Is there any reason not to bundle these functions with Yesod? If adding these seems reasonable, I'm happy to send a PR.

[0]: https://ersocon.net/blog/2016/10/26/how-to-add-meta-tags-in-yesod

snoyberg:
Seems fine to me, go for it 💪

PR #1663, opened 2020-03-30
Add functions for setting description and OG meta

jezen:
It's common that a website author will want to add a specific
description and Open Graph image to a given page in their website. These
functions are simple conveniences to add these meta tags to the document
head.

I decided against simply adding all possible meta tags, because not all
of them are useful, and even in the case of Open Graph tags, many of
them should be set only once and they should use the same value for the
entire website. In those cases, it's probably better for the website
author to add those tags in their layout template.

Closes https://github.com/yesodweb/yesod/issues/1659

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-core
 
+## 1.6.18
+
+* Add functions for setting description and OG meta [#1663](https://github.com/yesodweb/yesod/pull/1663)
+
 ## 1.6.17.3
 
 * Support for `unliftio-core` 0.2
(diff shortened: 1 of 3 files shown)


yesod-core/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-core
 
+## 1.6.18
```
snoyberg: Looks like this also needs a `yesod-core.cabal` update.

jezen:
@snoyberg Am I ok to squash and merge this?

snoyberg:
Yes :+1:

Issue #1656, opened 2020-01-26
Remove ListT instances (?)


MaxGabriel:
It looks like ListT is deprecated, based on compiling with `-Wdeprecations`. I'm not familiar with the background on this, but it seems like it should be removed. If others are in agreement about this, it would be an easy issue for a newcomer

```
yesod-core                       > /Users/maximiliantagher/Documents/Clones/Haskell/yesodweb/yesod/yesod-core/src/Yesod/Core/Class/Handler.hs:22:1: warning: [-Wdeprecations]
yesod-core                       >     Module ‘Control.Monad.Trans.List’ is deprecated:
yesod-core                       >       This transformer is invalid on most monads
yesod-core                       >    |
yesod-core                       > 22 | import Control.Monad.Trans.List     ( ListT    )
yesod-core                       >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
yesod-core                       >
yesod-core                       > /Users/maximiliantagher/Documents/Clones/Haskell/yesodweb/yesod/yesod-core/src/Yesod/Core/Class/Handler.hs:79:42: warning: [-Wdeprecations]
yesod-core                       >     In the use of type constructor or class ‘ListT’
yesod-core                       >     (imported from Control.Monad.Trans.List):
yesod-core                       >     Deprecated: "This transformer is invalid on most monads"
yesod-core                       >    |
yesod-core                       > 79 | GO(ListT)
yesod-core                       >    |                                          ^^^^^
yesod-core                       >
yesod-core                       > /Users/maximiliantagher/Documents/Clones/Haskell/yesodweb/yesod/yesod-core/src/Yesod/Core/Class/Handler.hs:79:75: warning: [-Wdeprecations]
yesod-core                       >     In the use of type constructor or class ‘ListT’
yesod-core                       >     (imported from Control.Monad.Trans.List):
yesod-core                       >     Deprecated: "This transformer is invalid on most monads"
yesod-core                       >    |
yesod-core                       > 79 | GO(ListT)
yesod-core                       >    |                                                                           ^^^^^
yesod-core                       >
yesod-core                       > /Users/maximiliantagher/Documents/Clones/Haskell/yesodweb/yesod/yesod-core/src/Yesod/Core/Class/Handler.hs:79:122: warning: [-Wdeprecations]
yesod-core                       >     In the use of type constructor or class ‘ListT’
[...]

PR #1667, opened 2020-04-09
Fix warnings in yesod-core

JonathanLorimer:
I originally set out to resolve this [issue](https://github.com/yesodweb/yesod/issues/1656) but it turned out to be a fairly trivial change, so I handled a couple other warnings that were coming up with `-Wall` enabled. 

I don't think that there were any changes to the logic, and the test suite seems to be passing.

This is my first PR against Yesod so let me know if I missed something!

- ~~Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)~~
- ~~Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddocks for new, public APIs~~

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Class/Handler.hs
+++ b/yesod-core/src/Yesod/Core/Class/Handler.hs
@@ -19,7 +19,6 @@ import Control.Monad.Trans.Class (lift)
 import Data.Conduit.Internal (Pipe, ConduitM)
 
 import Control.Monad.Trans.Identity ( IdentityT)
-import Control.Monad.Trans.List     ( ListT    )
 import Control.Monad.Trans.Maybe    ( MaybeT   )
 import Control.Monad.Trans.Except   ( ExceptT  )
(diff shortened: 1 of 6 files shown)


yesod-core/src/Yesod/Core/Class/Handler.hs
```diff
@@ -19,7 +19,6 @@ import Control.Monad.Trans.Class (lift)
 import Data.Conduit.Internal (Pipe, ConduitM)
 
 import Control.Monad.Trans.Identity ( IdentityT)
-import Control.Monad.Trans.List     ( ListT    )
```
snoyberg, line 22: This is a breaking change, and should not be included right now.

JonathanLorimer: Good point, closing.

PR #1680, opened 2020-06-17
When statusIs fails, print a preview of the body

MaxGabriel:
My team makes frequent use of `statusIs`, but in virtually all cases where `statusIs` fails, we need to add a call to `printBody` to do further debugging, which slows down development.

Following in the footsteps of `requireJSONResponse`, this PR automatically prints a portion of the body when `statusIs` fails, assuming the body looks like a text-based response (e.g. not a JPEG). Assuming the body is too large, it points people to `printBody`, which aids in discoverability of that very helpful function.

Anecdotally, I've found that a status code alone is often misleading and leads people on a wild goose chase, because e.g. many different authorization checks might trigger a 403.

~~I'm opening this PR as a draft to confirm people like the idea of doing this. If so I'll do a closer review of the code (this is my first draft basically), and also write some tests + test the code works in all cases.~~

~~(skipping checklist for now, since this is a draft)~~

Update: Ready for review!

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -278,6 +278,9 @@ import Data.ByteArray.Encoding (convertToBase, Base(..))
 import Network.HTTP.Types.Header (hContentType)
 import Data.Aeson (FromJSON, eitherDecode')
 import Control.Monad (unless)
+import qualified Data.Set as Set
+import qualified Yesod.Core.Content as Content
+import qualified Data.ByteString.Lazy as LBS
 
 {-# DEPRECATED byLabel "This function seems to have multiple bugs (ref: https://github.com/yesodweb/yesod/pull/1459). Use byLabelExact, byLabelContain, byLabelPrefix or byLabelSuffix instead" #-}
 {-# DEPRECATED fileByLabel "This function seems to have multiple bugs (ref: https://github.com/yesodweb/yesod/pull/1459). Use fileByLabelExact, fileByLabelContain, fileByLabelPrefix or fileByLabelSuffix instead" #-}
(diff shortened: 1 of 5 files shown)


yesod-test/Yesod/Test.hs
```diff
@@ -569,17 +572,63 @@ assertEqualNoShow :: (HasCallStack, Eq a) => String -> a -> a -> YesodExample si
...
+        [_, charSet] -> any (`T.isInfixOf` charSet) ["utf-8", "us-ascii"]
+        _ -> False
+
+      isInferredUTF8FromContentType = BS8.takeWhile (/= ';') contentTypeBS `Set.member` assumedUTF8ContentTypes
+
+  in isUTF8FromCharset || isInferredUTF8FromContentType
+
+-- | List of Content-Types that are assumed to be UTF-8 (e.g. JSON)
+assumedUTF8ContentTypes :: Set.Set BS8.ByteString
+assumedUTF8ContentTypes = Set.fromList $ map Content.simpleContentType
```
MaxGabriel, line 611: This list includes everything from `Yesod.Core.Content` that I thought was likely to be UTF-8. This isn't a perfect assumption—XML might be `ISO-8859-1` without specifying it, and this list is obviously incomplete. However, it should be a fairly strong heuristic and cover the vast majority of cases

snoyberg:
I like the idea, but haven't reviewed the code

MaxGabriel:
Cool, no need to review the code yet, I need to clean it up anyway

yesod-test/Yesod/Test/Internal.hs
```diff
@@ -0,0 +1,64 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | This module exposes functions that are used internally by yesod-test.
+-- The functions exposed here are _not_ a stable API—they may be changed or removed without any major version bump.
+--
+-- That said, you may find them useful if your application can accept API breakage.
+module Yesod.Test.Internal
+  ( getBodyTextPreview
+  , contentTypeHeaderIsUtf8
+  , assumedUTF8ContentTypes
```
MaxGabriel, line 10: I thought it was plausible people would want these for any custom assertion functions they'd write. But, they don't feel like something that should be part of the stable API

yesod-test/Yesod/Test.hs
```diff
@@ -569,17 +570,25 @@ assertEqualNoShow :: (HasCallStack, Eq a) => String -> a -> a -> YesodExample si
...
+statusIs number = do
+  withResponse $ \(SResponse status headers body) -> do
+    let mContentType = lookup hContentType headers
+        isUTF8ContentType = maybe False contentTypeHeaderIsUtf8 mContentType
+
+    liftIO $ flip HUnit.assertBool (H.statusCode status == number) $ concat
+      [ "Expected status was ", show number
+      , " but received status was ", show $ H.statusCode status
+      , if isUTF8ContentType
+          then ". For debugging, the body was: " <> (T.unpack $ getBodyTextPreview body)
```
MaxGabriel, line 589: I'm explicitly saying "For debugging" here, to help clarify that the contents of the body are unrelated to the assertion failure

yesod-test/Yesod/Test/Internal.hs
MaxGabriel: But, I'm pretty flexible on this and don't have much experience with releasing Internal modules

yesod-test/test/main.hs
```diff
@@ -125,6 +126,19 @@ main = hspec $ do
...
+            contentTypeHeaderIsUtf8 "application/custom; charset=UTF-8" @?= True
+            contentTypeHeaderIsUtf8 "application/custom; charset=utf-8" @?= True
+        it "matches content-types with an ASCII charset" $ do
+            contentTypeHeaderIsUtf8 "application/custom; charset=us-ascii" @?= True
+        it "matches content-types that we assume are UTF-8" $ do
+            contentTypeHeaderIsUtf8 "text/html" @?= True
+            contentTypeHeaderIsUtf8 "application/json" @?= True
+        it "doesn't match content-type headers that are binary data" $ do
+            contentTypeHeaderIsUtf8 "image/gif" @?= False
+            contentTypeHeaderIsUtf8 "application/pdf" @?= False
```
MaxGabriel, line 140: Technically PDFs can be non-binary Postscript data, but this is fairly rare to my knowledge, and more of a relic

MaxGabriel:
Ready for review!

yesod-test/Yesod/Test.hs
snoyberg: I like that!

yesod-test/yesod-test.cabal
```diff
@@ -1,5 +1,5 @@
 name:               yesod-test
-version:            1.6.9.1
+version:            1.6.9.2
```
snoyberg, line 2: This is my only quibble on the whole thing. It may be better to make this a minor version bump (1.6.10.0) since it has user facing behavioral changes. But I honestly don't feel strongly about this. Feel free to merge/release as you see fit.

MaxGabriel: I'm fine with that, and it also exposes new APIs

MaxGabriel:
Bumped to 1.6.10, should be good to go!

PR #1683, opened 2020-06-29
fixed: yesod-bin: when C-c, kill yesod-bin and children process

ncaq:
* deleted: `setDelegateCtlc True`
* added: `setCreateGroup True`

When you use a group,
the child process will be terminated when the parent process is terminated.

After submitting your PR:


snoyberg:
I'm not sure if and how this will affect Windows, but I think this is worth trying out. Thanks!

PR #1689, opened 2020-07-31
Drop dependency on semigroups

felixonmars:
They are not needed on GHC we support.

### DIFF
--- a/yesod-form/yesod-form.cabal
+++ b/yesod-form/yesod-form.cabal
@@ -32,7 +32,6 @@ library
                    , email-validate        >= 1.0
                    , persistent
                    , resourcet
-                   , semigroups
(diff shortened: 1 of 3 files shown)


yesod-form/yesod-form.cabal
```diff
@@ -45,6 +44,9 @@ library
     if flag(network-uri)
       build-depends: network-uri >= 2.6
 
+    if impl(ghc < 8.0)
```
snoyberg, line 47: Since we already have a `base >= 4.10` dep, we only support this GHC anyway, so I think we can drop this conditional entirely.

felixonmars: Indeed. Updated!

snoyberg:
Thanks!

PR #1691, opened 2020-08-07
Absolute templates path

### LINKED ISSUE #1690: widgetFileReload cannot find "templates" directory when loaded from another dir
When you have a multi package project and one of the packages uses yesod and uses `widgetFileReload` if you `stack ghci` from the project root or similar then it won't compile because a relative path is used which expect you are loading from the yesod using package's root. but in this case we're not

http://hackage.haskell.org/package/yesod-1.6.0.2/docs/src/Yesod.Default.Util.html#globFile

would you be open to using `pathRelativeToCabalPackage` https://hackage.haskell.org/package/th-utilities-0.2.0.1/docs/TH-RelativePaths.html#v:pathRelativeToCabalPackage or something similar in yesod so that these are converted to absolute paths?

dten:

After submitting your PR:

fixes #1690

i'm 99% certain this can't break any usages of `widgetFileReload` or `widgetFileNoReload` because those those `Q Exp` must have already been evaluated in the correct project. unless someone was jumping through a lot of hoops to run ghc manually

### DIFF
--- a/yesod/Yesod/Default/Util.hs
+++ b/yesod/Yesod/Default/Util.hs
@@ -5,6 +5,7 @@
 module Yesod.Default.Util
     ( addStaticContentExternal
     , globFile
+    , globFilePackage
     , widgetFileNoReload
     , widgetFileReload
     , TemplateLanguage (..)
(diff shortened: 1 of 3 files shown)


yesod/Yesod/Default/Util.hs
```diff
@@ -64,6 +66,11 @@ addStaticContentExternal minify hash staticDir toRoute ext' _ content = do
 globFile :: String -> String -> FilePath
 globFile kind x = "templates/" ++ x ++ "." ++ kind
 
+-- | `globFile` but returned path is absolute and within the package the Q Exp is evaluated
+-- @since 1.6.0.3
```
snoyberg, line 70: Since this introduces a new identifier, the version number should be 1.6.1.0

dten:
sorted it :)

PR #1697, opened 2020-09-22
Generate appropriate Handler and Widget synonyms for polymorphic sites

d86leader:
A fix for #1629 

I've looked into #1630 which is an attempt to fix the same issue, and mine does a similar thing, but doesn't break declarations without typevars. Also mine has tests.

The tests show that the code [like in the docs](https://hackage.haskell.org/package/yesod-core-1.6.18/docs/Yesod-Core-Dispatch.html#v:mkYesod) wouldn't compile before my change, but compile as expected with it.
I wasn't sure how to create tests for template haskell, so the tests fail at compile time.

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-core
 
+## 1.6.18.4
+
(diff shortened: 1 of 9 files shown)


snoyberg (approved):
Awesome changeset, thanks!

PR #1710, opened 2020-12-01
Fix up missing HasCallStack

pbrisbin:
Two chanages:

- 24acd4e3 Add missing `HasCallStack`

  Even though functions that *use* this one all have `HasCallStack`, the fact
  that this function itself doesn't means that test failures are reported as
  from this line anyway:

  ```console
  Failures:

    ./Yesod/Test.hs:1571:28:
    1) ...

  ```

  This should correct that.

- 07d76095 Add missing `HasCallStack`s

  As far as I could tell, all of these functions call `failure`, or call things
  that call `failure`.

---

  N/A

  N/A

After submitting your PR:

### DIFF
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-test
 
+## 1.6.11
+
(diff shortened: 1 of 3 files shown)


snoyberg (approved):
Thanks!

PR #1717, opened 2021-02-08
Devel SSL

Burtannia:

After submitting your PR:

This change was made in response to the discussion found [here](https://groups.google.com/g/yesodweb/c/ztCwgWeuErs).

I've tested this on a site I currently have in development and it seems to work in allowing a certificate to be provided for localhost.

Assuming everything else is fine, there are a couple of minor issues / oddities that may need fixing before this is merged. If you access the development site via https then a minute or so later the line `data: end of file` appears in stdout. Also if you issue a CTRL+C command (Ubuntu) there is a pause of a couple of seconds before the above text is printed and then the process exits.

I'm not sure where the `data: end of file` is coming from but my guess is that the delay in the process exiting is to do with a file handle that's not yet been closed due to lazy IO. I tried to work around this by using `hGetContents` and `tlsSettingsMemory` to read the files into strict `ByteString`s rather than passing `FilePath`s to `tlsSettings` (which ultimately uses a lazy `ByteString`). Unfortunately this didn't seem to make any difference so I'm out of ideas. Below is a snippet of what the aforementioned attempt looked like for anyone interested:

```
readStrict fp = withBinaryFile fp ReadMode hGetContents
runProxyTls port app = do
    certOpts <- traverse (bitraverse readStrict readStrict)
                $ bisequence $ (certPath &&& keyPath) opts
    let certDef = $(embedFile "certificate.pem")
        keyDef = $(embedFile "key.pem")
        tlsSettings = uncurry tlsSettingsMemory $ fromMaybe (certDef, keyDef) certOpts
    runTLS tlsSettings (setPort port defaultSettings') $ \req send -> do
```

Another peculiarity is that once you access the site on, for example, `https://localhost:3443`, Chrome then refuses to connect with  `http://localhost:3000` until the cache is cleared. Not sure why this is is the case but it doesn't happen on Firefox so I don't think it's anything on the `yesod devel` end of things.

Any input on the above "issues" would be appreciated.

### DIFF
--- a/yesod-bin/Devel.hs
+++ b/yesod-bin/Devel.hs
@@ -9,6 +9,7 @@ module Devel
     ) where
 
 import           Control.Applicative                   ((<|>))
+import           Control.Arrow                         ((&&&))
 import           UnliftIO                              (race_)
 import           Control.Concurrent                    (threadDelay)
 import           Control.Concurrent.STM
@@ -18,6 +19,7 @@ import           Control.Monad                         (forever, unless, void,
 import Data.ByteString (ByteString, isInfixOf)
 import qualified Data.ByteString.Lazy                  as LB
 import           Conduit
+import           Data.Bitraversable                    (bisequence)
 import           Data.FileEmbed                        (embedFile)
 import qualified Data.Map                              as Map
 import           Data.Maybe                            (isJust)
@@ -56,7 +58,7 @@ import           Network.Wai                           (requestHeaderHost,
                                                         responseLBS)
 import           Network.Wai.Handler.Warp              (defaultSettings, runSettings,
                                                         setPort, setHost)
-import           Network.Wai.Handler.WarpTLS           (runTLS,
+import           Network.Wai.Handler.WarpTLS           (runTLS, tlsSettings,
                                                         tlsSettingsMemory)
 import           Network.Wai.Parse                     (parseHttpAccept)
 import           Say
@@ -126,6 +128,8 @@ data DevelOpts = DevelOpts
       , proxyTimeout    :: Int
       , useReverseProxy :: Bool
       , develHost       :: Maybe String
+      , certPath        :: Maybe FilePath
+      , keyPath         :: Maybe FilePath
       } deriving (Show, Eq)
 
 -- | Run a reverse proxy from the develPort and develTlsPort ports to
@@ -170,10 +174,11 @@ reverseProxy opts appPortVar = do
                 manager
         defaultSettings' = maybe id (setHost . fromString) (develHost opts) defaultSettings
         runProxyTls port app = do
-          let cert = $(embedFile "certificate.pem")
-              key = $(embedFile "key.pem")
-              tlsSettings = tlsSettingsMemory cert key
-          runTLS tlsSettings (setPort port defaultSettings') $ \req send -> do
+          let certDef = $(embedFile "certificate.pem")
+              keyDef = $(embedFile "key.pem")
+              certOpts = bisequence $ (certPath &&& keyPath) opts
+              theSettings = maybe (tlsSettingsMemory certDef keyDef) (uncurry tlsSettings) certOpts
+          runTLS theSettings (setPort port defaultSettings') $ \req send -> do
             let req' = req
                     { requestHeaders
                         = ("X-Forwarded-Proto", "https")
--- a/yesod-bin/Devel.hs
+++ b/yesod-bin/Devel.hs
@@ -9,7 +9,6 @@ module Devel
     ) where
 
 import           Control.Applicative                   ((<|>))
-import           Control.Arrow                         ((&&&))
 import           UnliftIO                              (race_)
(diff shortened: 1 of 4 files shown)


yesod-bin/main.hs
```diff
@@ -30,12 +30,14 @@ data Command = Init [String]
...
              | Touch
              | Devel { develSuccessHook :: Maybe String
-                     , develExtraArgs    :: [String]
+                     , develExtraArgs   :: [String]
                      , develPort        :: Int
                      , develTlsPort     :: Int
                      , proxyTimeout     :: Int
                      , noReverseProxy   :: Bool
                      , develHost        :: Maybe String
+                     , certPath         :: Maybe FilePath
```
snoyberg, line 39: If you combine these into a single field `cert :: Maybe (FilePath, FilePath)`, the opt parser can enforce that either both arguments or neither argument are provided, which will perhaps reduce user error

yesod-bin/Devel.hs
```diff
@@ -170,27 +174,28 @@ reverseProxy opts appPortVar = do
...
-                        : (case lookup "host" (requestHeaders req) of
-                            Nothing ->
-                                case requestHeaderHost req of
-                                    Just host -> (("Host", host):)
-                                    Nothing -> id
-                            Just _ -> id)
-                          (requestHeaders req)
-                    }
-            app req' send
+            let certDef = $(embedFile "certificate.pem")
```
snoyberg, line 177: Can you reformat to match the previous indentation level so that the diff is smaller?

yesod-bin/Devel.hs
```diff
@@ -170,27 +174,28 @@ reverseProxy opts appPortVar = do
...
+                keyDef = $(embedFile "key.pem")
+                certOpts = bisequence $ (certPath &&& keyPath) opts
```
snoyberg, line 179: I'm personally not a fan of this style of code, I'd prefer an explicit `case` here for readability. But I think if you follow my recommendation below the point will be moot.

yesod-bin/yesod-bin.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-bin
-version:         1.6.0.6
+version:         1.6.1
```
snoyberg, line 2: Can you also include a ChangeLog entry?

snoyberg (changes requested):
Overall looks good, thanks! Just a few comments

yesod-bin/Devel.hs
Burtannia: Done! Out of interest, is there a reason why this part is indented in this way? I changed it because it didn't seem to match the rest of the file.

yesod-bin/yesod-bin.cabal
Burtannia: Of course! I planned to do it after the other changes had been finalised but it's done now.

yesod-bin/main.hs
Burtannia: Took me a while to figure this out as I'm not very familiar with the parser library. I couldn't find a function in the library docs which would require both or neither value but I think what I've just implemented is a reasonable solution.

yesod-bin/main.hs
```diff
@@ -167,6 +169,15 @@ develOptions = Devel <$> optStr ( long "success-hook" <> short 's' <> metavar "C
                             <> help "Disable reverse proxy" )
                      <*> optStr (long "host" <> metavar "HOST"
                             <> help "Host interface to bind to; IP address, '*' for all interfaces, '*4' for IP4, '*6' for IP6")
+                     <*> optionPair
```
snoyberg, line 172: Another possibility here:

```haskell
optional ((,) <$> strOption (long "cert" ...) <*> strOption (long "key" ...)
```

snoyberg:
> Done! Out of interest, is there a reason why this part is indented in this way? I changed it because it didn't seem to match the rest of the file.

Probably not, things accrete over time. I don't mind fixing formatting, but I prefer those changes to be in separate PRs.

> Took me a while to figure this out as I'm not very familiar with the parser library. I couldn't find a function in the library docs which would require both or neither value but I think what I've just implemented is a reasonable solution.

I'll include a recommendation in an inline comment, sorry that my initial comment was unclear.

snoyberg (approved):
Thanks!

yesod-bin/main.hs
Burtannia: Change implemented!

snoyberg:
I'll merge/release after CI passes

Burtannia:
@snoyberg Do you have any thoughts about the behaviour I mentioned in the initial post where `data: end of file` is being printed? I've been unable to find where it's coming from or why it "randomly" appears a minute or so into viewing the site via https. I know it's not a massive issue but it just seemed a bit odd.

snoyberg:
Sorry, I didn't notice those comments initially! No, I don't know the details of it, but a quick search shows that it does seem to pop up in warp-tls.

Issue #1720, opened 2021-04-05
Change the order of things checked for languages


pbrisbin:
We've been moving to a more mature system of localization, and as part of that we are setting our users' language in their Session at log-in, so our Frontend client can just trust API responses are correctly localized in the session. This has been really great.

Sometimes we need a particular API response in a specific language for just that request. However, overriding it via the URL doesn't work because of the order things are checked in:

> - The _LANG user session variable.
> - The _LANG get parameter.
> - The _LANG cookie.
> - Accept-Language HTTP header.

[Source](https://www.stackage.org/haddock/lts-17.9/yesod-core-1.6.18.8/Yesod-Core-Handler.html#v:languages).

This doesn't work for our use case, but I think it's objectively surprising and a borderline bug. I would expect properties of _this request_ to be checked ahead of properties of _this session_, not the reverse. I think it would make more sense to "fall back" to something ambient (session, cookie) only if something more explicit (param, header) isn't given.

So, I suggest:

- Accept-Language HTTP header.
- The _LANG get parameter.
- The _LANG cookie.
- The _LANG user session variable.

As the order to check.

I could totally be missing a reason for the current order, and the behavior change is subtle enough that it could really break in-the-wilds apps in hard-to-debug-ways, so I wonder: if I opened a PR to change this, would it be accepted?

snoyberg:
The accept-language header should IMO definitely end up at the end of the list. Basically every browser will send it by default, and if the user selects a different language than what their browser defaults to, we need to respect that.

I think we're both in agreement that the get parameter should trump the cookie, so no issue there.

I'm surprised that you think the _LANG cookie should trump the user session variable, or at the very least that this is something worth worrying about. These seem to be at the same level of "per session"-ness. What's the reason here?

The user session variable vs get parameter (and cookie for that matter) are an interesting case. I think the reason I set it up this way in the first place is that, if someone copy-pastes a URL and sends it to another user, we'd like it to show up in the language that they're expecting. But I can definitely see your point here, and I'd be open to making the get parameter supersede anything else.

In other words, IMO, swapping the current preference between the first two would be the best step forward.

pbrisbin:
[...]

PR #1721, opened 2021-04-08
Stop moving session language ahead of reqLangs

pbrisbin:
`Yesod.Core.Handler.languages` checks first for a language set in the user's
session, prepending that value to `YesodRequest{reqLangs}`, so it is respected
above all else if present.

For context, `reqLangs` itself also includes the session, but just later in
line:

```hs
langs' = catMaybes [ lookup langKey gets -- Query _LANG
                   , lookup langKey cookies     -- Cookie _LANG
                   , lookupText langKey session -- Session _LANG
                   ] ++ langs                    -- Accept-Language(s)
```

In #1720, it was raised that allowing the session (something implicitly present
for any request) to override a query parameter (something explicitly given on
that request) is surprising.

We decided (without knowing what order `reqLangs` was doing) that query, cookie,
session, accept was best and `languages` should be changed to do that.
Conveniently, this just makes `languages` equivalent to `reqLangs`, so that is
what this patch does.

---

I'll do this in a fast-follow:

After submitting your PR:

- [x] Send a message to the mailing list about the order change

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,6 +1,6 @@
 # ChangeLog for yesod-core
 
-## 1.6.18.9
+## 1.6.19.0
 
 * Change order of priority in `languages`[#1721](https://github.com/yesodweb/yesod/pull/1721)
 
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for yesod-core
 
+## 1.6.18.9
+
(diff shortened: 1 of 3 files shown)


snoyberg:
Can you also add a comment in the Haddocks that the behavior of the function changed, and in which version?

pbrisbin:
Before I fire off the yesodweb message, are you good with `1.6.18.9` or do you want to do `1.6.19.0`?

snoyberg:
Let's do 1.6.19.0

pbrisbin:
This is all set, just waiting on the Approve :)

snoyberg:
Thanks!

snoyberg:
And released to Hackage

PR #1722, opened 2021-04-10
Forms: selectFieldGrouped

### LINKED ISSUE #1253: Html tag 'optgroup' in 'selectField'
Hello.

I'd like to add a select field populated from database in which entries are grouped. Currently I use 'BootstrapBasicForm'.

What is the simplest way to do that?

Thanks!

schoettl:
PR: Add `selectFieldGrouped`. Closes #1253 

I'm trying to implement the API [described here](https://github.com/yesodweb/yesod/issues/1253#issuecomment-816447511).

After submitting your PR:

### DIFF
--- a/yesod-form/Yesod/Form/Fields.hs
+++ b/yesod-form/Yesod/Form/Fields.hs
@@ -175,7 +175,7 @@ timeField = timeFieldTypeTime
 -- 
 -- Add the @time@ package and import the "Data.Time.LocalTime" module to use this function.
 --
--- Since 1.4.2
+-- @since 1.4.2
 timeFieldTypeTime :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m TimeOfDay  
 timeFieldTypeTime = timeFieldOfType "time"
 
@@ -185,7 +185,7 @@ timeFieldTypeTime = timeFieldOfType "time"
 -- 
 -- Add the @time@ package and import the "Data.Time.LocalTime" module to use this function.
 --
--- Since 1.4.2
+-- @since 1.4.2
 timeFieldTypeText :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m TimeOfDay
 timeFieldTypeText = timeFieldOfType "text"
 
@@ -365,7 +365,7 @@ $newline never
 
 -- | Creates an input with @type="email"@ with the <http://w3c.github.io/html/sec-forms.html#the-multiple-attribute multiple> attribute; browsers might implement this as taking a comma separated list of emails. Each email address is validated as described in 'emailField'.
 --
--- Since 1.3.7
+-- @since 1.3.7
 multiEmailField :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m [Text]
 multiEmailField = Field
     { fieldParse = parseHelper $
@@ -431,6 +431,8 @@ selectFieldList :: (Eq a, RenderMessage site FormMessage, RenderMessage site msg
 selectFieldList = selectField . optionsPairs
 
 -- | Creates a @\<select>@ tag with @\<optgroup>@s for selecting one option.
+--
+-- @since 1.7.0
 selectFieldListGrouped :: (Eq a, RenderMessage site FormMessage, RenderMessage site msg)
                 => [(msg, [(msg, a)])]
                 -> Field (HandlerFor site) a
@@ -611,6 +613,8 @@ $newline never
         showVal = either (\_ -> False)
 
 -- | A structure holding a list of options. Typically you can use a convenience function like 'mkOptionList' or 'optionsPairs' instead of creating this directly.
+--
+-- Extended by 'OptionListGrouped' in 1.7.0.
 data OptionList a
   = OptionList
     { olOptions :: [Option a]
@@ -622,11 +626,13 @@ data OptionList a
     }
 
 -- | Convert grouped 'OptionList' to a normal one.
+--
+-- @since 1.7.0
 flattenOptionList :: OptionList a -> OptionList a
 flattenOptionList (OptionListGrouped os re) = OptionList (concatMap snd os) re
 flattenOptionList ol = ol
 
--- | Since 1.4.6
+-- | @since 1.4.6
 instance Functor OptionList where
     fmap f (OptionList options readExternal) =
       OptionList ((fmap.fmap) f options) (fmap f . readExternal)
@@ -641,6 +647,8 @@ mkOptionList os = OptionList
     }
 
 -- | Creates an 'OptionList', using a 'Map' to implement the 'olReadExternalGrouped' function.
+--
+-- @since 1.7.0
 mkOptionListGrouped :: [(Text, [Option a])] -> OptionList a
 mkOptionListGrouped os = OptionListGrouped
     { olOptionsGrouped = os
@@ -653,7 +661,7 @@ data Option a = Option
     , optionExternalValue :: Text -- ^ The representation of this value stored in the form.
     }
 
--- | Since 1.4.6
+-- | @since 1.4.6
 instance Functor Option where
     fmap f (Option display internal external) = Option display (f internal) external
 
@@ -670,6 +678,8 @@ optionsPairs opts = do
   return $ mkOptionList (zipWith mkOption [1 :: Int ..] opts)
 
 -- | Creates an 'OptionList' from a list of (display-value, internal value) pairs.
+--
+-- @since 1.7.0
 optionsPairsGrouped
   :: (MonadHandler m, RenderMessage (HandlerSite m) msg)
   => [(msg, [(msg, a)])] -> m (OptionList a)
@@ -746,7 +756,7 @@ optionsPersist filts ords toDisplay = fmap mkOptionList $ do
 -- | An alternative to 'optionsPersist' which returns just the 'Key' instead of
 -- the entire 'Entity'.
 --
--- Since 1.3.2
+-- @since 1.3.2
 #if MIN_VERSION_persistent(2,5,0)
 optionsPersistKey
   :: (YesodPersist site
--- a/yesod-form/Yesod/Form/Fields.hs
+++ b/yesod-form/Yesod/Form/Fields.hs
@@ -80,7 +80,7 @@ import Database.Persist (Entity (..), SqlType (SqlString), PersistRecordBackend,
 import Database.Persist (Entity (..), SqlType (SqlString), PersistEntity, PersistQuery, PersistEntityBackend)
 #endif
 import Text.HTML.SanitizeXSS (sanitizeBalance)
-import Control.Monad (when, unless)
+import Control.Monad (when, unless, forM_)
 import Data.Either (partitionEithers)
 import Data.Maybe (listToMaybe, fromMaybe)
 
@@ -749,7 +749,7 @@ selectFieldHelper outside onOpt inside opts' = Field
         opts <- fmap olOptions $ handlerToWidget opts'
         outside theId name attrs $ do
             unless isReq $ onOpt theId name $ not $ render opts val `elem` map optionExternalValue opts
-            flip mapM_ opts $ \opt -> inside
+            forM_ opts $ \opt -> inside
                 theId
                 name
                 ((if isReq then (("required", "required"):) else id) attrs)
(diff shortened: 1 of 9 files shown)


yesod-form/Yesod/Form/Fields.hs
```diff
@@ -768,6 +835,22 @@ selectFieldHelper outside onOpt inside opts' = Field
...
+-- A helper function for constucting 'selectField's. You may want to use this when you define your custom 'selectField's or 'radioField's.
+selectFieldHelperGrouped
+        :: (Eq a, RenderMessage site FormMessage)
+        => (Text -> Text -> [(Text, Text)] -> WidgetFor site () -> WidgetFor site ()) -- ^ Outermost part of the field
+        -> (Text -> Text -> Bool -> WidgetFor site ()) -- ^ An option for None if the field is optional
+        -> (Text -> Text -> [(Text, Text)] -> Text -> Bool -> Text -> WidgetFor site ()) -- ^ Other options
+        -> (Text -> WidgetFor site ()) -- ^ Group headers placed inbetween options
+        -> HandlerFor site (OptionList a)
+        -> Field (HandlerFor site) a
+selectFieldHelperGrouped outside onOpt inside group opts' =
```
schoettl, line 848: @snoyberg can you give me a hint here?

When I pass `optsFlat` instead of `opts'` as last parameter (both have type `OptionList`), I get this error:

```
/home/jakob/projects/yesod/yesod-form/Yesod/Form/Fields.hs:849:44: error:
    • Couldn't match expected type ‘HandlerFor site (OptionList a)’
                  with actual type ‘Field m0 (OptionList a)’
    • In the fourth argument of ‘selectFieldHelper’, namely
        ‘optsFlat
```
I don't know how to fmap the `OptionList` while "keeping the surrounding monad"...

schoettl:
I'm thinking about if it's a breaking change to add a constructor to the `OptionList` type as I did in this PR. If users do pattern matching or use the field accessors they will get compiler warnings. Is that OK?

On the other hand, not reusing and adopting the `OptionList` would lead to more code and code duplication, I think.

snoyberg:
Let's consider it a major version bump. Note that CI is breaking, I'll give a more thorough review when it's fixed.

schoettl:
I'd love to fix it but I'm not capable yet :/

Maybe you could give me a hint on how I can extract the `OptionList` from a Handler monad in a function returning `Field (HandlerFor site) a` (see comment in `selectField` function). That's my basic problem, from there I can probably continue.

I have a basic understanding of monads but I'm not grasping the meaning of [`Field`](https://hackage.haskell.org/package/yesod-form-1.6.7/docs/Yesod-Form-Types.html#t:Field) having the Handler monad as constructor parameter... is `Field` another monad then or is it still the Handler monad?

snoyberg:
It looks like you already pushed a new commit that fixed CI, are you still looking for feedback on how to fix a compile error?

schoettl:
Hi Michael, right, I'm still looking for help. I fixed CI only by commenting out the problems... I'll mark you in two problem lines of the diff, OK?

yesod-form/Yesod/Form/Fields.hs
```diff
@@ -427,25 +428,39 @@ selectFieldList :: (Eq a, RenderMessage site FormMessage, RenderMessage site msg
...
 -- | Creates a @\<select>@ tag for selecting one option. Example usage:
 --
 -- > areq (selectField $ optionsPairs [(MsgValue1, "value1"),(MsgValue2, "value2")]) "Which value?" Nothing
 selectField :: (Eq a, RenderMessage site FormMessage)
             => HandlerFor site (OptionList a)
             -> Field (HandlerFor site) a
-selectField = selectFieldHelper
-    (\theId name attrs inside -> [whamlet|
+selectField ol = do
+  -- ol' <- ol -- TODO how to extract the option list for pattern matching?
```
schoettl, line 444: @snoyberg I need help

yesod-form/Yesod/Form/Fields.hs
```diff
@@ -768,6 +822,26 @@ selectFieldHelper outside onOpt inside opts' = Field
...
+        => (Text -> Text -> [(Text, Text)] -> WidgetFor site () -> WidgetFor site ()) -- ^ Outermost part of the field
+        -> (Text -> Text -> Bool -> WidgetFor site ()) -- ^ An option for None if the field is optional
+        -> (Text -> Text -> [(Text, Text)] -> Text -> Bool -> Text -> WidgetFor site ()) -- ^ Other options
+        -> (Text -> WidgetFor site ()) -- ^ Group headers placed inbetween options
+        -> HandlerFor site (OptionList a)
+        -> Field (HandlerFor site) a
+selectFieldHelperGrouped outside onOpt inside group opts' = do
+    let optsFlat = fmap flattenOptionList opts'
+    selectFieldHelper outside onOpt inside optsFlat -- {
+    --   fieldView = \theId name attrs val isReq -> do
```
schoettl, line 838: @snoyberg Here I'm trying change only part of the `Field` (i.e. `fieldView`). It should be possible right?

yesod-form/Yesod/Form/Fields.hs
snoyberg: Please look at the implementation of `selectFieldHelper`. It looks like you'll need to modify things in there. Overall it seems like the approach you're trying to take is incompatible with how these functions are structured.

schoettl: ... because `selectField` is a pure function and I can pass but not modify monadic values... I think I understand now.

schoettl:
Not yet tested because using yesod-form in a real project fails...

stack.yml:
```
packages:
- .
- ../nn2gs
- ../yesod/yesod-core
- ../yesod/yesod-auth
- ../yesod/yesod-static
- ../yesod/yesod-form
```

Building the real project fails with:
```
* * * * * * * *
Warning: There are cabal settings for this project which may prevent GHCi from loading your code properly.
In some cases it can also load some projects which would otherwise fail to build.

-XNoImplicitPrelude will be used, but GHCi will likely fail to build things which depend on the implicit prelude.
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

-XCPP will be used, but it can cause issues with multiline strings.
See https://downloads.haskell.org/~ghc/7.10.2/docs/html/users_guide/options-phases.html#cpp-string-gaps
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

-XTemplateHaskell will be used, but it may cause compilation issues due to different parsing of '$' when there's no space after it.
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

-XQuasiQuotes will be used, but it may cause parse failures due to a different meaning for list comprehension syntax like [x| ... ]
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

-XOverloadedStrings will be used, but it can cause type ambiguity in code not usually compiled with it.
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

-XTypeFamilies will be used, but it implies -XMonoLocalBinds, and so can cause type errors in code which expects generalized local bindings.
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

-XGADTs will be used, but it implies -XMonoLocalBinds, and so can cause type errors in code which expects generalized local bindings.
It is specified for:
    nn2gs-backend:lib nn2gs-backend:exe:nn2gs-backend nn2gs-backend:exe:nn2gs-license-manager
But not for: 
    nn2gs:lib nn2gs:exe:nn2gs-exe yesod-auth:lib yesod-core:lib yesod-form:lib yesod-static:lib

To resolve, remove the flag(s) from the cabal file(s) and instead put them at the top of the haskell files.

It isn't yet possible to load multiple packages into GHCi in all cases - see
https://ghc.haskell.org/trac/ghc/ticket/10827
* * * * * * * *

The following GHC options are incompatible with GHCi and have not been passed to it: -threaded -O2
Configuring GHCi with the following packages: nn2gs, nn2gs-backend, yesod-auth, yesod-core, yesod-form, yesod-static
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help

/home/jakob/projects/yesod/yesod-auth/Yesod/Auth/Routes.hs:19:0: error:
     Fehler: nicht beendeter Kommentar
       19 | /page/#Text/*Texts     PluginR
          | 
   |
19 | /page/#Text/*Texts     PluginR
   | ^
`gcc' failed in phase `C pre-processor'. (Exit code: 1)
```

yesod-form/Yesod/Form/Fields.hs
```diff
@@ -637,6 +668,28 @@ optionsPairs opts = do
...
+  :: (MonadHandler m, RenderMessage (HandlerSite m) msg)
+  => [(msg, [(msg, a)])] -> m (OptionList a)
+optionsPairsGrouped opts = do
+  mr <- getMessageRender
+  let mkOption (external, (display, internal)) =
+          Option { optionDisplay       = mr display
+                 , optionInternalValue = internal
+                 , optionExternalValue = pack $ show external
+                 }
+      opts' = enumerateSublists opts -- :: [(msg, [(Int, (msg, a))])]
```
schoettl, line 683: Keep this comment? It helps to understand whats going on (as long it's not outdated). Much better would be a type annotation but `msg` and `a` seem to be not available here.

Same thing below at `yss`.

schoettl (commented):
I can also add the `@since` comments if you tell me what the next (major) version should be. 1.7.0?

snoyberg:
Yes, it should be 1.7.0.

snoyberg:
Also: please update the changelog to note the additions and breaking changes.

yesod-form/Yesod/Form/Fields.hs
snoyberg: If you use `ScopedTypeVariables` and an explicit `forall`, you can add in a type signature I believe. That would be preferable.

schoettl:
Thank you for the feedback. I've updated changelog and versions (also in cabal file). How about the other `yesod-*` packages? They stay at 1.6.* I guess?

snoyberg:
The other versions should stay at their current version, though you'll need to relax some upper bounds (see the CI failures).

schoettl:
> The other versions should stay at their current version, though you'll need to relax some upper bounds (see the CI failures).

Alright, done.

snoyberg (approved):
Thank you!

schoettl:
Great, thank you, too! Do you know when the new version will show up at hackage/stackage? Or is it just a matter of hours?

snoyberg:
More like minutes, I just finished uploading. It would have been faster, but Hackage has introduced some new requirements on minimum cabal versions.

schoettl:
Nice, now it works.

Though I had to add yesod, yesod-auth, and yesod-form with specific versions to the `extra-deps` list in stack.yaml. `stack build` fails otherwise because the stack resolver seems not be able to use the yesod-form-1.7.0 together with the other yesod-* packages.

PR #1723, opened 2021-05-05
Support persistent-2.13

parsonsmatt:

After submitting your PR:

### DIFF
--- a/stack-persistent-213.yaml
+++ b/stack-persistent-213.yaml
@@ -16,10 +16,7 @@ packages:
 - ./yesod-websockets
 extra-deps:
 - lift-type-0.1.0.1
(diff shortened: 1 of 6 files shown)


stack-persistent-213.yaml
```diff
@@ -0,0 +1,22 @@
+resolver: nightly-2021-03-31
```
snoyberg, line 1: Can you add this config to the CI build matrix here? https://github.com/yesodweb/yesod/blob/master/.github/workflows/tests.yml#L24

snoyberg:
It looks like this isn't yet released to Hackage, right?

parsonsmatt:
The most recent commit is using the Hackage releases in the `extra-deps`.

Issue #1725, opened 2021-05-09
Is it safe to use System.Random to generate session tokens?


Smart-Hypercube:
https://github.com/yesodweb/yesod/blob/5d8566ad5c35b3225f6b2b14c57854cb678abffc/yesod-core/src/Yesod/Core/Dispatch.hs#L96

https://github.com/yesodweb/yesod/blob/5d8566ad5c35b3225f6b2b14c57854cb678abffc/yesod-core/src/Yesod/Core/Internal/Request.hs#L132

Yesod currently uses System.Random as the default random number generator, and uses this to generate session tokens. This looks very dangerous to me, because System.Random is only a pseudo-random number generator and has not been designed with security in mind. An attacker may easily collect enough information about the internal state of the generator and predict all future outputs.

I suggest using [System.Entropy](https://www.stackage.org/haddock/nightly-2021-05-09/entropy-0.4.1.6/System-Entropy.html).

snoyberg:
That's probably a good call. Are you interested in sending such a pull request?

Smart-Hypercube:
Oh, I just realized this token is not session token. Yesod core does not have the "session token" concept. This is CSRF token and only used by `defaultCsrfMiddleware`...
Given that this middleware is optional, isn't it strange to include the CSRF token in `YesodRequest`? A better design may be storing CSRF tokens in session data. The token generation logic can also be moved into `setCsrfCookieWithCookie` where it belongs.
CSRF tokens should still be securely generated though. It is user-facing and not protected by encryption.

snoyberg:
I don't really remember the original design around CSRF tokens, and I'm not particularly interested in changing the way it works right now unless there's a concrete reason. Improving how they're generated for proper protection does still seem worthwhile though.

PR #1726, opened 2021-05-11
Use secure entropy source to generate CSRF tokens

Smart-Hypercube:
Related issue: #1725 

---

After submitting your PR:

### DIFF
--- a/yesod-core/yesod-core.cabal
+++ b/yesod-core/yesod-core.cabal
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.19.0
+version:         1.7.0
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>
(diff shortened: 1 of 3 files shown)


yesod-core/yesod-core.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.19.0
+version:         1.7.0
```
snoyberg, line 2: Why is this a major version bump?

snoyberg (commented):
Code looks OK to me, but I'm concerned about the major version bump.

Smart-Hypercube:
I'm not familiar with this version style. According to https://github.com/yesodweb/yesod/blob/master/CONTRIBUTING.md#versioning , I changed behavior of existing public API (`yreGen` is public), so I bumped B part. What version number do you think it should be? I'll change it.

snoyberg:
I wouldn't consider this a semantic breaking change, but rather a bug fix. Either C or D would be appropriate here. I'd stick to a C bump, and would go for version 1.6.20. Thanks!

PR #1728, opened 2021-05-17
yesod-core: detect loops in breadcrumbs

NorfairKing:
This is a proof of concept for a loop detector in the breadcrumbs.
I'm not entirely happy with this because it is quadratic time instead of the linear time which it could be with a hashset or so.
However: this _should_ not matter because you will probably have a very small number of breadcrumbs.
So I'm not sure, what do you think @snoyberg ?

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Class/Breadcrumbs.hs
+++ b/yesod-core/src/Yesod/Core/Class/Breadcrumbs.hs
@@ -29,8 +29,8 @@ breadcrumbs = do
     go back Nothing = return back
     go back (Just this) = do
         (title, next) <- breadcrumb this
-        if next `elem` (map (Just . fst) back)
+        if this `elem` map fst back
         then
-          error $ "infinite recursion in breadcrumbs at" <> show title
+          error $ "yesod-core: infinite recursion in breadcrumbs at " <> show title
         else
           go ((this, title) : back) next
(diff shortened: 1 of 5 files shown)


snoyberg:
This is technically a breaking change due to the addition of `Eq`, but given that the TH automatically adds that I don't think it's a real problem. I'm OK with the quadratic approach here for the reasons you gave, though I'd probably refactor slighting to use `elem` on the `this` instead of `next` to avoid the `Just` wrapping.

NorfairKing:
@snoyberg I think I did the refactor that you had in mind. I also added a test to make sure that we do indeed get a 500 response.

snoyberg:
Build is failing for LTS 11.

yesod-core/src/Yesod/Core/Class/Breadcrumbs.hs
```diff
@@ -28,4 +29,8 @@ breadcrumbs = do
     go back Nothing = return back
     go back (Just this) = do
         (title, next) <- breadcrumb this
-        go ((this, title) : back) next
+        if this `elem` map fst back
```
snoyberg, line 32: I think it would make more sense to perform this check _before_ calling `breadcrumb`. There's a minor performance improvement (not something I care about here), but IMO the code is easier to read. Actually, I'd probably go for a guard here:

```haskell
go back (Just this)
  | this `elem` map fst back = error ...
  | otherwise = do ...
```

NorfairKing: Oh, doh! This makes a lot of sense.
I tried this out in the latest commit. It does require another constraint but I'm pretty sure that a `Show` instance is also generated.

snoyberg: I'll merge with that constraint, and if anyone complains we can remove it later. Thanks!

snoyberg (changes requested):
Doh, I just noticed one more thing. Can you please include a minor version bump and changelog entry about the modification?

snoyberg (approved):

PR #1745, opened 2022-01-19
Add missing documentation to 'warp'

stevehartdata:
This adds a list that was missing in the documentation for ``Yesod.Core.Dispatch.warp``. As this is only a documentation change, the version number does not need to be bumped.


snoyberg (approved):
Thanks!

Issue #1747, opened 2022-02-19
Add "colorField" for <input type="color">


ciukstar:
At the moment I see no `Field` record for html `<input type="color">` in the module `Yesod.Form.Fields`.
Please add `colorField` for `<input type="color">`.
Thanks.

jezen:
@ciukstar Is there anything stopping you from sending in a pull request? If you'd like to give that a go, I'd be happy to review it. I haven't ever used an input of this type, so I haven't felt the lack of it yet, but I'd be interested to see what you come up with 🙂

PR #1748, opened 2022-02-20
Color field

ciukstar:
Reference issue #1747

This PR adds a new `colorField` function to create an html color field (`<input type="color">`) as described at [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color)
<hr>

After submitting your PR:

### DIFF
--- a/yesod-form/Yesod/Form/Fields.hs
+++ b/yesod-form/Yesod/Form/Fields.hs
@@ -954,6 +954,8 @@ prependZero t0 = if T.null t1
 
 -- | Creates an input with @type="color"@.
 --   The input value must be provided in hexadecimal format #rrggbb.
+--
+-- @since 1.7.1 
 colorField :: Monad m => RenderMessage (HandlerSite m) FormMessage => Field m Text
 colorField = Field
     { fieldParse = parseHelper $ \s ->
--- a/yesod-form/Yesod/Form/Fields.hs
+++ b/yesod-form/Yesod/Form/Fields.hs
@@ -62,6 +62,7 @@ module Yesod.Form.Fields
     , optionsPairs
     , optionsPairsGrouped
     , optionsEnum
+    , colorField
     ) where
 
 import Yesod.Form.Types
@@ -116,6 +117,8 @@ import Data.String (IsString)
 #if !MIN_VERSION_base(4,8,0)
 import Data.Monoid
 #endif
+
+import Data.Char (isHexDigit)
 
 defaultFormMessage :: FormMessage -> Text
 defaultFormMessage = englishFormMessage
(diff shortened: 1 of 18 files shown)


yesod-form/Yesod/Form/Fields.hs
```diff
@@ -948,3 +951,23 @@ prependZero t0 = if T.null t1
...
+    , fieldView = \theId name attrs val _ -> [whamlet|
+$newline never
+<input ##{theId} name=#{name} *{attrs} type=color value=#{either id id val}>
+|]
+    , fieldEnctype = UrlEncoded
+    }
+  where
+      isHexColor :: String -> Bool
+      isHexColor ['#',a,b,c,d,e,f] = all isHexDigit [a,b,c,d,e,f]
+      isHexColor _ = False
```
schoettl, line 973: Hex numbers in CSS can also be `#fff` (3 letters). Are you sure that all browsers ever only set 6 letter hex digits?

ciukstar: Quoting from [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color): 

> "The value must be in seven-character hexadecimal notation, meaning the "#" character followed by two digits each representing red, green, and blue, like this: #rrggbb".

It is supposed to cover only HTML, not CSS.
Please let me know if I have to look in other non-HTML specs such as CSS. I'm not sure of this.

schoettl: Ah, I guess you're right! BTW, this would be the most official spec for HTML input: https://html.spec.whatwg.org/multipage/input.html#color-state-(type=color)
… the "bookkeeping details" are also interesting.

jezen:
@ciukstar Thanks for the PR! Looks good to me 👍 I've started running the build on CI for this, and hopefully we can get this over the line for you soon.

ciukstar:
Hi! I see 14 successful and 4 failed checks. Do I need to do something to make it pass all the checks?

jezen:
@ciukstar Sorry for the slow response, I've been busy escaping the war in Ukraine.

I'm not sure why those MacOS builds are failing. Perhaps @snoyberg might know?

snoyberg:
I have no idea why it would have stopped working recently, I'm assuming that it's to do with GitHub Actions upgrading MacOS images. I see this error message:

```
yesod-bin                        > <command line>: can't load framework: Cocoa (not found)
```

I'd recommend ignoring the failure for now.

PR #1750, opened 2022-03-10
Add before hooks and a function instance to Example

parsonsmatt:

After submitting your PR:

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -117,12 +117,19 @@ module Yesod.Test
     , yesodSpecApp
     , YesodExample
     , YesodExampleData(..)
+    , yesodExampleDataFromApp
+    , yesodExampleDataFromTestApp
     , TestApp
     , YSpec
     , testApp
+    , minimalTestApp
     , YesodSpecTree (..)
     , ydescribe
     , yit
+
+    -- * Hooks
+    , beforeApp
+    , beforeWithApp
 
     -- * Modify test site
     , testModifySite
@@ -220,9 +227,11 @@ module Yesod.Test
     , htmlQuery
     , parseHTML
     , withResponse
+    -- * YesodExa
     ) where
 
 import qualified Test.Hspec.Core.Spec as Hspec
+import qualified Test.Hspec.Core.Hooks as Hspec
 import qualified Data.List as DL
 import qualified Data.ByteString.Char8 as BS8
 import Data.ByteString (ByteString)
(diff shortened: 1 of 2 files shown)


yesod-test/Yesod/Test.hs
```diff
@@ -1576,27 +1598,98 @@ failure :: (HasCallStack, MonadIO a) => T.Text -> a b
...
+        (TestApp site, r)
+
+    evaluateExample example params action =
+        Hspec.evaluateExample
+            (action $ \(testApp, r) -> do
+                yed <- yesodExampleDataFromTestApp testApp
+                _ <- evalSIO (example r) yed
+                return ())
+            params
+            ($ ())
```
parsonsmatt, line 1636: This lets us use `beforeApp` and provide the value created as a lambda.

```haskell
beforeApp (pure "asdf") $ do
    it "has asdf" $ \asdf -> do
        pure () :: YesodExample site ()
        liftIO $ asdf `shouldBe` "asdf"
```

yesod-test/Yesod/Test.hs
```diff
@@ -397,13 +424,8 @@ yesodSpecWithSiteGeneratorAndArgument getSiteAction yspecs =
...
         site <- getSiteAction' a
-        app <- toWaiAppPlain site
-        evalSIO y YesodExampleData
-            { yedApp = app
-            , yedSite = site
-            , yedCookies = M.empty
-            , yedResponse = Nothing
-            }
+        yed <- yesodExampleDataFromApp site
+        evalSIO y yed
```
parsonsmatt, line 428: So, a problem with the current implementation (as surfaced in the pending test), is that we can't persist state from the `YesodExampleData` that is generated. We're doing `evalSIO y yed` and throwing away the modified state.

This means our `beforeApp` and `beforeWithApp` hooks can modify the database and return database entities, but those entities (and any changes made to the app, like sessions or logins) won't be kept.

A proper fix would probably thread the `YesodExampleData` through the whole thing, and then only let it get cleared out in the `Example` instance.

yesod-test/test/main.hs
```diff
@@ -502,6 +502,23 @@ main = hspec $ do
...
+        before (pure $ minimalTestApp app) $ do
+            it "can run regular tests" $ do
+                get ("get-json-response" :: Text)
+                statusIs 200
+                xs <- requireJSONResponse
+                assertEq "The value is [1]" xs [1 :: Integer]
+            describe "Root Route" $ do
+                beforeApp (get ("/" :: Text)) $ do
+                    it "can do stuff" $ \() -> do
+                        liftIO $ pendingWith "This test currently fails because the SIO type can't share the state. When we do an `evalSIO` in `beforeApp`, that throws away all the changes made, so requests are not persisted."
```
parsonsmatt, line 515: sad face

parsonsmatt (commented):
While this code is useful, it's not as useful as I'd like it to be. Not sure if we want to go forward with this implementation or do a bigger refactoring.

yesod-test/Yesod/Test.hs
snoyberg: I haven't looked in depth yet, but would using a mutable variable be a solution here?

parsonsmatt: The `SIO` stuff is already a mutable variable. We're just not threading it through and re-using it afterwards.

PR #1752, opened 2022-03-22
Create instances for ToContent Void, ToTypedContent Void

lykahb:
Some handlers never respond with the success response. They always `redirect`, `sendFile` or return some other error.  Giving them the type `Handler ()` implies that they may return an empty success response. The type `Handler Void` is more precise and helps to understand the handler behavior at a glance.

---

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+## 1.6.22.0
+
 * Add missing list to documentation for ``Yesod.Core.Dispatch.warp``. [#1745](https://github.com/yesodweb/yesod/pull/1745)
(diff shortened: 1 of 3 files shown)


snoyberg (approved):
LGTM! Just out of curiosity, what’s the use case where this came up?

lykahb:
There were two use cases where the specific type was used only to make the handler content type monomorphic:
- a handler with `sendResponse ("application/pdf" :: ContentType, content)`. I replaced this exception-based response with `Handler (ContentType, Content)`.
- a handler that sends `sendStatusJSON Status.status202 payload` on its happy path. For that one the `Handler Void` is the most fitting type

snoyberg:
Got it thanks!

PR #1754, opened 2022-03-22
Allow building with GHC 9.2

TeofilC:

After submitting your PR:

I haven't added a new CI test for 9.2 as one of the dependencies (hjsmin) is missing from stackage nightly

### DIFF
--- a/yesod-bin/AddHandler.hs
+++ b/yesod-bin/AddHandler.hs
@@ -16,6 +16,9 @@ import Distribution.PackageDescription.Parse (readGenericPackageDescription)
 #else
 import Distribution.PackageDescription.Parse (readPackageDescription)
 #endif
+#if MIN_VERSION_Cabal(3, 6, 0)
+import Distribution.Utils.Path
+#endif
 import Distribution.PackageDescription.Configuration (flattenPackageDescription)
 import Distribution.PackageDescription (allBuildInfo, hsSourceDirs)
 import Distribution.Verbosity (normal)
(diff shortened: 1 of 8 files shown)


yesod-bin/AddHandler.hs
```diff
@@ -247,4 +250,8 @@ getSrcDir cabal = do
 #endif
     let buildInfo = allBuildInfo pd
         srcDirs = concatMap hsSourceDirs buildInfo
-    return $ fromMaybe "." $ listToMaybe srcDirs
+#if MIN_VERSION_Cabal(3, 6, 0)
+    return $ maybe "." getSymbolicPath $ listToMaybe srcDirs
+#else
+    return $ fromMaybe "." $ listToMaybe srcDirsFilePath
```
schoettl, line 256: I'm not a regular haskell code reviewer, so sorry if I'm wrong. But where does `srcDirsFilePath` come from? I can't find the definition in this file.

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -182,3 +182,10 @@ notStrict = Bang NoSourceUnpackedness NoSourceStrictness
 
 instanceD :: Cxt -> Type -> [Dec] -> Dec
 instanceD = InstanceD Nothing
+
+conPCompat :: Name -> [Pat] -> Pat
+conPCompat n pats = ConP n
+#if MIN_VERSION_template_haskell(2,18,0)
+                         []
+#endif
+                         pats
```
schoettl, line 191: Isn't here an `#else` missing? Or does `ConP` actually have a different number of args in different haskell versions?

yesod-bin/AddHandler.hs
TeofilC: Good spot! That was a typo the "FilePath" shouldn't be there. I've force pushed a fix

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
TeofilC: Yes that's right. There's a new argument in template-haskell 2.18

snoyberg (approved):

PR #1756, opened 2022-03-25
Remove sometimes failing superfluous  test

jappeace:
This test sometimes fails on nix builds.
I'm not sure why, but it should be superfluous  with
"thread killed = 500" test anyway.

They test both for async exceptions.
Just a different one.

After submitting your PR:


jappeace:
this was introduced in https://github.com/yesodweb/yesod/pull/1753

snoyberg (approved):

PR #1762, opened 2022-04-13
Support `annotated-exception`

### LINKED ISSUE #1761: Support `annotated-exception`
`annotated-exception` wraps exceptions in a semipermeable `AnnotatedException e` type, such that catching `AnnotatedException e` will also catch a thrown `e` (and provide a `mempty` for the annotations). So,

```haskell
catchAnnotated :: Exception e => IO a -> (AnnotatedException e -> IO a) -> IO a
catchAnnotated = Control.Exception.catch
```

will never let an `e` escape.

However, only the `catch` and `try` functions in `annotated-exception` will also "see through" the `AnnotatedException e` and allow you to catch a plain `e`.

```haskell
catchUnAnnotated :: Exception e => IO a (e -> IO a) -> IO a
catchUnAnnotated = Control.Exception.Annotated.catch
```

`yesod` currently uses `HandlerContents` as an exception to short-circuit execution of a handler. If an application is using `AnnotatedException`, then the current exception catching facilities will fail to handle an `AnnotatedException HandlerContents` *as* a `HandlerContents`, and will instead fail.

You can work around this by defining an exception handling middleware:

```haskell
removeAnnotatedException :: HandlerFor site a -> HandlerFor site a
removeAnnotatedException action =
    action `catch` \(AnnotatedException _ (e :: SomeException)) ->
        throwIO e
```

But this would "just work" if `yesod` were integrated with `annotated-exception`, if *only* to catch the `HandlerContents`.

parsonsmatt:
Fixes #1761 

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Core/Internal/Run.hs
+++ b/yesod-core/src/Yesod/Core/Internal/Run.hs
@@ -25,6 +25,7 @@ import qualified Control.Exception as EUnsafe
 import Yesod.Core.Internal.Response
 import           Data.ByteString.Builder      (toLazyByteString)
 import qualified Data.ByteString.Lazy         as BL
+import           Control.Exception.Annotated  (AnnotatedException (..))
 import           Control.Monad.IO.Class       (MonadIO, liftIO)
 import           Control.Monad.Logger         (LogLevel (LevelError), LogSource,
                                                liftLoc)
@@ -116,7 +117,7 @@ basicRunHandler rhe handler yreq resState = do
             return (HCContent defaultStatus tc))
         (\e ->
             case fromException e of
-                Just e' -> return e'
+                Just (AnnotatedException _ e') -> return e'
                 Nothing -> HCError <$> toErrorHandler e)
 
     -- Get the raw state and return
--- a/yesod-core/yesod-core.cabal
+++ b/yesod-core/yesod-core.cabal
@@ -27,6 +27,7 @@ library
 
     build-depends:   base                  >= 4.10     && < 5
                    , aeson                 >= 1.0
+                   , annotated-exception   >= 0.1.0.0 
                    , auto-update
                    , blaze-html            >= 0.5
                    , blaze-markup          >= 0.7.1


parsonsmatt:
I've made a PR to add `annotated-exception` to STackage which is probably a prerequisite: https://github.com/commercialhaskell/stackage/pull/6540

yesod-core/src/Yesod/Core/Internal/Run.hs
```diff
@@ -116,7 +117,7 @@ basicRunHandler rhe handler yreq resState = do
             return (HCContent defaultStatus tc))
         (\e ->
             case fromException e of
-                Just e' -> return e'
+                Just (AnnotatedException _ e') -> return e'
```
parsonsmatt, line 120: This deserves a bit of commentary.

The [`instance Exception e => Exception (AnnotatedException e)`](https://hackage.haskell.org/package/annotated-exception-0.2.0.0/docs/Control-Exception-Annotated.html#t:AnnotatedException) has a slightly ~magical~ ~terrible~ ~horrifying~ unusual implementation of `fromException`.

The [tests for this demonstrate the pattern](https://github.com/parsonsmatt/annotated-exception/blob/master/test/Control/Exception/AnnotatedSpec.hs#L41-L65), but tl;dr, the `Exception` instance here on `AnnotatedException` will work to catch both an `AnnotatedException HandlerContents` *and* a `HandlerContents`, which is then promoted to `AnnotatedException mempty (hc :: HandlerContents)`.

schoettl:
Hi Matt,
Maybe it makes sense to clarify first, if this library can make it into Yesod as a dependency at the moment. Before you put too much work in this PR.
Don't get me wrong but I'm not sure about that. Because your library is pretty new, has no ratings yet and very few downloads on hackage. And maintainers are usually reluctant to add dependencies that are not (yet) popular, at least not if they are not adding a great value for many users, I think.

parsonsmatt:
I'm happy to let this sit until folks feel good integrating it. At the very least, the issue and PR presence is useful for folks that inevitably run into trouble when using `annotated-exception` with Yesod.

snoyberg (approved):

snoyberg:
I’m OK with the change if you find this useful. But there are merge conflicts now, sorry for the delayed review

lorenzo:
@parsonsmatt do you need help finishing this? I'm interested

parsonsmatt:
I need to verify that the `HCError <$> toHandlerError e` does the right thing here - I suspect it may not have the right behavior

yesod-core/src/Yesod/Core/Internal/Run.hs
```diff
@@ -116,7 +117,7 @@ basicRunHandler rhe handler yreq resState = do
...
                 Nothing -> HCError <$> toErrorHandler e)
```
parsonsmatt, line 121: OK, digging into this some more:

```haskell
-- | Convert a synchronous exception into an ErrorResponse
toErrorHandler :: SomeException -> IO ErrorResponse
toErrorHandler e0 = handleAny errFromShow $
    case fromException e0 of
        Just (HCError x) -> evaluate $!! x
        _ -> errFromShow e0
```

So I think *this* function needs to also check for `AnnotatedException` - since the `fromException e0 :: Maybe HCContents` won't work if the underlying application code is calling `throwWithCallStack`.

```haskell
-- | Generate an @ErrorResponse@ based on the shown version of the exception
errFromShow :: SomeException -> IO ErrorResponse
errFromShow x = do
  text <- evaluate (T.pack $ show x) `catchAny` \_ ->
          return (T.pack "Yesod.Core.Internal.Run.errFromShow: show of an exception threw an exception")
  return $ InternalError text
```

Fortunately this should work fine - perhaps we want to unwrap the `AnnotatedException` before showing it?

parsonsmatt (commented):
We have a Yesod middleware at work that handles this unwrapping, and as long as it's the *last* middleware, everything works fine.

I want to write some tests and verify that this is properly integrated before recommending merging.

Issue #1749, opened 2022-03-10
[`yesod-test`] Expose hook helpers


parsonsmatt:
In the work app, I've created a few extras:

```haskell
instance YesodDispatch site => Example (a -> SIO (YesodExampleData site) b) where
  type Arg (a -> SIO (YesodExampleData site) b) =
    (TestApp site, a)

  evaluateExample example' params action =
      Hspec.evaluateExample
          (action $ \((site, middleware), a) -> do
              app <- toWaiAppPlain site
              _ <- evalSIO (example' a) YesodExampleData
                  { yedApp = middleware app
                  , yedSite = site
                  , yedCookies = M.empty
                  , yedResponse = Nothing
                  }
              return ())
          params
          ($ ())

-- this is extremely dangerous. i should definitely NOT do this as soon as
-- yesod-test exports the SIO constructor. or even just evalSIO.

evalSIO :: SIO s a -> s -> IO a
evalSIO sio s =
  let (ReaderT f) = unsafeCoerce sio
  in
    newIORef s >>= f

beforeApp
  :: YesodDispatch site
  => SIO (YesodExampleData site) a
  -> SpecWith (TestApp site, a)
  -> SpecWith (TestApp site)
beforeApp action =
  beforeWith $ \testapp@(site, middleware) -> do
    app <- toWaiAppPlain site
    let yesodExampleData =
          YesodExampleData
            { yedApp = middleware app
            , yedSite = site
            , yedCookies = M.empty
            , yedResponse = Nothing
            }
    a <- evalSIO action yesodExampleData
    pure (testapp, a)

beforeWithApp
  :: YesodDispatch site
  => (a -> SIO (YesodExampleData site) b)
  -> SpecWith (TestApp site, b)
  -> SpecWith (TestApp site, a)
beforeWithApp action =
 beforeWith $ \(testapp@(site, middleware), a) -> do
    app <- toWaiAppPlain site
    let yesodExampleData =
          YesodExampleData
            { yedApp = middleware app
            , yedSite = site
            , yedCookies = M.empty
            , yedResponse = Nothing
            }
    b <- evalSIO (action a) yesodExampleData
    pure (testapp, b)
```

This allows you to share test setup nicely and nest that context.

```haskell
spec = withApp $ do
    beforeApp setupUser $ do
       it "something only a user needs" $ \(Entity userId User{..} -> do
           -- this is in YesodExample 
    beforeAppWith (\user -> (,) user <$> setupOrganization user) $ do
[...]

PR #1763, opened 2022-04-13
more of a comment than a question...

parsonsmatt:
So #1749 had me try an d implement hooks, which didn't work, because we go from `TestApp site -> YesodExampleData site` in the various example-runners and immediately throw that info away. So there is not a way to get hook behavior with the current implementation of the `yesod-test`.

I refactored it and made `yesod-hspec` in this repository, which follows the `hspec` pattern a bit more closely and allows hooks.

Not sure if this should be a huge breaking change to `yesod-test` or an alternative.

After submitting your PR:


yesod-hspec/test/main.hs
```diff
@@ -0,0 +1,662 @@
...
+                    addRequestHeader ("Content-Type","text/plain")
+                statusIs 415
+
+    describe "hooks" $ yesodSpec app $ do
+        ybefore_ (get ("/" :: Text)) $ do
+            yit "works" $ do
+                statusIs 200
+            ybefore_ (post ("/cookie/foo" :: Text)) $ do
+                yit "works again" $ do
+                    statusIs 303
```
parsonsmatt, line 364: This is the exciting thing. Now if you want a bunch of tests that share some setup, you can get it. This makes it much easier to write small tests that make small assertions and reduce code duplication.

parsonsmatt:
hmm now that i have figured this out, i am going to try and get it into a more minimal change so it will hopefully break fewer things

PR #1765, opened 2022-04-20
Add API for idempotent page description editing

ivanbakel:
`setDescription` and `setDescriptionI` are slightly surprising in that they are not idempotent, and multiple calls to these functions will insert multiple `<meta>` tags in the page head. Since these functions present like `setTitle` and `setTitleI`, I would have expected them to have a similar idempotent behaviour.

This adds idempotent versions of these functions, which behave like `setTitle` and `setTitleI` in that they store information in the page data and can be called any number of times, taking the final value supplied.

Because they use the page data, these functions can't be drop-in replacements for the existing functions - users with custom layouts will have to manually edit them to use the new idempotent functions, since they have to explicitly include the new `pageDescription` field in their layouts (while the old functions write to the page head.)

### Points for review

 * Is promoting `description` as a meta tag to something stored in the page data sensible? Should the same approach be taken for other meta tags? Should arbitrary meta tags appear in the page data?
 * Is using the page data a sensible place to store this kind of data? Should the per-request cache be used instead?

---

After submitting your PR:

### DIFF
--- a/yesod-core/ChangeLog.md
+++ b/yesod-core/ChangeLog.md
@@ -1,5 +1,13 @@
 # ChangeLog for yesod-core
 
+## 1.6.23
+
+* Add idempotent versions of `setDescription`, `setDescriptionI`. These functions
+  have odd behaviour when called multiple times, so they are now warned against.
+  This can't be a silent change - if you want to switch to the new functions, make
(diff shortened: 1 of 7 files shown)


snoyberg (approved):
Nice improvement! And great docs!

PR #1768, opened 2022-05-10
Fix quote ' not matching in htmlContain* functions

jappeace:
This sometimes occurred in our code base when generating
names with the fakedata package, someone named o'conner
randomly fails a particular test.

Also add tests for the other matching function and fixed them.
Furthermore, I added logging of the matches as well.

- [x] (not relevant) Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [x] (not relevant)  Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddocks for new, public APIs

After submitting your PR:

### DIFF
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -2,8 +2,8 @@
 
 ## 1.6.14
 
-* Fix quotes not matching in htmlContain* functions.
-* Add logging of the matches found of these functions.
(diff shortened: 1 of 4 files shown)


snoyberg (approved):
Test case demonstrates the problem nicely!

Issue #1771, opened 2022-06-27
Yesod is catching Timeout, which breaks the timeout middleware


pbrisbin:
It seems Yesod is catching the [`Timeout` exception](https://hackage.haskell.org/package/base-4.16.1.0/docs/System-Timeout.html#t:Timeout) used by the standard `timeout` function and turning it into a 500 instead of letting to get back to `timeout`. This means that when we use the [`Timeout` middleware](https://hackage.haskell.org/package/base-4.16.1.0/docs/System-Timeout.html#t:Timeout), instead of getting a 503 as desired, we get see `InternalError "<<timeout>>"` come through as a 500.

## Steps

Given the following executable:

```hs
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE QuasiQuotes #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}

module Main where

import Control.Concurrent (threadDelay)
import Data.Maybe (fromMaybe)
import Data.Text (Text)
import Network.HTTP.Types.Status (status200)
import Network.Wai (responseLBS)
import Network.Wai.Handler.Warp as Warp
import Network.Wai.Middleware.Timeout as Middleware
import System.Environment (lookupEnv)
import Text.Read (readMaybe)
import Yesod hiding (respond)

data App = App

mkYesod "App" [parseRoutes|
/ RootR GET
|]

getRootR :: Handler Text
getRootR = do
    liftIO $ threadDelay $ 3 * 1000000
    pure "Hello word\n"

instance Yesod App

simpleApp :: Application
simpleApp _req respond = do
    threadDelay $ 3 * 1000000
    respond $ responseLBS status200 [] "Hello world\n"

main :: IO ()
main = do
    app <- do
        x <- lookupEnv "YESOD"
        if x == Just "1" then toWaiAppPlain App else pure simpleApp

    port <- fromMaybe 3000 . (readMaybe =<<) <$> lookupEnv "PORT"
    Warp.run port $ Middleware.timeout 2 app
```

## Expected behavior

When run using the `simpleApp`, it does as expected (a 503):

```console
stack exec -- timeout-example
```

```console
% curl -D - localhost:3000
HTTP/1.1 503 Service Unavailable
Transfer-Encoding: chunked
Date: Mon, 27 Jun 2022 20:35:32 GMT
Server: Warp/3.3.21
```

## Actual behavior

[...]

PR #1772, opened 2022-07-06
Make catching exceptions configurable.

jappeace:
Fixes https://github.com/yesodweb/yesod/issues/1771

This is done by adding a function to Yesod
typeclass which can match on any exception
and tell the framework if it should rethrow
or not.

I used an overridable function because it seemed
more flexible then a whitelist.
A user can now for example choose to throw
everything, or catch everything as easily.

---

After submitting your PR:


yesod-core/src/Yesod/Core/Class/Yesod.hs
```diff
@@ -70,6 +73,17 @@ class RenderRoute site => Yesod site where
...
 
+    -- | @since 1.6.23.2
+    --   Should we catch an exception, or rethrow it.
+    --   Rethrowing an exception lets the webserver deal with it
+    --   (usually warp).
+    --   catching allows yesod to render the error page.
+    --   the default 'defaultCatchBehavior' is to catch everything
+    --   (even async), except for the
+    --   'Warp.ConnectionClosedByPeer' constructor.
+    catchBehavior :: Proxy site -> SomeException -> CatchBehavior
```
snoyberg, line 84: This may be too limiting an API. We've often run into situations later where people say they want to be able to change behavior at runtime. What do you think of something like:

```haskell
catchBehavior :: site -> SomeException -> IO CatchBehavior
```

snoyberg: Also, we may want to change the default behavior back to what it used to be (not catching async exceptions).

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -169,6 +169,13 @@ newtype WaiSubsite = WaiSubsite { runWaiSubsite :: W.Application }
 -- @since 1.4.34
 newtype WaiSubsiteWithAuth = WaiSubsiteWithAuth { runWaiSubsiteWithAuth :: W.Application }
 
+-- | @since 1.6.23.2
+data CatchBehavior = Rethrow -- ^ Rethrow an exception and let the webserver deal with it (usually warp)
```
snoyberg, line 173: I'm not sure what other behaviors may be relevant in the future, but we may want other possibilities than just rethrow and catch (e.g., "wrap up with this function"). Providing an abstract type and smart constructors would allow for future extensibility.

snoyberg (changes requested):
@pbrisbin would you be able to look this over and see if it addresses your needs?

yesod-core/src/Yesod/Core/Class/Yesod.hs
snoyberg: @pbrisbin  thoughts?

pbrisbin: Hmm. Well this is tricky because I'm coming at it from a couple of different perspectives.

As a user of Yesod _on behalf of my users_, I'd agree that I'd **prefer** if the normal error handling and error page triggered on `ThreadKilled` and `AllocationLimitExceeded`. Because, why not?

As an operator of Yesod, I'm not sure if I'd hold that preference for "most async exceptions", because I think there are a lot of async exceptions that are normal in the context of handling web requests, don't ever reach or impact end users, and I don't want them inflating error metrics as if they had.

As someone familiar with the implementation inolved, I don't agree that I'd **expect** the normal error handling and error page triggered on these two exceptions in particular. In my head, the scope of handling a web request is

```
Warp
  Wai
   |----------|
   | Yesod    |
   |   My App |
   |----------|
```

In the case of these two exceptions in particular, I'd expect everything I've boxed there to be dead, like `kill -9` dead. And so I personally wouldn't expect any error recovery to be possible at that level for such exceptions.

Knowing what I've learned recently about async exceptions, I think I could safely apply this expectation to "most async exceptions". I think they all represent `kill -9`-like messaging and so there should be no recovery.

jappeace: so I guess for most users this default of catching everything except timeout & ConnectionClosedByPeer constructor is sensible, but as an operator you could still override this behavior? In other words, everyone is happy?

pbrisbin: As long as we end up in a place where `TimeoutError` behaves correctly, I'm happy. If we also end up in a place where it's easy for me to say "don't catch any async exceptions", I'm very happy.

I think "don't catch async exceptions" is a more sensible default, for the reasons I tried to describe above, but it's subjective and I don't mind if you go the other way as you're writing the code and Michael can choose to merge it. I just don't see either of us convincing the other, but I don't think we need to.

pbrisbin:
Everything's good from my end, but just to reply to your questions:

> how would this work then? if the callback doesn't throw an exception we should let yesod do the error rendering? It's not very obvious to me from just the type signature.

Yeah I was being vague, because I don't know exactly how it might work. I just know that your original PR (basically) replaced an `UnliftIO.catch` with an `EUnsafe.catch`, so I thought an initial stab at configurability would be to inject exactly that:

```hs
class Yesod where
  -- | By default we catch all exceptions to an error page, replace this with
  -- 'UnliftIO.Exceptions.catch' to only catch synchronous exceptions
  catch = EUnsafe.catch
```

But I know this probably won't work because `Yesod(catch)` has to be at some `m` that both of those could satisfy.

> This would work, but I think by default yesod should handle exceptions properly

Yeah, my example showed it with either default behavior. I was just curious if I was missing something that meant using `yesodMiddleware` to add exception-handling not function, but it sounds like it that's not the case.

yesod-core/src/Yesod/Core/Class/Yesod.hs
jappeace: I added a test to verify by default we don't catch `Timeout` either: https://github.com/yesodweb/yesod/pull/1772/files#diff-bd8750d9a236376b742e82438d8e92aeeb76f1d55827fc85c21154c5c4e8f9a0R378

I think that's what you meant, right?

There is a rethrow async function as well (although not as well tested): https://github.com/yesodweb/yesod/pull/1772/files#diff-1fdc39d8a397f906d8f4f69a087048f3432df0b1fd2238ee5459ff1e89bb0edfR655

snoyberg: > I think "don't catch async exceptions" is a more sensible default

I think I lean towards this way of thinking, I was already nervous about the original approach since it violates what I generally recommend for async exception handling. I'd be tempted to call the previous change in behavior a bug (since it silently broke other people's code) and go back to the original default, still exposing this configuration point.

To fully flesh out the thinking here, I see four points in the design space for defaults:

1. Rethrow all async exceptions (original behavior)
2. Rethrow a specific list of async exceptions (current PR)
3. Catch a specific list of async exceptions (not proposed yet)
4. Catch all async exceptions (current behavior)

I'm not convinced (2) is a good compromise point, since including Timeout may seem like a good idea here, but honestly seems arbitrary to me. It's entirely possible there's someone else out there using a different async exception for middlewares that we're not aware of.

snoyberg:
> But I know this probably won't work because Yesod(catch) has to be at some m that both of those could satisfy.

Given the fact that everything involved is a `MonadUnliftIO` (I think), specializing this to `IO` should work, and the internals of Yesod can wrap it up using `withRunInIO`. In other words, a new method along these lines may solve the problem:

```haskell
catchHandlerExceptions :: site -> IO a -> (SomeException -> IO a) -> IO a
```

yesod-core/src/Yesod/Core/Class/Yesod.hs
jappeace: both choices are arbitrary, in anycase I'll rewrite it to the 'rethrow async exception' behavior as default. having this configurable is good enough for me.

pbrisbin (approved):
I really like this implementation! Thanks for working through it.

snoyberg (approved):
LGTM, thanks!

PR #1775, opened 2022-08-17
Add withRadioField a more flexible radio option renderer

jappeace:
This re-expresses radioField into the new more flexible
function.
Which gives an adhoc example on how to use it as well.

This function passes the radio input to a callback function
to let said function decide how it should be rendered.
These changes allow you to make a radio table for example,
for selecting some row.

After submitting your PR:


schoettl:
Good idea! Looks good to me, without having tested it.

snoyberg (approved):
LGTM thanks!

PR #1786, opened 2022-10-11
Add versions of 'htmlAnyContain', 'htmlAllContain', 'htmlNoneContain' which match strings verbatim

### LINKED ISSUE #1784: [yesod-test] bring back support for matching of preescaped strings
When I upgraded `yesod-test` to version 1.6.15 I noticed some tests started to fail, e.g.:
```
htmlAnyContain "meta[name=description]" "content=\"This is the site description\""
```
Fails with:
```
None of meta[name=description] contain content="This is the site description matches: ["<meta content=\"This is the site description\" name=\"description\" />"]
```

AFAIU this is because `htmlAnyContain` was changed to support matching e.g. quotes (`'`) [by escaping input by default](https://github.com/yesodweb/yesod/pull/1768/files#diff-6d415466ccc6dcf02506aa9f79ed82be2422bbbefd44bbafe603086b30527a97R719).

I guess it's not possible to support both of these cases with the same API? To work around this locally I have the old version of `htmlAnyContain` around which does not escape input. If that's sounds like a sensible solution, I could open a PR which brings back that version under a new name (e.g. `htmlAnyContainPreEscaped`) to avoid other people (and my future self) running into this issue.

eahlberg:
Fixes #1784

After submitting your PR:

### DIFF
--- a/yesod-test/ChangeLog.md
+++ b/yesod-test/ChangeLog.md
@@ -1,5 +1,8 @@
 # ChangeLog for yesod-test
 
+## 1.6.16
+
+* Add `htmlAnyContainPreEscaped`, `htmlAllContainPreEscaped`, `htmlNoneContainPreEscaped`. [#1786](https://github.com/yesodweb/yesod/pull/1786)
 
 ## 1.6.15
 


yesod-test/Yesod/Test.hs
```diff
@@ -204,6 +204,9 @@ module Yesod.Test
     , htmlAllContain
     , htmlAnyContain
     , htmlNoneContain
+    , htmlAllContainPreEscaped
+    , htmlAnyContainPreEscaped
+    , htmlNoneContainPreEscaped
```
eahlberg, line 209: Is `htmlAllContainRaw` etc better than `htmlAllContainPreEscaped`? `PreEscaped` was used since that's what it's called in the [Blaze docs](https://hackage.haskell.org/package/blaze-markup-0.8.2.8/docs/Text-Blaze-Internal.html#v:preEscapedString)

yesod-test/test/main.hs
```diff
@@ -201,6 +201,9 @@ main = hspec $ do
             yit "finding html" $ do
                 get ("/html" :: Text)
                 statusIs 200
+                htmlAnyContainPreEscaped "meta" "content=\"site description\""
+                htmlAllContainPreEscaped "meta"  "content=\"site description"
```
eahlberg, line 205: It'd be nice if there are any more concise examples!

yesod-test/Yesod/Test.hs
```diff
@@ -754,6 +757,61 @@ htmlNoneContain query search = do
...
+-- > htmlAllContainPreEscaped "meta" " "content=\"site description\""-- Every <meta> tag contain the string "content=site description" verbatim
+--
+-- Since 16.6.6
+htmlAllContainPreEscaped :: HasCallStack => Query -> String -> YesodExample site ()
+htmlAllContainPreEscaped query search = do
+  matches <- htmlQuery query
+  case matches of
+    [] -> failure $ "Nothing matched css query: " <> query
+    _ -> liftIO $ HUnit.assertBool ("Not all "++T.unpack query++" contain "++search  ++ " matches: " ++ show matches) $
+          DL.all (DL.isInfixOf search) (map (TL.unpack . decodeUtf8) matches)
```
eahlberg, line 776: Except for the escaping, this is the exact same code as `htmlAllContain`, it might make refactor this a bit.

yesod-test/test/main.hs
```diff
@@ -201,6 +201,9 @@ main = hspec $ do
...
+                htmlNoneContainPreEscaped "meta" "foobar"
```
eahlberg, line 206: This checks absence in the same was as the test for `htmlNoneContain`, I think it'd be better with a test case which illustrate the difference between the two functions.

PR #1798, opened 2023-03-30
yesod-persistent: don't depend on persistent-template if we don't have to

peti:
If our version of `persistent` is sufficiently new, then we don't need a dependency on the empty `persistent-template` package.

### DIFF
--- a/yesod-persistent/yesod-persistent.cabal
+++ b/yesod-persistent/yesod-persistent.cabal
@@ -13,12 +13,15 @@ homepage:        http://www.yesodweb.com/
 description:     API docs and the README are available at <http://www.stackage.org/package/yesod-persistent>
 extra-source-files: README.md ChangeLog.md
 
+flag persistent_has_persistent_template
+  default: True
+  manual: False
+
 library
     default-language: Haskell2010
     build-depends:   base                      >= 4.10     && < 5
                    , yesod-core                >= 1.6      && < 1.8
-                   , persistent                >= 2.8
-                   , persistent-template       >= 2.1
+e if we don't have to)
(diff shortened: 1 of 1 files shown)


yesod-persistent/yesod-persistent.cabal
```diff
@@ -13,12 +13,13 @@ homepage:        http://www.yesodweb.com/
 description:     API docs and the README are available at <http://www.stackage.org/package/yesod-persistent>
 extra-source-files: README.md ChangeLog.md
 
+flag persistent_has_persistent_template
+  default: True
```
snoyberg, line 17: IIRC, this should have `manual: False`, right?

peti:
"manual" is false by default according to https://cabal.readthedocs.io/en/stable/cabal-package.html#pkg-field-flag-manual. We could add it anyways, just to be explicit, I guess, but technically it should make no difference.

yesod-persistent/yesod-persistent.cabal
peti: Interesting ... the automatic flag selection doesn't seem to work in stack. At least that is how I interpret the CI failures. Does stack have a different default for manual than cabal-install does?

snoyberg: Stack doesn't do dependency solving, and therefore it would need to be told to override any defaults set in the cabal file.

peti: I have added `manual: False`.

peti:
I have re-based the pull request on the current `master` version of the repository.

PR #1801, opened 2023-04-22
Forms i18n romanian

ciukstar:

After submitting your PR:

### DIFF
--- a/yesod-form/Yesod/Form/I18n/Romanian.hs
+++ b/yesod-form/Yesod/Form/I18n/Romanian.hs
@@ -27,4 +27,4 @@ romanianFormMessage (MsgInvalidBool t) = "Valoare booleană nevalidă: " `mappen
 romanianFormMessage MsgBoolYes = "Da"
 romanianFormMessage MsgBoolNo = "Nu"
 romanianFormMessage MsgDelete = "Șterge?"
-romanianFormMessage (MsgInvalidHexColorFormat t) = "Culoarea nevalidă. Formatul necesar este #rrggbb în hexazecimal: " `mappend` t
+romanianFormMessage (MsgInvalidHexColorFormat t) = "Culoare nevalidă. Formatul necesar este #rrggbb în hexazecimal: " `mappend` t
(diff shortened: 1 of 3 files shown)


yesod-form/Yesod/Form/I18n/Romanian.hs
```diff
@@ -0,0 +1,31 @@
...
+
+import Yesod.Form.Types (FormMessage (..))
+import Data.Monoid (mappend)
+import Data.Text (Text)
+
+-- |
+-- Romanian translation
+--
+-- @since 1.7.5
+englishFormMessage :: FormMessage -> Text
```
jezen, line 12: I think this ought to be Romanian, rather than English.

```suggestion
romanianFormMessage :: FormMessage -> Text
```

Also, perhaps it's worth writing this with LambdaCase (though I recognise none of the other modules do this).

jezen (changes requested):
I think we just need to update the function name, and then this is good to go.

yesod-form/Yesod/Form/I18n/Romanian.hs
```diff
@@ -0,0 +1,31 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Yesod.Form.I18n.Romanian where
+
+import Yesod.Form.Types (FormMessage (..))
+import Data.Monoid (mappend)
+import Data.Text (Text)
+
+-- |
+-- Romanian translation
```
jezen, line 9: Why are these separate lines? I would expect to see:

```suggestion
-- | Romanian translation
--
```

yesod-form/Yesod/Form/I18n/Romanian.hs
ciukstar: Changed function name from ```englishFormMessage``` to ```romanianFormMessage``` in module ```Yesod.Form.I18n.Romanian```  (fdfac8de6f270f2eefaf44b90cae766bbca3c936).

```LambdaCase``` extension is supported since ```7.6.1```.  Can I still use it? I'm happy to change if that's not a problem.

yesod-form/Yesod/Form/I18n/Romanian.hs
ciukstar: Fixed (fdfac8de6f270f2eefaf44b90cae766bbca3c936).

yesod-form/Yesod/Form/I18n/Romanian.hs
jezen: Good point about less extensive support. It's only syntax sugar anyway, so it's not worth the compromise IMO.

PR #1802, opened 2023-06-20
Add attoparsec-aeson for upcoming aeson-2.2

9999years:


yesod-core/yesod-core.cabal
```diff
@@ -26,7 +26,8 @@ library
     hs-source-dirs: src
 
     build-depends:   base                  >= 4.10     && < 5
-                   , aeson                 >= 1.0
+                   , aeson                 >= 2.2
+                   , attoparsec-aeson      >= 2.2
```
parsonsmatt, line 30: `attoparsec-aeson` defines `Data.Aeson.Parser` module which I don't think we're importing - what's the motivation for this change?

parsonsmatt: Huh. I could have sworn I did a `rg Data.Aeson.Parser` and didn't find anything 🤔 

I filed [this issue](https://github.com/haskell/aeson/issues/1048) which should allow us to have a more lax version bound than `>= 2.2` at least - [according to the `attoparsec-aeson` Hackage page](https://hackage.haskell.org/package/attoparsec-aeson-2.1.0.0), we shoudl eb able to depend on `attoparsec-aeson >= 2.1.0.0` along with the existing `aeson` bound.

parsonsmatt, line 30:
```suggestion
                   , aeson                 >= 1.0
                   , attoparsec-aeson      >= 2.1
```

This ought to fix things

PR #1815, opened 2023-08-17
Add yesod test getCookies utility

josephsumabat:

Recently was trying to write a test which required that I clear cookies and set them again later on in the test. 

```
cookies <- YT.getCookies
YT.clearCookies

<do something with cookies after>....
```

Unfortunately `getRequestCookies` only allows you to access the cookies in a `RequestBuilder site` context since the type of request is `RequestBuilder site () -> YesodExample site ()` thus something like 

```
cookies <- request $ do
  getRequestCookies
```
unfortunately does not typecheck and you can only access the cookies from within the request builder. This PR adds the utility funciton `getCookies` which simply enables the first use case allowing cookies to be accessed outside of a request context without resorting to using an `IORef` 

After submitting your PR:


yesod-test/Yesod/Test.hs
```diff
@@ -329,6 +331,13 @@ getTestYesod = fmap yedSite getSIO
 getResponse :: YesodExample site (Maybe SResponse)
 getResponse = fmap yedResponse getSIO
 
+-- | Get the cookies set for the current test. Differs from `getRequestCookies`
+-- in that this allows you to access cookies outside of a request builder
+--
+-- Since 1.6.16
```
parsonsmatt, line 337:
```suggestion
-- @since 1.6.16
```

parsonsmatt (approved):
lgtm! do you mind updating the cabal file to make an easy releate?

Thanks!

yesod-test/Yesod/Test.hs
parsonsmatt: Oh, looks like 1.6.16 was already released, so needs to be 1.6.17

josephsumabat: Bumped!

PR #1818, opened 2023-09-11
Add attoparsec-aeson to support aeson-2.2

Vekhir:
The module `Data.Aeson.Parser` is moved into attoparsec-aeson for aeson >=2.2. For aeson <2.2, attoparsec-aeson is an empty package, since the module exists within aeson.

This is the fix as suggested in #1802, but that PR was never updated for that.

After submitting your PR:

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -14,3 +14,6 @@ packages:
 - ./yesod
 - ./yesod-eventsource
 - ./yesod-websockets
+
+extra-deps:
+- attoparsec-aeson-2.1.0.0
--- a/stack.yaml.lock
+++ b/stack.yaml.lock
@@ -3,10 +3,17 @@
 # For more information, please see the documentation at:
 #   https://docs.haskellstack.org/en/stable/lock_files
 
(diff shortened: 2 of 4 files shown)


parsonsmatt:
Looks like we need to add `aeson-attoparsec` to Stackage before CI is going to pass.

Vekhir:
@parsonsmatt Isn't it already? https://www.stackage.org/package/attoparsec-aeson (with version 2.1.0.0). It probably wouldn't hurt to update it though

Additionally, the error messages states, confusingly, that the `latest matching version is 2.2.0.0`.

snoyberg:
Looks like the new package needs to be added to `extra-deps` in the `stack.yaml` file.

Vekhir:
Added `attoparsec-aeson` 2.1.0.0 to `stack.yaml` and updated the `stack.yaml.lock` file

snoyberg (approved):
Thanks!

Vekhir:
Thanks for the quick release!

PR #1819, opened 2023-09-13
Add an options data structure to allow fine-tuned control of what instances are generated for a route

Benjamin-McRae-Tracsis:

After submitting your PR:

### DIFF
--- a/yesod-core/yesod-core.cabal
+++ b/yesod-core/yesod-core.cabal
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.24.6
+version:         1.6.25.0
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>
--- a/yesod-core/yesod-core.cabal
+++ b/yesod-core/yesod-core.cabal
@@ -26,7 +26,7 @@ library
     hs-source-dirs: src
 
     build-depends:   base                  >= 4.10     && < 5
-                   , aeson                 >= 1.0
+                   , aeson                 >= 1.0      && <= 2.1.2.1
                    , auto-update
                    , blaze-html            >= 0.5
                    , blaze-markup          >= 0.7.1
--- a/yesod-core/yesod-core.cabal
+++ b/yesod-core/yesod-core.cabal
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.24.4
+version:         1.6.25
 license:         MIT
 license-file:    LICENSE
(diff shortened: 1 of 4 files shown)


parsonsmatt (changes requested):
So that's going to be a breaking change for any library user that uses the `Read` instance of the type

To avoid a major breaking change, we need to keep the default behavior of `deriving Read`, but provide an option for the user to disable `Read` derivation.

Benjamin-McRae-Tracsis:
I can try to work on that.

How would you want this configurable? I can make a mkRouteConsOpt that requires a config data structure, which has three booleans: `hasShow`, `hasEq`, `hasRead`, and then make duplicate versions of all the functions that call mkRouteCons that are suffixed with Opts to take said Opts.

parsonsmatt:
I'm not terribly particular. I think @snoyberg mentioned using a directive in the actual routes file would be fine - maybe something like:

```
!no-derive-read
```

as a *very* quick, patch-level fix.

Defining an `Opts` type make sense - I'd prefer a specific name (ie `RouteDeriveOpts`), and ideally a hidden constructor with setter functions so new fields aren't a breaking change. However, since we don't *need* to use this to derive instances - users can use `StandaloneDeriving` to write `deriving stock Read (Route App)`, after all - I don't anticipate many changes to this type, so hiding the constructor and setters would be a matter of taste rather than real pragmaticism.

Benjamin-McRae-Tracsis:
Are there other examples of such directives? I'm looking at the parser and I'm not sure whether it does.

yesod-core/yesod-core.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-core
-version:         1.6.24.5
+version:         1.6.24.6
```
parsonsmatt, line 2:
```suggestion
version:         1.6.25.0
```

parsonsmatt (changes requested):
PR looks great, thank you! I made some suggestions for the version number + `@since` comments.

parsonsmatt (approved):
Looks great! Thank you 😄

snoyberg:
@parsonsmatt absolutely, I'd be happy to. What's your Hackage username?

parsonsmatt:
It's `parsonsmatt` there too. Thank you!

snoyberg:
@parsonsmatt I've given you maintainer rights on Hackage for yesod-core and a bunch of other Yesod packages too. If you need any more, just let me know. Thank you!

PR #1828, opened 2023-11-22
Upstream helper functions to Yesod.Form.Option.

LukeTemp:
The 2 new functions are:
- optionsFromList': creates an OptionList from a List, using the PathPiece instance for the external value and a custom function for the user-facing value.
- optionsEnum': creates an OptionList from an enumeration.

Also bump the version and update the changelog.

- [X] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)

After submitting your PR:

- [X] Update the Changelog.md file with a link to your PR
- [] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

### DIFF
--- a/yesod-form/ChangeLog.md
+++ b/yesod-form/ChangeLog.md
@@ -3,8 +3,8 @@
 
 ## 1.7.7
 
-* Added `optionsFromList'` to create an OptionList from a List, using the PathPiece instance for the external value and a custom
-function for the user-facing value. Also added `optionsEnum'` to create an OptionList from an enumeration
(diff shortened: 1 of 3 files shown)


snoyberg (commented):
This overall seems like a very specific solution, I'm not sure it belongs in the library.

snoyberg:
@jezen more than happy to defer to you on the decision of inclusion

jezen (approved):
This is good — the examples are really clear. Thank you.

It'd be great to see automated tests, and haddocks for each of the function arguments, but let's go with this as-is for now.

PR #1830, opened 2023-12-11
Fix/multi select field grouped

SJost:
fix multiSelectField being partial

Constructor OptionListGrouped could not be processed by function multiSelectField. This patch fixes this, using <optgroup> tags within the generated Html.

--------------

After submitting your PR:

### DIFF
--- a/yesod-form/Yesod/Form/Fields.hs
+++ b/yesod-form/Yesod/Form/Fields.hs
@@ -52,7 +52,7 @@ module Yesod.Form.Fields
     , withRadioField
     , checkboxesField
     , checkboxesFieldList
-    , multiSelectField
+    , multiSelectField    
     , multiSelectFieldList
     , Option (..)
     , OptionList (..)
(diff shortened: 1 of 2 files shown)


SJost:
@schoettl Happy to help. Though I am not sure what to do with the remaining open instructions to update the Changelog.md? How do I know which version the PR is merged into, if at all?

SJost:
Furthermore, in the version that I am actually using, I have an additional `Maybe (SomeMessage site)` argument that is used as the `None` Option, if `isReq` is `False`, since my users had problems with de-selecting. 

I could add versions `selectField'` and `multiSelectField'` with this option, if desired, but I thought it might only dilute this PR.

schoettl:
> Furthermore, in the version that I am actually using, I have an additional `Maybe (SomeMessage site)` argument that is used as the `None` Option, if `isReq` is `False`, since my users had problems with de-selecting.

You mean your users have problems to clear the selection in the multi select field, which is usually done in desktop browsers by Ctrl + click? TBH, I would leave this to the developer and not provide an extra implementation for that.

Regarding the open checkboxes, I think you can just add a note in the changelog for the new version you set.

And it looks to me as if CI would be started at approval of a maintainer. I think it has timed-out too often.

PR #1837, opened 2024-07-08
Stop testing against stackage lts 11/12

dten:
since the requirement for attoparsec-aeson >= 2.1 was added this will no longer ever succeed

After submitting your PR:

Since the MR that required attoparsec-aeson >= 2.1 https://github.com/yesodweb/yesod/pull/1818 then the older LTS that only have older aesons will never pass. This is confusing for any future MRs and seems like it should be removed it no one expects it to pass again.

Similarly there are two macos tests that fail as `stack` is not present, perhaps these should be removed until fixed also


snoyberg (approved):
Thanks!

PR #1840, opened 2024-07-29
Implemenatation of chooseRadio function along with tests

chrissound:

After submitting your PR:

- [X] Update the Changelog.md file with a link to your PR

I've been asked to upstream this work, so here it is! It's basically a helper function to make it easier to set radio form inputs.

### DIFF
--- a/yesod-test/Yesod/Test.hs
+++ b/yesod-test/Yesod/Test.hs
@@ -966,10 +966,10 @@ genericNameFromHTML match label html =
 -- Given this HTML, and we want to submit @color=Red@ to the server:
 --
 -- > <form method="POST">
+-- >   <input id="color-red" name="color" value="Red">
 -- >   <label for="color-red">Red</label>
--- >   <input id="color-red" name="color" value="Red" />
+-- >   <input id="color-blue" name="color" value="Blue">
 -- >   <label for="color-blue">Blue</label>
--- >   <input id="color-blue" name="color" value="Blue" />
 -- > </form>
 --
 -- You can set this parameter like so:
@@ -980,18 +980,16 @@ genericNameFromHTML match label html =
 -- @since 1.6.17
 chooseRadio :: T.Text -- ^ The text contained within the @\<label>@.
        -> RequestBuilder site ()
-chooseRadio v = do
+chooseRadio labelText = do
   mres <- fmap rbdResponse getSIO
-  res <-
-    case mres of
+  res <- case mres of
       Nothing -> failure "chooseRadio: No response available"
-      Just res -> return res
-  let body = simpleBody res
+      Just res -> pure res
 
-  let name = genericNameFromHTML (==) (v) body
+  let name = genericNameFromHTML (==) labelText (simpleBody res)
   case name of 
-    Right name' -> addPostParam name' $ v
-    Left e -> failure $ e
+    Right name' -> addPostParam name' labelText
+    Left err -> failure err
 
 byLabelWithMatch :: (T.Text -> T.Text -> Bool) -- ^ The matching method which is used to find labels (i.e. exact, contains)
                  -> T.Text                     -- ^ The text contained in the @\<label>@.


yesod-test/Yesod/Test.hs
```diff
@@ -958,6 +959,40 @@ genericNameFromHTML match label html =
...
 
+
+-- | Choose a given value from a radio input, identified by the value of it's corresponding label.
+-- This is a very simple and naive function, which could fail in more complex scenarios. Potentially still quite useful.
+--
+-- Given this HTML, and we want to submit @color=Red@ to the server:
+--
+-- > <form method="POST">
+-- >   <label for="color-red">Red</label>
+-- >   <input id="color-red" name="color" value="Red" />
```
jezen, line 970: I think typical radio input markup should look like this:

```html
<input id="color-red" name="color" value="red">
<label for="color-red">Red</label>
```

_i.e._, the input should come before the label. Also, self-closing tags don't need a trailing slash.

yesod-test/Yesod/Test.hs
```diff
@@ -958,6 +959,40 @@ genericNameFromHTML match label html =
...
+--
+-- You can set this parameter like so:
+--
+-- > request $ do
+-- >   chooseRadio "Red"
+--
+-- @since 1.6.17
+chooseRadio :: T.Text -- ^ The text contained within the @\<label>@.
+       -> RequestBuilder site ()
+chooseRadio v = do
```
jezen, line 983: I think `v` could be a better name. Perhaps `labelText`?

yesod-test/Yesod/Test.hs
```diff
@@ -958,6 +959,40 @@ genericNameFromHTML match label html =
...
+  mres <- fmap rbdResponse getSIO
+  res <-
+    case mres of
+      Nothing -> failure "chooseRadio: No response available"
+      Just res -> return res
```
jezen, line 988: Since you're only using `mres` in one place, you may as well inline it.

```suggestion
  res <- case rbdResponse <$> getSIO of
    Nothing -> failure "chooseRadio: No response available"
    Just res -> pure res
```

yesod-test/Yesod/Test.hs
```diff
@@ -958,6 +959,40 @@ genericNameFromHTML match label html =
...
+  let body = simpleBody res
```
jezen, line 989: Similar deal here — this binding is only used once, so it may as well be inlined below.

yesod-test/Yesod/Test.hs
```diff
@@ -958,6 +959,40 @@ genericNameFromHTML match label html =
...
+
+  let name = genericNameFromHTML (==) (v) body
```
jezen, line 991: Unless my eyes (and brain) are failing me, I think the parentheses are redundant here.

```suggestion
  let name = genericNameFromHTML (==) labelText (simpleBody res)
```

yesod-test/Yesod/Test.hs
```diff
@@ -958,6 +959,40 @@ genericNameFromHTML match label html =
...
+  case name of 
+    Right name' -> addPostParam name' $ v
+    Left e -> failure $ e
```
jezen, line 994: The function application operators are redundant in both cases. Also, let's try to use _slightly_ more descriptive names 🙂 

```suggestion
    Right name' -> addPostParam name' labelText
    Left err -> failure err
```

jezen (changes requested):
This is an excellent addition the the library, though I think we ought to tidy up some minor details first.

chrissound:
Thanks, will try sort this all out later today.

yesod-test/Yesod/Test.hs
chrissound: I'm pretty sure this can't be inlined unless using LamdbdaCase `>>= \case`? Otherwise we need `>>= (\v -> case v of ...)` but I think that is less readable but happy to do that if preferred.

Issue #1850, opened 2024-10-24
instance Default WidgetFileSettings vanished


thielema:
After a `cabal update` I get the following type error when compiling `hledger-web`:
~~~~
    • No instance for ‘Data.Default.Internal.Default
                         WidgetFileSettings’
        arising from a use of ‘def’
~~~~

Did something change in one of the `yesod` packages?

jezen:
The [file where this instance is defined][0] hasn't changed in the best part of eight years. I don't see any changes to Yesod that would have caused this in the past half year either. Could the change be somewhere else, like in hledger?

[0]: https://github.com/nh2/yesod/blob/master/yesod/Yesod/Default/Util.hs

thielema:
On Thu, 24 Oct 2024, Jezen Thomas wrote:

> The file where this instance is defined hasn't changed in the best part 
> of eight years. I don't see any changes to Yesod that would have caused 
> this in the past half year either. Could the change be somewhere else, 
> like in hledger?

I assumed it could be caused by Cabal choosing two different versions of 
the same yesod package at different places. But Cabal has not shown a 
warning about such a conflict. Hm.

thielema:
On Thu, 24 Oct 2024, Jezen Thomas wrote:

> Could it perhaps be related to this?
> 
> #1848 (comment)

That one looks different.

I assumed my problem could have to do with an orphan instance, but 
"instance Default WidgetFileSettings" is defined right after 
"WidgetFileSettings" in Yesod.Default.Util.

thielema:
On Thu, 24 Oct 2024, thielema wrote:

> I assumed my problem could have to do with an orphan instance, but 
> "instance Default WidgetFileSettings" is defined right after 
> "WidgetFileSettings" in Yesod.Default.Util.

I think I found the problem. yesod imports package data-default-class and 
defines instance Data.Default.Class.Default WidgetFileSettings, whereas 
hledger expects instance Data.Default.Default WidgetFileSettings from 
package data-default. Formerly, data-default just re-exported Default 
class from data-default-class, but in data-default-0.8.0.0 there seems to 
be a custom definition. data-default-0.8.0.0 was released yesterday.

I do not know whether this means that data-default-class becomes 
deprecated or what the relation between data-default and 
data-default-class is intended to be.

thielema:
The author of `data-default-class` tells, that in future `data-default` should be preferred:
https://github.com/mauke/data-default/issues/29

jezen:
[...]

PR #1852, opened 2024-10-28
Switched to data-default instead of data-default-class. Fixed https://github.com/yesodweb/yesod/issues/1850

Oleksandr-Zhabenko:
Switched to data-default instead of data-default-class. Fixed https://github.com/yesodweb/yesod/issues/1850

### DIFF
--- a/stack.yaml.lock
+++ b/stack.yaml.lock
@@ -11,6 +11,20 @@ packages:
       size: 114
   original:
     hackage: attoparsec-aeson-2.1.0.0
+- completed:
+    hackage: crypton-1.0.0@sha256:637e58581978c84ef1288d14fa9cac1d2905ef60e319924293bc11250aca882d,14527
+    pantry-tree:
+      sha256: 4b5e5511567c0fe735a224cb8b2b278e1caa79344f2940d030d169e69b1b81e1
+      size: 23275
+  original:
+    hackage: crypton-1.0.0
+- completed:
+    hackage: crypton-conduit-0.2.3@sha256:31f44243b42f344c65be6cd2c39c07994d9186d19d15988656620d1de85aca37,1946
+    pantry-tree:
+      sha256: 06781001956f2ccfae0e6b4f33c213bd3121c6462f8534e9dca87bf51e4663e0
+      size: 592
+  original:
+    hackage: crypton-conduit-0.2.3
 snapshots:
 - completed:
     sha256: 694573e96dca34db5636edb1fe6c96bb233ca0f9fb96c1ead1671cdfa9bd73e9
--- a/yesod-bin/yesod-bin.cabal
+++ b/yesod-bin/yesod-bin.cabal
@@ -1,5 +1,5 @@
 name:            yesod-bin
-version:         1.6.2.3
+version:         1.6.2.4
 license:         MIT
 license-file:    LICENSE
 author:          Michael Snoyman <michael@snoyman.com>
@@ -31,7 +31,7 @@ executable             yesod
                      , conduit            >= 1.3
                      , conduit-extra      >= 1.3
                      , containers         >= 0.2
-                     , data-default-class
+                     , data-default       >= 0.8
                      , directory          >= 1.2.1
                      , file-embed
                      , filepath           >= 1.1
--- a/yesod-test/yesod-test.cabal
+++ b/yesod-test/yesod-test.cabal
@@ -64,7 +64,7 @@ test-suite test
                           , containers
                           , html-conduit
                           , yesod-core
-                          , yesod-form >= 1.6
+                          , yesod-form >= 1.7.9
                           , text
                           , wai
                           , wai-extra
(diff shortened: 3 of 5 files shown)


jezen (approved):

jezen (changes requested):
@Oleksandr-Zhabenko FYI, this is failing to build.

Oleksandr-Zhabenko:
> @Oleksandr-Zhabenko FYI, this is failing to build.

Stackage still is missing the newer version that will completely resolve the issue. It is present only on Hackage. Since the files changed are all .cabal files, I thought that it would update the needed versions from the Hackage, but it does not. It looks like it is needed to use conditional builds in the .cabal files and CPP extensions in the source files to check the package versions.

Oleksandr-Zhabenko:
But if we use CPP and conditional .cabal files then the issue [#1850](https://github.com/yesodweb/yesod/issues/1850) is still actual, because hledger still uses data-default package, which is preferred in the future, instead of data-default-class.

Oleksandr-Zhabenko:
[This two commits](https://github.com/yesodweb/yesod/compare/master...Oleksandr-Zhabenko:yesod:data-default)
seem to resolve the issue.

Oleksandr-Zhabenko:
@jezen To use just one commit, 
[this one](https://github.com/yesodweb/yesod/compare/master...Oleksandr-Zhabenko:yesod:data-default) is done.

jezen:
There are a couple of build failures on ubuntu-latest, --resolver lts-22.

One is like this.

```
tls                          > /tmp/stack-1603e7557b70df64/tls-1.8.0/Network/TLS/Parameters.hs:417:39: error: [GHC-39999]
tls                          >     • No instance for ‘Default ValidationCache’
tls                          >         arising from a use of ‘def’
tls                          >     • In the ‘sharedValidationCache’ field of a record
tls                          >       In the expression:
tls                          >         Shared
tls                          >           {sharedCredentials = mempty,
tls                          >            sharedSessionManager = noSessionManager, sharedCAStore = mempty,
tls                          >            sharedValidationCache = def, sharedHelloExtensions = []}
tls                          >       In an equation for ‘def’:
tls                          >           def
tls                          >             = Shared
tls                          >                 {sharedCredentials = mempty,
tls                          >                  sharedSessionManager = noSessionManager, sharedCAStore = mempty,
tls                          >                  sharedValidationCache = def, sharedHelloExtensions = []}
tls                          >     |
tls                          > 417 |             , sharedValidationCache = def
tls                          >     |                                       ^^^
```

The other is like this.

```
yesod-form-multi             > /home/runner/work/yesod/yesod/yesod-form-multi/Yesod/Form/MultiInput.hs:38:0: error:
yesod-form-multi             >      warning: extra tokens at end of #ifdef directive
yesod-form-multi             >        38 | #ifdef MIN_VERSION_shakespeare(2,0,18)
yesod-form-multi             >           | 
yesod-form-multi             >    |
yesod-form-multi             > 38 | #ifdef MIN_VERSION_shakespeare(2,0,18)
yesod-form-multi             >    | ^
```

Do you think you can fix these? Happy to support you on this.

Oleksandr-Zhabenko:
As for the CPP preprocessing macros, I think it is better to delete the ifdef part and use just the following if-else.

Oleksandr-Zhabenko:
I have just tried to resolve the non-building cases, this led to updating many dependincies in the stack.yaml file on the project level. But still, the connection package is not compiled because it has no appropriate version, it is needed to contact the @vincenthz to provide some workaround for this.

psibi:
> And the connection-0.3.1 is the last available version, and their source repository is archived on GitHub by the author.
https://github.com/vincenthz/hs-connection

Can you switch to this maintained fork instead: https://hackage.haskell.org/package/crypton-connection ?

Oleksandr-Zhabenko:
> Can you switch to this maintained fork instead: https://hackage.haskell.org/package/crypton-connection ?
connection is a hidden dependency of one of the dependencies of the repository. It is even hard to find out which one use it because there are more than 1000 packages that use connection as a indirect dependency.
It seems it is needed for some dependency of the yesod-auth or yesod-auth-oauth, but I am not sure.

Oleksandr-Zhabenko:
> > And the connection-0.3.1 is the last available version, and their source repository is archived on GitHub by the author.
> > https://github.com/vincenthz/hs-connection
> 
> Can you switch to this maintained fork instead: https://hackage.haskell.org/package/crypton-connection ?

It seems to be resolved. I will now squash commits to have just one final with the needed changes.

psibi:
I see that there are still some CI errors. Probably we can remove the old resolver. I think having CI for the last 3 GHC major versions should be good enough.

Oleksandr-Zhabenko:
> I see that there are still some CI errors. Probably we can remove the old resolver. I think having CI for the last 3 GHC major versions should be good enough.

I have the same opinion. The fails are for the lts-16 resolver only, and is connected with base-4.13* and basement-0.14* series instead of the provided newer versions.

Oleksandr-Zhabenko (commented):
Now it builds.

Vlix:
I think that most of what's in this PR is now rendered unnecessary after #1861 .
And I don't think you need to add the `>= 0.8` constraint to the `data-default` dependency. The only thing that has happened through the years is changes to instances. The API has always been the same, so if you're only using the type class and its method `def`, a plain `data-default` without constraints is all you need. (or if you really want one, use `data-default < 1`)

Vlix:
When I'm done with #1862, it should include the switch to `data-default`. Sorry for kind of high-jacking this PR, it kind of just happened :sweat_smile:

PR #1864, opened 2025-03-20
When showing HandlerContent, include a content snippet

dtpowl:
When possible, shows a snippet of the underlying `Content` when `show`ing (or `displayException`ing) an `HCContent`.

----

After submitting your PR:


yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -306,6 +306,20 @@ data Content = ContentBuilder !BB.Builder !(Maybe Int) -- ^ The content and opti
...
+      truncated = take maxLength (show builder)
+      excess = case maybeLength of
+        (Just length) -> length - maxLength
+        Nothing -> 0
+    in case (excess > 0) of
+      True -> truncated ++ "... (+" ++ show excess ++ ")"
+      False -> truncated
+contentToTruncatedString (ContentSource _) _ = "ContentSource"
+contentToTruncatedString (ContentFile _ _) _ = "ContentFile"
+contentToTruncatedString (ContentDontEvaluate _) _ = "ContentDontEvaluate"
```
parsonsmatt, line 321: This function is exported. Can we add a doc string with @since annotation?

```suggestion
-- | doc me up boss
--
-- @since 1.6.28.0
contentToTruncatedString :: Content -> Int -> String
contentToTruncatedString (ContentBuilder builder maybeLength) maxLength =
    let
      truncated = take maxLength (show builder)
      excess = case maybeLength of
        (Just length) -> length - maxLength
        Nothing -> 0
    in case (excess > 0) of
      True -> truncated ++ "... (+" ++ show excess ++ ")"
      False -> truncated
contentToTruncatedString (ContentSource _) _ = "ContentSource"
contentToTruncatedString (ContentFile _ _) _ = "ContentFile"
contentToTruncatedString (ContentDontEvaluate _) _ = "ContentDontEvaluate"
```

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -306,6 +306,20 @@ data Content = ContentBuilder !BB.Builder !(Maybe Int) -- ^ The content and opti
              | ContentFile !FilePath !(Maybe FilePart)
              | ContentDontEvaluate !Content
 
+contentToTruncatedString :: Content -> Int -> String
+contentToTruncatedString (ContentBuilder builder maybeLength) maxLength =
+    let
+      truncated = take maxLength (show builder)
```
parsonsmatt, line 312: hmm. `show` is probably not what we want here, but it may be. Calling `show` on a string-ish type usually will add quotes around it, so it can be copy/pasted into Haskell code.

```
λ> import Data.ByteString.Builder
λ> show ("hello" :: Builder)
"\"hello\""
λ> print ("hello" :: Builder)
"hello"
λ> putStrLn "hello"
hello
λ> print ("hello" :: String)
"hello"
```

(`print` is `putStrLn . show`)

Likewise, `show "\n" == "\"\\n\""`

But, what we have is a bunch of bytes. Do we interpret them as ASCII and do `Data.ByteString.Char8.unpack :: ByteString -> String`? Or do we treat them as UTF8 and do `Data.Text.Encoding.decodeUtf8` ?

yesod-core/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # ChangeLog for yesod-core
 
+## 1.6.27.1
```
parsonsmatt, line 3: Since we add a new function to public API, this is a minor version bump

```suggestion
## 1.6.28.0
```

Some may argue that changing behavir of `Show` is also a breaking change, but I am not inclined to agree with that.

yesod-core/src/Yesod/Core/Types.hs
dtpowl: If we accepted `TypedContent` here instead of `Content`, we should be able to infer the encoding. If it's ASCII or UTF-8 or some other text encoding, the right thing to do will be obvious. And if it's _not_ text, maybe we shouldn't include the snippet at all.

parsonsmatt: Great call

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -306,6 +308,24 @@ data Content = ContentBuilder !BB.Builder !(Maybe Int) -- ^ The content and opti
...
+--
+-- @since 1.6.28.0
+contentToTruncatedString :: Content -> I.Int64 -> String
+contentToTruncatedString (ContentBuilder builder maybeLength) maxLength =
+    let
+      truncated = (T.unpack . Data.Text.Encoding.decodeUtf8) $ L.toStrict $ L.take maxLength (BB.toLazyByteString builder)
+      excess = case maybeLength of
+        (Just length) -> length - (fromIntegral maxLength)
+        Nothing -> 0
+    in case (excess > 0) of
```
schoettl, line 322: Just my curiosity, is it a code formatter that surrounds `Just length`, `excess > 0` and `fromIntegral maxLength` with parens or is it your personal preference?

dtpowl: It's less a preference and more a habit inherited from working in other languages; I'm pretty new to Haskell.

I don't really think the parens aid legibility in the examples you pointed out, so I'll probably remove them.

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -444,7 +464,11 @@ data HandlerContents =
     | HCWaiApp !W.Application
 
 instance Show HandlerContents where
-    show (HCContent status (TypedContent t _)) = "HCContent " ++ show (status, t)
+    show (HCContent status (TypedContent t c))
+      = mconcat [ "HCContent "
+                , show (status, t)
+                , contentToTruncatedString c 1000
+                ]
```
parsonsmatt, line 471: Maybe want to parens + space it

```suggestion
      = mconcat [ "HCContent "
                , show (status, t)
                , " ("
                , contentToTruncatedString c 1000
                , ")"
                ]
```

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -306,6 +308,24 @@ data Content = ContentBuilder !BB.Builder !(Maybe Int) -- ^ The content and opti
...
              | ContentDontEvaluate !Content
 
+-- | Represents a Content as a String, rendering at most a specified number of
+-- bytes of the content, and annotating it with the remaining length.
+--
+-- @since 1.6.28.0
+contentToTruncatedString :: Content -> I.Int64 -> String
+contentToTruncatedString (ContentBuilder builder maybeLength) maxLength =
+    let
+      truncated = (T.unpack . Data.Text.Encoding.decodeUtf8) $ L.toStrict $ L.take maxLength (BB.toLazyByteString builder)
```
parsonsmatt, line 318: style nit: it's nice to have a consistent flow of function application. You can do almost all `.` with a single `$`, or all `$`:

```suggestion
      truncated = T.unpack $ Data.Text.Encoding.decodeUtf8 $ L.toStrict $ L.take maxLength $ BB.toLazyByteString builder
```


```suggestion
      truncated = T.unpack . Data.Text.Encoding.decodeUtf8 . L.toStrict . L.take maxLength $ BB.toLazyByteString builder
```

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -306,6 +308,24 @@ data Content = ContentBuilder !BB.Builder !(Maybe Int) -- ^ The content and opti
              | ContentFile !FilePath !(Maybe FilePart)
              | ContentDontEvaluate !Content
 
+-- | Represents a Content as a String, rendering at most a specified number of
+-- bytes of the content, and annotating it with the remaining length.
+--
+-- @since 1.6.28.0
+contentToTruncatedString :: Content -> I.Int64 -> String
```
parsonsmatt, line 315: `Int64` is a reasonable choice here but it's a little nicer to use `Integer` or `Int` here - those are in `Prelude` and don't require imports for end users.

yesod-core/src/Yesod/Core/Types.hs
```diff
@@ -35,9 +35,11 @@ import           Data.Serialize                     (Serialize (..),
 import           Data.String                        (IsString (fromString))
 import           Data.Text                          (Text)
 import qualified Data.Text                          as T
+import           Data.Text.Encoding
 import qualified Data.Text.Lazy.Builder             as TBuilder
 import           Data.Time                          (UTCTime)
 import           GHC.Generics                       (Generic)
+import qualified GHC.Int                            as I
```
parsonsmatt, line 42: `GHC.Int` is kind of an internal module - generally preferable to import from `Data.Int` which is less GHC internals-y

yesod-core/src/Yesod/Core/Types.hs
dtpowl: Agreed, but `take` in `Data.ByteString.Lazy` actually [requires](https://hackage.haskell.org/package/bytestring-0.12.2.0/docs/Data-ByteString-Lazy.html#g:13) Int64.

yesod-core/src/Yesod/Core/HandlerContents.hs
```diff
@@ -0,0 +1,37 @@
+module Yesod.Core.HandlerContents
```
dtpowl, line 1: I broke this type definition out into a separate file to avoid a circular dependency. Maybe someone else can see a better way to solve this?

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
 simpleContentType :: ContentType -> ContentType
 simpleContentType = fst . B.break (== _semicolon)
 
+decoderForCharset :: Maybe B.ByteString -> L.ByteString -> TL.Text
+decoderForCharset (Just encodingSymbol)
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "utf-8")        = LE.decodeUtf8With EE.lenientDecode
```
dtpowl, line 245: See the [IANA documentation](https://www.iana.org/assignments/character-sets/character-sets.xhtml) for the character set symbols. I didn't handle synonyms here.

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "US-ASCII")     = TL.fromStrict . fst . decodeASCIIPrefix . B.toStrict
```
parsonsmatt, line 246: If you enable `OverloadedStrings` then you can write:

```suggestion
  | encodingSymbol == "US-ASCII"     = TL.fromStrict . fst . decodeASCIIPrefix . B.toStrict
```

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "latin1")       = LE.decodeLatin1
```
parsonsmatt, line 247: i generally do not recommend using a formatter that aligns `=` like that - very sensitive to alignment breaking in other cases

instead, indenting after the `=` helps to have alignment on the expressions

```suggestion
  | encodingSymbol == (encodeUtf8 $ T.pack $ "latin1") = 
      LE.decodeLatin1
```

this is one of those things where the very nice aesthetics of Mathy Lookin Haskell Code don't play super nice with code-on-computer (vs code-on-paper)

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+
+textDecoderFor :: ContentType -> L.ByteString -> Maybe TL.Text
+textDecoderFor ct =
+  let packString  = (encodeUtf8 . T.pack)
+      (t, params) = NWP.parseContentType ct
+      charset     = lookup (packString "charset") params
+      typeIsText  = B.isPrefixOf (packString "text") t             ||
+                    B.isPrefixOf (packString "application/json") t ||
+                    B.isPrefixOf (packString "application/rss")  t ||
+                    B.isPrefixOf (packString "application/atom") t
```
parsonsmatt, line 263: Similar note re alignment - generally nicer to have operator first

```suggestion
      typeIsText = 
          B.isPrefixOf (packString "text") t
              || B.isPrefixOf (packString "application/json") t
              || B.isPrefixOf (packString "application/rss")  t
              || B.isPrefixOf (packString "application/atom") t
```

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "Shift_JIS")    = TL.pack . Enc.decodeLazyByteString Enc.ShiftJIS
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "Windows-31J")  = TL.pack . Enc.decodeLazyByteString Enc.CP932
+  | otherwise = LE.decodeUtf8With EE.lenientDecode
+decoderForCharset Nothing = LE.decodeUtf8With EE.lenientDecode
+
+textDecoderFor :: ContentType -> L.ByteString -> Maybe TL.Text
+textDecoderFor ct =
+  let packString  = (encodeUtf8 . T.pack)
+      (t, params) = NWP.parseContentType ct
+      charset     = lookup (packString "charset") params
```
parsonsmatt, line 259:
```suggestion
      (t, params) = 
          NWP.parseContentType ct
      charset = 
          lookup (packString "charset") params
```

more diff-friendly way to get alignment on the expressions

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "latin1")       = LE.decodeLatin1
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "GB18030")      = TL.pack . Enc.decodeLazyByteString Enc.GB18030
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "windows-1251") = TL.pack . Enc.decodeLazyByteString Enc.CP1251
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "Shift_JIS")    = TL.pack . Enc.decodeLazyByteString Enc.ShiftJIS
+  | encodingSymbol == (encodeUtf8 $ T.pack $ "Windows-31J")  = TL.pack . Enc.decodeLazyByteString Enc.CP932
+  | otherwise = LE.decodeUtf8With EE.lenientDecode
+decoderForCharset Nothing = LE.decodeUtf8With EE.lenientDecode
+
+textDecoderFor :: ContentType -> L.ByteString -> Maybe TL.Text
+textDecoderFor ct =
```
parsonsmatt, line 256: This is point-free - we can make the below a *bit* more legible if we accept the parameter explicitly:

```suggestion
textDecoderFor :: ContentType -> L.ByteString -> Maybe TL.Text
textDecoderFor ct bytes =
```

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+      (t, params) = NWP.parseContentType ct
+      charset     = lookup (packString "charset") params
+      typeIsText  = B.isPrefixOf (packString "text") t             ||
+                    B.isPrefixOf (packString "application/json") t ||
+                    B.isPrefixOf (packString "application/rss")  t ||
+                    B.isPrefixOf (packString "application/atom") t
+      decoder = decoderForCharset charset
+  in if isJust charset || typeIsText
+     then Just <$> decoder
+     else \_ -> Nothing
```
parsonsmatt, line 267: If we take `bytes` above, then this is:

```suggestion
  in if isJust charset || typeIsText
     then Just (decoder bytes)
     else Nothing
```

I'm also a little - hmm - about `isJust`. We're in `Maybe`. We can use the `Monad` and `Alternative` instances here.

```haskell
if isJust charset
  then Just decoder
  else Nothing

-- ==> is the same as

do
  _ <- charset
  pure decoder
```

Now, we also want to get the `typeIsText`. We have a function `guard :: (Alternative m) => Bool -> m ()` which looks like:

```haskell
guard condition = 
  if condition
    then pure () -- aka Just ()
    else empty   -- aka Nothing
```

So we can combine that with `<|>` (which "lifts" the notion of `||` from boolean "or" to an aplicative/alternative "or") -

```haskell
do
  void charset <|> guard typeIsText
  pure $ decoder bytes
```

And if we're doing `do`, then we can put the `let` in `do` instead of as an expr:

```haskell
let
    foo = bar
in do
    blah

-- ===>

do
    let foo = bar
    blah
```

yesod-core/src/Yesod/Core/HandlerContents.hs
```diff
@@ -0,0 +1,37 @@
...
+
+data HandlerContents =
+      HCContent !H.Status !TypedContent
+    | HCError !ErrorResponse
+    | HCSendFile !ContentType !FilePath !(Maybe W.FilePart)
+    | HCRedirect !H.Status !Text
+    | HCCreated !Text
+    | HCWai !W.Response
+    | HCWaiApp !W.Application
+instance Show HandlerContents where
```
parsonsmatt, line 23:
```suggestion
data HandlerContents =
      HCContent !H.Status !TypedContent
    | HCError !ErrorResponse
    | HCSendFile !ContentType !FilePath !(Maybe W.FilePart)
    | HCRedirect !H.Status !Text
    | HCCreated !Text
    | HCWai !W.Response
    | HCWaiApp !W.Application
    
instance Show HandlerContents where
```

yesod-core/src/Yesod/Core/HandlerContents.hs
```diff
@@ -0,0 +1,37 @@
...
+                , fromMaybe "" $ TL.unpack <$> typedContentToSnippet tc 1000
+                , ")"
+                ]
+    show (HCError e) = "HCError " ++ show e
+    show (HCSendFile ct fp mfp) = "HCSendFile " ++ show (ct, fp, mfp)
+    show (HCRedirect s t) = "HCRedirect " ++ show (s, t)
+    show (HCCreated t) = "HCCreated " ++ show t
+    show (HCWai _) = "HCWai"
+    show (HCWaiApp _) = "HCWaiApp"
+instance Exception HandlerContents
```
parsonsmatt, line 37:
```suggestion
    show (HCWai _) = "HCWai"
    show (HCWaiApp _) = "HCWaiApp"
    
instance Exception HandlerContents
```

yesod-core/src/Yesod/Core/HandlerContents.hs
parsonsmatt: I think this is fine - just make sure `HandlerContents` is being properly re-exported from the original module in which it is defined.

You'll need to modify the `yesod-core.cabal` file and add this as either an `exposed-modules` (which makes it public so you'll want to add a CHANGELOG entry + some haddocks here) or `other-modules` (in which case it is private and you don't need to do quite so much ((though having an `@since` tag in some moduel docs is nice for the developer trawlin code)))

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+
+contentToSnippet :: Content -> (L.ByteString -> Maybe TL.Text) -> I.Int64 -> Maybe TL.Text
+contentToSnippet (ContentBuilder builder maybeLength) decoder maxLength = do
+  truncatedText <- decoder . L.take maxLength $ BB.toLazyByteString builder
+  pure $ truncatedText <> (TL.pack excessLengthString)
+  where
+    excessLength = fromMaybe 0 $ (subtract $ fromIntegral maxLength) <$> maybeLength
+    excessLengthString = case excessLength > 0 of
+      False -> ""
+      True -> "...+ " <> (show excessLength)
```
parsonsmatt, line 277: We can return `L.ByteString` here and leave the decoding to callsites. That simplifies our signature and use a bit.

```suggestion
contentToSnippet :: Content -> I.Int64 -> Maybe L.ByteString
contentToSnippet (ContentBuilder builder maybeLength) maxLength = do
  truncatedText <- decoder . L.take maxLength $ BB.toLazyByteString builder
  pure $ truncatedText <> (TL.pack excessLengthString)
  where
    excessLength = fromMaybe 0 $ (subtract $ fromIntegral maxLength) <$> maybeLength
    excessLengthString = case excessLength > 0 of
      False -> ""
      True -> "...+ " <> (_f excessLength)
```

For `_f`, consider [`Data.ByteStrying.Lazy.Char8`](https://hackage.haskell.org/package/bytestring-0.12.2.0/docs/Data-ByteString-Lazy-Char8.html) which can `pack :: [Char] -> ByteString` or for supreme efficiency, using [`intDec :: Int -> Builder`](https://hackage.haskell.org/package/bytestring-0.12.2.0/docs/Data-ByteString-Builder.html#v:intDec) for constructing this, and then `BB.toLazyByteString`.

yesod-core/src/Yesod/Core/Content.hs
```diff
@@ -226,6 +240,53 @@ typeOctet = "application/octet-stream"
...
+contentToSnippet (ContentFile _ _) _ _ = Nothing
+contentToSnippet (ContentDontEvaluate _) _ _ = Nothing
+
+-- | Represents TypedContent as a String, rendering at most a specified number of
+-- bytes of the content, and annotating it with the remaining length. Returns Nothing
+-- if the content type indicates the content is binary data.
+--
+-- @since 1.6.28.0
+typedContentToSnippet :: TypedContent -> I.Int64 -> Maybe TL.Text
+typedContentToSnippet (TypedContent t c) maxLength = contentToSnippet c (textDecoderFor t) maxLength
```
parsonsmatt, line 288: If we extract the decoding responsibility, then we have:

```suggestion
typedContentToSnippet (TypedContent t c) maxLength = textDecoderFor t $ contentToSnippet c maxLength
```

parsonsmatt (commented):
lots of comments, many of which are just on style/idiomatic Haskell

yesod-core/src/Yesod/Core/Content.hs
dtpowl: This makes sense, but I'll probably rename this function while I'm at. Obviously there's no behavioral difference, but `textDecoderFor` _sounds_ like a unary function that accepts a content type and returns a decoder; `decodeTextForContentType` _sounds_ like a two-place function that accepts both a content type and bytes.

parsonsmatt:
This got smuggled i nwith yesod-core-1.7.0.0

PR #1875, opened 2025-06-07
Ordering pragmas and removing trailing whitespace

Vlix:
Split off from #1862 

---

- Ran `fourmolu` over the code base
- Only picked the pragma changes
- Went through all files manually where `fourmolu` gave "haddock parse errors"
- Let VSCode automatically remove whitespaces (and enforce trailing newlines) by saving the files.


schoettl (approved):
I viewed all changes so far. Looks good to me.

PR #1876, opened 2025-06-08
lots of CPP and cleanup of 'Data.Monoid/Semigroup' stuff

Vlix:
Mostly found all the imports of 'Data.Monoid' and 'Data.Semigroup' and removed as much as I can without breaking it.
Used CPP for backwards compatibility where needed.

Thought about using CPP to avoid deprecation warning of 'parseMonad', but defining a helper function seemed way more straightforward.

### DIFF
--- a/demo/streaming/streaming.hs
+++ b/demo/streaming/streaming.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE TemplateHaskell #-}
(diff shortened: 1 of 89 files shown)


Vlix:
Also a split-off from #1862

Vlix:
Noticed pragmas were overused, so at least wanted to remove CPP pragmas that were obviously superfluous.

... tests are failing because of `stack`, it seems :(
Will have to be rerun.

demo/streaming/streaming.hs
```diff
@@ -6,7 +7,9 @@
 import Data.Conduit
 import qualified Data.Conduit.Binary as CB
 import Control.Concurrent.Lifted (threadDelay)
-import Data.Monoid ((<>))
+#if !MIN_VERSION_base(4,11,0)
+import Data.Semigroup ((<>))
+#endif
```
parsonsmatt, line 12: `base-4.11` corresponds to GHC 8.4. We currently have a `base` lower bound on `4.10` for GHC 8.2. 8.4 was released in March 2018, almost 7 years ago, well outside of any reasonable need to preserve support.

Honestly I'm in favor of bumping the `base` lower bound and deleting all this CPP.

Vlix: @parsonsmatt 
Should I also bump the (patch) version of every package I set to `base >= 4.11`? (And add to the Changelog?)

parsonsmatt (approved):

PR #1877, opened 2025-06-16
Removing and fixing imports and other warnings

Vlix:
Last part of #1862 

---

- Mostly import removal / adjusting.
- refactors to not get 'incomplete-uni-patterns' warning
- built with some extra warnings to catch more spots to tweak (e.g. `-Woperator-whitespace`)

### DIFF
--- a/demo/streaming/streaming.hs
+++ b/demo/streaming/streaming.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes #-}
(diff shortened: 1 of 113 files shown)


demo/streaming/streaming.hs
```diff
@@ -3,10 +3,10 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 
+import Yesod.Core
```
Vlix, line 6: Was removed erroneously in the first batch of pragma orderings

parsonsmatt (approved):
Looks good to me!

PR #1881, opened 2025-07-08
Allow disabling of resources generation

parsonsmatt:
This PR adds a setting to `RouteOpts` to allow users to control whether or not the code generation includes the `resourcesSite :: [ResourceTree String]` value.

In our web app, the generated TH code takes up about 40,000 lines. A good 1/3 of this is the `resourcesApp` value. But we already generate this in another file with `parseRouteFile`, so it is redundant.

On the timing impact: before this PR, our `P50(duration_s)` for compiling the module that only does `mkYesodData` is 6.456s. We have just under 2k routes. After, {i will fill in after running the test and collecting data}.

After submitting your PR:

### DIFF
--- a/yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
+++ b/yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
@@ -9,12 +9,14 @@ module Yesod.Routes.TH.RenderRoute
     , mkRouteCons
     , mkRouteConsOpts
     , mkRenderRouteClauses
+    , shouldCreateResources
 
     , RouteOpts
     , defaultOpts
     , setEqDerived
     , setShowDerived
     , setReadDerived
+    , setCreateResources
     ) where
 
 import Yesod.Routes.TH.Types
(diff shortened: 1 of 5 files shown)


yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -65,13 +78,35 @@ setShowDerived b rdo = rdo { roDerivedShow = b }
...
 -- @since 1.6.25.0
 instanceNamesFromOpts :: RouteOpts -> [Name]
-instanceNamesFromOpts (MkRouteOpts eq shw rd) = prependIf eq ''Eq $ prependIf shw ''Show $ prependIf rd ''Read []
+instanceNamesFromOpts MkRouteOpts {..} = prependIf roDerivedEq ''Eq $ prependIf roDerivedShow ''Show $ prependIf roDerivedRead ''Read []
     where prependIf b = if b then (:) else const id
 
+-- |
+--
+-- @since 1.6.28.0
+
```
L0neGamer: what is since, here?

parsonsmatt: `@since` declarations show up in the generated Haddock as the version that the term was introduced

Issue #1880, opened 2025-07-07
Allow generating nested routes separately from the main routes


parsonsmatt:
We've got around 2k routes and the call to `mkYesodData` is becoming a problem for us.

Fortunately, there's a relatively easy touch point to significantly improve this: allow nested routes to be generated separately.

Currently, we use the `TemplateHaskell` code to parse the routes in a standalone module. This module is imported, and we refer to the `routes :: [ResourceTree String]` when we call `mkYesodData "WebApp" routes`.

Each new route adds five things to this file: 

1. An entry in `resourcesWebApp` - which is a copy/paste from `routes` as input.
2. A constructor in `Route WebApp` (or, if it is nested, a new constructor in a separate datatype)
3. A parse route case
4. A render route case
5. a `RouteAttrs` case

While we can improve other things, the biggest win will be in allowing the Yesod generate code to refer to prior work in other modules.

I suggest a function `mkYesodNestedRoute :: String -> String -> [ResourceTree String] -> DecsQ`, used like:

```haskell
mkYesodNestedRouteData "WebApp" "AdminR" routes
```

That finds the `AdminR` route directory and creates the relevant data and functions for it.

Then, in `mkYesodData`, we do a lookup on the presence of the type. If `AdminR` already exists, then we don't bother generating it: we refer to it. Same with the other cases: if we have a function `AdminR -> Set Text` for route attributes, then we can delegate to it.

jezen:
Just thinking out loud — are subsites not a potential solution to this problem?

parsonsmatt:
I think they are a potential solution, but they require a pretty significant code migration app-side, and the difficulty/UX of writing subsite code is worse than writing regular code. The changes I'm planning will allow you to break things up significantly more easily and incrementally.

jezen:
Yeah, that makes sense. We're breaking apart our application with subsites, but there's certainly a learning curve, _and_ we've had to upstream some changes to make it actually work.

TeofilC:
Regarding `mkYesodData` being slow, I looked into this a while ago and I had some ideas that I never ended up implementing: https://github.com/yesodweb/yesod/issues/1889

PR #1887, opened 2025-07-30
Code generation and delegation to nested routes

parsonsmatt:
# Nested Route Separation

We have almost 2,000 total routes in our codebase now, with 1,100 of them on the top-level.
We have been leveraging nested routes to organize the routes and make certain things easier - namely, reducing the overall size of the `Route WebApp` which makes pattern matching more efficient in compilation.

However, compiling `mkYesodData` for our application is a large bottleneck.
Compilation takes a significant amount of time and much work must be redone, even when most of the routing information could have been saved.
One thing that would significantly help is the ability to separate out the sub-route datatypes, generate the instance for these separately, and refer to them elsewhere.

Another thing that would help is having a finer grained `YesodDispatch` facility - allowing us to, in tests, specify more precisely which parts of the route structure we actually need to deal with.
Currently, we need almost 7k modules in order to compile a test that refers to `YesodDispatch`, since this brings in the transitive dependencies of every web handler.
With a finer grained `YesodDispatch`, able to refer to route fragments, we'd be able to avoid depending on more of the website than we really need to.
We expect this to yield significant benefits when we move more testing infrastructure to buck2 - right now, changing anything used by any part of the web app would require running *all* tests that exercise *anything* in the webapp!
With more granular `YesodDispatch`, we gain the ability to only run tests on parts of the app that actually use the route structure in question.

---

This PR is a breaking change, and so will be `yesod-core-1.7.0.0`. When I tried this branch on our package, it required 0 modifications to upgrade - *most* of the breaking changes here are breaking "advanced power user" type functions, like the ability to customize the `MkDispatchSettings` to have highly custom route behaviors. We don't use any of those, and I have no idea how widespread their use is. I will admit to finding them *very* difficult to use and adapt, which is part of why I ended up scrapping their use in the `ParseRoute` generation (the class isn't even used).

For the most part, though, all prior behavior remains unchanged. The performance of routing may be slightly worse - `YesodDispatchNested` returns a `Maybe Application` (but with the `Request` pulled out - only the "send response" side is in the Maybe) to support fallthrough. And if you enable fallthrough, then the app will do more checking (proportional to the number of routes and nested routes) rather than early-404ing. But IMO for perf, we're better off coming up with a trie-based router than trying to optimize this one much more.

For an example of the testing facilities this provides, see https://github.com/parsonsmatt/hspec-yesod/pull/7 - particularly, [this file](https://github.com/parsonsmatt/hspec-yesod/pull/7/files#diff-cb2b506f90802fec4d57dbac12a6d60f3e34219f5b0e8781cc5d12da89f8b3bd) has an introduction to how the tests are separated and defined.

---

Fix for #1880 

After submitting your PR:


yesod-core/src/Yesod/Routes/Class.hs
```diff
@@ -16,10 +20,49 @@ class Eq (Route a) => RenderRoute a where
...
+-- | This class acts as a delegation class for 'RenderRoute' on nested
+-- route fragments.
+--
+-- @since 1.6.28.0
+class Eq a => RenderRouteNested a where
+    -- | Render the fragment of the route. To form a complete route, you'll
+    -- need to `mappend` this with the result from the parent `renderRoute`
+    -- or `renderRouteNested` call.
+    --
+    -- @since 1.6.28.0
```
parsonsmatt: I think this requirement is questionable. The instance *can* know about the entire path, and so perhaps we should bake this into the implementation. Would probably make testing subroutes easier.

yesod-core/src/Yesod/Routes/Class.hs
```diff
@@ -16,10 +20,49 @@ class Eq (Route a) => RenderRoute a where
...
+    parseRouteNested
+        :: ([Text], [(Text, Text)])
+        -- ^ The path of the URL split on forward slashes, and a list of query parameters with their associated value.
+        --
+        -- Unlike for 'parseRoute', this will not include URL fragments as
+        -- part of parent routes. It is expected that parents will take the
+        -- resulting @a@ and construct the parent from that.
+        -> Maybe a
```
parsonsmatt: Meanwhile, this... hmm. Seems like `renderRouteNested` and `parseRouteNested` should have inverse behavior. So if `renderRouteNested` knows about parent stuff, then `parseRouteNested` maybe also should work on a complete route? 

Would be kind of weird. So maybe this *is* wrong. Maybe the testing facilities should "just" use the route/dispatching behavior that they're concerned with. But then the testing facilities do diverge slightly from the real world router/behavior...

yesod-core/src/Yesod/Core/Internal/TH.hs
```diff
@@ -253,9 +257,20 @@ mkYesodGeneralOpts opts appCxt' namestr mtys isSub f resS = do
...
-    dispatchDec    <- mkDispatchInstance site appCxt f res
-    parseRoute <- mkParseRouteInstance appCxt site res
+    routeAttrsDec  <-
+        case roFocusOnNestedRoute opts of
+            Nothing ->
+                pure <$> mkRouteAttrsInstance appCxt site res
+            Just target ->
+                mkRouteAttrsInstanceFor appCxt (ConT (mkName target)) target res
+
+    dispatchDec    <- mkDispatchInstance opts (roFocusOnNestedRoute opts) site appCxt f res
```
lf-:
```suggestion
    dispatchDec <- mkDispatchInstance opts (roFocusOnNestedRoute opts) site appCxt f res
```

we probably should get a code formatter though...

yesod-core/test/Hierarchy.hs
```diff
@@ -9,6 +9,8 @@
 {-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE ViewPatterns #-}
 
+{-# OPTIONS_GHC -Wno-orphans #-}
```
lf-: testing note for a future PR: I'm wondering if it might be helpful to have snapshot/golden tests for the TH output of running the generator; not to catch bugs, but for documentation of how it works.

One thing I had trouble with reading this PR is understanding what the generated code looks like, and I think this could help with that.

lf- (commented):
I don't have enough context to properly review this, but I had some thoughts/comments as I was reading it as an outsider.

Having some kind of checked-in golden files for documentation purposes would be extremely helpful in understanding the yesod/persistent TH in general, I think; this was confusing even when I was originally poking at Yesod years ago.

yesod-core/test/Hierarchy.hs
parsonsmatt: It doesn't help that the generator is highly indirected and the generated code is really difficult to read! Part of my work here is separating the parser and dispatch code so the parse code at least can be straightforward. The dispatch code may need to remain complex, but I think I can make it nicer - the `MkDispatchSettings` record is exported fully open, so it's gonna be a breaking change major bump, *but* I think this is going to be very low impact, and anyone truly impacted can help fix it up.

yesod-core/src/Yesod/Routes/Class.hs
parsonsmatt: Currently redesigning this, actually - `renderRouteNested` will need to get the parent dyns but will otherwise create the entire route itself.

L0neGamer:
Hi Matt! @jezen asked me to test this out on our codebase, and from what I can tell there is some kind of error when we use mkYesodSubData.

Here's an example excerpt:

```hs
mkYesodSubData "(SubsiteClass subsite master) => SubsiteData subsite"
  $(makeRelativeToProject "config/routes.yesodroutes" >>= parseRoutesFile)
```

results in many errors of the format:

```
src/Subsite/Data.hs:158:1: error: [GHC-53634]
    Out of scope type variable ‘subsite’ in the RHS of a family instance.
      All such variables must be bound on the LHS.
    |
158 | mkYesodSubData "(SubsiteClass subsite master) => SubsiteData subsite"
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

I can't immediately spot why in your PR.

parsonsmatt:
OK, I think we have incorporated your test case- mind giving it another go? Thanks for your patience 🙏🏻

yesod-core/src/Yesod/Core/Class/Dispatch.hs
```diff
@@ -1,26 +1,150 @@
...
     yesodDispatch :: YesodRunnerEnv site -> W.Application
 
+-- | This class enables you to dispatch on a route fragment without needing
+-- to know how to dispatch on the entire route structure. This allows you
+-- to break up route generation into multiple files.
+--
+-- For details on use, see 'setFocusOnNestedRoute'.
+--
+-- @since 1.7.0.0
+class RenderRouteNested a => YesodDispatchNested a where
```
parsonsmatt, line 43: This is the really important addition here. We can give instances of `YesodDispatchNested` to sub-route fragments, and that can be consumed in `hspec-yesod` to allow tests to only depend on the actual sub-route handlers that they need.

yesod-core/src/Yesod/Core/Class/Dispatch.hs
```diff
@@ -51,3 +175,50 @@ subHelper (SubHandlerFor f) YesodSubRunnerEnv {..} mroute =
...
+        status
+            | W.requestMethod env == "GET" = status301
+            | otherwise                    = status307
+
+        dest = joinPath y (resolveApproot y env) segments' []
+        dest' =
+            if S.null (W.rawQueryString env)
+                then dest
+                else dest `mappend`
+                     byteString (W.rawQueryString env)
```
parsonsmatt, line 245: This implementation is pretty much the same as the prior `toWaiAppYre` - the main difference being that we call `yesodDispatchNested` instead of `yesodDispatch`. Notably, for a `Route Site`, this is exactly equivalent.

yesod-core/src/Yesod/Core/Internal/TH.hs
```diff
@@ -212,35 +192,26 @@ mkYesodGeneral = mkYesodGeneralOpts defaultOpts
...
                    -> [ResourceTree String]
                    -> Q([Dec],[Dec])
 mkYesodGeneralOpts opts appCxt' namestr mtys isSub f resS = do
-    let appCxt = fmap (\ctxs ->
-            case ctxs of
-                c:rest ->
-                    foldl' (\acc v -> acc `AppT` fst (nameToType v)) (ConT $ mkName c) rest
-                [] -> error $ "Bad context: " ++ show ctxs
-          ) appCxt'
+    appCxt <- buildAppCxt appCxt'
```
parsonsmatt, line 287: bit of a drive-by refactor to use `fail` instead of `error` + extract the shared logic into a helper.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -1,25 +1,52 @@
...
 {-# LANGUAGE ViewPatterns #-}
 
 module Yesod.Routes.TH.Dispatch
     ( MkDispatchSettings (..)
     , mkDispatchClause
     , defaultGetHandler
+    , SDC(..)
+    , mkDispatchInstance
+    , mkNestedDispatchInstance
+    , mkNestedSubDispatchInstance
```
parsonsmatt: This module is pretty much where the worst of the dragons be.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -30,8 +57,32 @@ data MkDispatchSettings b site c = MkDispatchSettings
...
+    -- @''YesodDispatchNested@ for top-level dispatch,
+    -- @''YesodSubDispatchNested@ for subsite dispatch.
+    --
+    -- @since 1.7.0.0
+    , mdsNestedDispatchFn :: Name
+    -- ^ The function to call for nested dispatch delegation.
+    -- @'yesodDispatchNested@ for top-level dispatch,
+    -- @'yesodSubDispatchNested@ for subsite dispatch.
+    --
+    -- @since 1.7.0.0
```
parsonsmatt, line 73: These are essentially the three biggest things- supporting route fallthrough (necessary to support lots of route separation), allowing us to dispatch to a separate class, and the dispatch function name.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -62,84 +166,139 @@ mkDispatchClause MkDispatchSettings {..} resources = do
...
-        [VarP envName, VarP reqName]
-        (NormalB $ helperE `AppE` pathInfo)
-        [FunD helperName $ clauses ++ [clause404']]
+    pure
+        ( childNames
+        , Clause
+            [VarP envName, VarP reqName]
+            (NormalB $ helperE `AppE` pathInfo)
+            [FunD helperName $ clauses ++ [clause404']]
+        )
```
parsonsmatt: This pattern is probably cleaner as a `WriterT` or `StateT` to accumulate.

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
```diff
@@ -210,56 +219,86 @@ generateParseRouteClausesInline
     -> ResourceTree a
     -> Q [Clause]
 generateParseRouteClausesInline accPats wrap resourceTree =
+    pure $ evalState (buildInlineParseClauses accPats wrap resourceTree) 0
+
+-- | A deterministic fresh-name supply: a monotonically increasing 'Int'
+-- counter, each value yielding a distinct name. Used in place of 'Q'\'s
+-- 'newName' so 'buildInlineParseClauses' can be pure.
+freshName :: String -> State Int Name
+freshName base = state $ \n -> (mkName (base ++ show n), n + 1)
```
L0neGamer: could just use a `Quote m` constraint and then be able to swap between a custom `State Int` wrapper and Q without having much more machinery.

L0neGamer (commented):
Good news is that this compiles with our local sites with only changes to our flake.nix.

I would love to review this but it is entirely overwhelming; +7k/-1k is hard enough even without mixing in merges from the master branch and the back and forths on implementation.

How would you recommend I review this?

yesod-core/ChangeLog.md
```diff
@@ -1,5 +1,25 @@
 # ChangeLog for yesod-core
 
+## 1.7.0.0
```
L0neGamer, line 3: The changelog is entirely overbloated with internal details

yesod-core/src/Yesod/Core/Class/Dispatch.hs
```diff
@@ -1,26 +1,156 @@
...
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+{-# OPTIONS_GHC -Wredundant-constraints #-}
```
L0neGamer: why is this needed?

yesod-core/test/YesodCoreTest/NestedDispatch/Parent0R/Child0R.hs
```diff
@@ -0,0 +1,64 @@
...
+module YesodCoreTest.NestedDispatch.Parent0R.Child0R where
+
+import Data.Text (Text)
+import qualified Data.Text as Text
+import Yesod.Core.Handler
+import YesodCoreTest.NestedDispatch.Resources
+import Yesod.Core
+import qualified Network.Wai as W
+
+--mkYesodDataOpts (setFocusOnNestedRoute (Just "Child0R") defaultOpts) "App" nestedDispatchResources
```
L0neGamer: lot of commented code here; is this also meant to be split?

yesod-core/test/YesodCoreTest/NestedDispatch/Runtime.hs
```diff
@@ -0,0 +1,454 @@
...
+import YesodCoreTest.NestedDispatch.Resources
+import YesodCoreTest.NestedDispatch.NestR (NestR(..))
+import YesodCoreTest.NestedDispatch.InnerR (InnerR(..))
+import YesodCoreTest.NestedDispatch.ParentR (ParentR(..))
+import YesodCoreTest.NestedDispatch.Parent0R (Parent0R(..), Child0R(..))
+import YesodCoreTest.NestedDispatch.Parent0R.Child0R
+import YesodCoreTest.RuntimeHarness (assertRequestRaw)
+import qualified Network.HTTP.Types as H
+
+mkYesod "App" nestedDispatchResources
```
L0neGamer, line 54: So does this only generate routing for routes that don't have their own nested routing? so `OuterR` would get its own routing?

L0neGamer: I think all of these tests (and especially this one which seems to be the one to demonstrate this new feature!) need better comments, preferably not too verbose.

Maybe a heading at the top explaining what the purpose is, then each non-obvious component within.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -30,8 +57,32 @@ data MkDispatchSettings b site c = MkDispatchSettings
...
+    -- match.
+    --
+    -- @since 1.7.0.0
+    , mdsNestedDispatchClass :: Name
+    -- ^ The class to check for nested dispatch delegation.
+    -- @''YesodDispatchNested@ for top-level dispatch,
+    -- @''YesodSubDispatchNested@ for subsite dispatch.
+    --
+    -- @since 1.7.0.0
+    , mdsNestedDispatchFn :: Name
```
L0neGamer: these could probably be covered by an enum that then gets the name later right? would reduce the space of what can be put here. You already have DispatchPhase.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -40,16 +91,69 @@ data SDC = SDC
...
+-- how subsite environments are constructed.
+--
+-- @since 1.7.0.0
+data NestedDispatchConfig = NestedDispatchConfig
+    { ndcRunnerFn      :: Name
+    -- ^ Runner function: @'yesodRunner@ for top-level, @'subHelper@ for subsites
+    , ndcDispatchFn    :: Name
+    -- ^ Nested dispatch function: @'yesodDispatchNested@ or @'yesodSubDispatchNested@
+    , ndcDispatchClass :: Name
+    -- ^ Nested dispatch class: @''YesodDispatchNested@ or @''YesodSubDispatchNested@
```
L0neGamer: all of these names are similarly only one of two things

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -40,16 +91,69 @@ data SDC = SDC
...
+    , ndcSubsiteEnv    = ComposedEnv
+    }
+
 -- | A simpler version of Yesod.Routes.TH.Dispatch.mkDispatchClause, based on
 -- view patterns.
 --
+-- The function returns the 'Clause' for the dispatch, along with
+-- a @['Name']@ corresponding to the names of types that require generating
+-- instances for delegation classes. For 'ParseRoute', that's
+-- 'ParseRouteNtested'. For 'YesodDispatch', that's 'YesodDispatchNested'.
```
L0neGamer: nit: spelling Ntested

yesod-core/src/Yesod/Routes/TH/Internal.hs
```diff
@@ -0,0 +1,379 @@
...
+        TyConI (DataD _ _ vs _ _ _) -> length vs
+        TyConI (NewtypeD _ _ vs _ _ _) -> length vs
+        TyConI (TySynD _ vs _) -> length vs
+        _ -> 0
+
+-- | Look up a type by 'Name' and return it fully applied with fresh
+-- type variables. This is needed because nested route datatypes may
+-- have type parameters (e.g., @NestedR subsite@), and TH functions
+-- like 'isInstance' require fully-applied types.
+fullyApplyType :: Name -> Q Type
```
L0neGamer, line 91: is it possible that `isInstance` will fail if the type parameters remain abstract like this?

yesod-core/src/Yesod/Routes/TH/Internal.hs
```diff
@@ -0,0 +1,379 @@
...
+-- as the head of a @YesodSubDispatchNested@ instance, which expects a fully
+-- applied (kind 'Type') route. So the arities must match exactly:
+--
+--   * Too few parameters on the nested datatype (e.g. it was generated by
+--     plain 'mkYesodSubData', kind 'Type') over-applies it.
+--   * Too many leaves it partially applied (kind @Type -> ...@).
+--
+-- Either way the result is a kind error reported deep inside generated code.
+-- 'checkNestedSubArity' turns it into an actionable, attributable message
+-- (suitable for 'fail').
```
L0neGamer: floating comment

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -174,40 +333,652 @@ mkDispatchClause MkDispatchSettings {..} resources = do
...
+
+mkMDS :: (Exp -> Q Exp) -> Q Exp -> Q Exp -> MkDispatchSettings a site b
+mkMDS unwrapper runHandlerE subDispatcher = MkDispatchSettings
+    { mdsRunHandler = runHandlerE
+    , mdsSubDispatcher = subDispatcher
+    , mdsGetPathInfo = [|W.pathInfo|]
+    , mdsSetPathInfo = [|\p r -> r { W.pathInfo = p }|]
+    , mdsMethod = [|W.requestMethod|]
+    , mds404 = [|void notFound|]
+    , mds405 = [|void badMethod|]
```
L0neGamer, line 1009: are there places these are overridden? if not then maybe these shouldn't be fields, and should be inlined where relevant

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -43,7 +58,20 @@ data RouteOpts = MkRouteOpts
...
+    -- @since 1.7.0.0
     , roParameterizedSubroute :: Bool
+    , roNestedRouteFallthrough :: Bool
+    -- ^ If 'True', then a nested route will fall through if it fails to
+    -- match. If 'False', then a nested route will throw 'notFound' if it
+    -- does not match.
+    --
+    -- Default: 'False'.
+    --
+    -- @since 1.7.0.0
```
L0neGamer, line 72: these docstrings won't be visible fwiw; you should add these default descriptions to the docs for defaultOpts

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -59,8 +87,76 @@ defaultOpts = MkRouteOpts
...
+-- Then, in a module for the route specifically, we can generate the route
+-- datatype and instances for later hooking in to the main instance.
+--
+-- @
+-- module App.Routes.NestR where
+--
+-- import App.Routes.Resources
+-- import Yesod.Core
+--
+-- mkYesodOpts (setFocusOnNestedRoute (Just "NestR") defaultOptions) "App" appResources
```
L0neGamer: I don't know whether it would be a good idea to do so, but would it make more sense to have `mkNestedYesodOpts "NestR" opts ...` instead of having it be a RouteOpt?

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -114,101 +224,409 @@ instanceNamesFromOpts MkRouteOpts {..} = prependIf roDerivedEq ''Eq $ prependIf
...
-    subrouteDecTypeArgs = fmap (tyvarbndr . snd) tyargs
+-- | Generate the constructors of a route data type, with custom
+-- 'RouteOpts'.
+--
+-- The first element in the return is the list of constructors for the
+-- @Route site@ datatype. The second element is the list of 'Dec' to
+-- declare nested datatypes or class instances of the 'RenderRouteNested'
+-- for those instances.
+--
+-- @since 1.6.25.0
```
L0neGamer: another floating comment

L0neGamer (commented):
Initial review; I'll continue in future.

yesod-core/src/Yesod/Routes/TH/Internal.hs
```diff
@@ -285,32 +289,64 @@ arityMismatchMessage callSite (ArityMismatch (SubsiteName subName) (RouteName ro
...
+-- | Abstracts fresh-'Name' generation so the TH clause builders can run either
+-- hygienically in 'Q' (production, via 'newName') or deterministically in
+-- @'State' 'Int'@ (tests, via a monotonic counter). The deterministic instance
+-- uses non-hygienic 'mkName', which is fine for the pure-inspection tests but
+-- would risk shadowing in real splices — so production stays on 'Q'\'s
+-- 'newName' and only the tests pin @m@ to @'State' 'Int'@.
+--
+-- @since 1.7.0.0
+class Monad m => FreshName m where
+    freshName :: String -> m Name
```
L0neGamer: this is just `Quote` from TH

L0neGamer (commented):
Please write the changelog in a simpler way. The latest commits also add so much more code and rewrite more things.

parsonsmatt:
Hey apologies that you're seeing it while I'm still churning it! I wanted to verify that the tests and examples here were working on your codebase, but I'm not *quite* ready to review, though it's very close.

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
parsonsmatt: I like the suggestion- writing `Just` here for the vanishingly rare case that you'd unset the option feels extraneous. API is now `setFocusonNestedRoute "NestR" opts` with an `unsetFocusOnNestedRoute` to clear it.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
parsonsmatt: The old route parser used the MDS record here to also generate parser instances. This made implementing the feature here really complicated, so I decoupled them. The MDS record is exported and power users could potentially be using it- otherwise, I'd agree, just inline.

yesod-core/src/Yesod/Core/Class/Dispatch.hs
parsonsmatt: It's not, good catch

yesod-core/ChangeLog.md
parsonsmatt: Lol thanks- Claude decided to draft the changelog without me asking. I've pared it down.

yesod-core/src/Yesod/Core/Class/Dispatch.hs
```diff
@@ -51,3 +179,50 @@ subHelper (SubHandlerFor f) YesodSubRunnerEnv {..} mroute =
...
+-- Handles @cleanPath@ redirects and a terminal 404 just like 'toWaiAppYre'.
+--
+-- @since 1.7.0.0
+toWaiAppYreNested
+    :: (Yesod (ParentSite a), YesodDispatchNested a, ToParentRoute a)
+    => Proxy a
+    -> ParentArgs a
+    -> YesodRunnerEnv (ParentSite a)
+    -> W.Application
+toWaiAppYreNested proxy parentArgs yre req =
```
parsonsmatt, line 212: This is what allows `hspec-yesod` to make an `Application` that only responds to the route fragment that is being used.

yesod-core/src/Yesod/Core/Internal/TH.hs
```diff
@@ -209,53 +190,103 @@ mkYesodGeneral :: [[String]]                -- ^ Appliction context. Used in Ren
...
+            Just target ->
+                -- Apply the site's type arguments to the focused child's
+                -- constructor, matching the focused ParseRoute / RenderRoute
+                -- paths; a bare 'ConT' here is a kind error for a
+                -- parameterized site.
+                mkRouteAttrsInstanceFor
+                    appCxt
+                    (applyTyArgs (ConT (mkName target)) tyArgs)
+                    target
+                    (rfResources foundation)
```
parsonsmatt, line 327: This is the sort of general behavior pattern here. If we're focusing on a nested route, we trim to only that target. Otherwise, we operate on the whole tree.

yesod-core/src/Yesod/Core/Internal/TH.hs
```diff
@@ -268,99 +299,147 @@ mkYesodGeneralOpts opts appCxt' namestr mtys isSub f resS = do
...
+-- @FlexibleInstances@, @MultiParamTypeClasses@ and @TypeFamilies@. The nested
+-- route datatype must also declare exactly as many type parameters as the
+-- subsite has type arguments (see the arity note below).
+--
+-- @since 1.7.0.0
+mkYesodSubDispatchInstance
+    :: String                -- ^ Foundation type with optional context, e.g. @"(MyClass a) => MySub a"@
+    -> [ResourceTree String] -- ^ Resources (e.g. @resourcesMySub@)
+    -> Q [Dec]
+mkYesodSubDispatchInstance = mkYesodSubDispatchInstanceOpts defaultOpts
```
parsonsmatt, line 385: This API is required to allow you to split routes that a subsite uses. 

I half expect that subsite users will not split compilation, because subsites are another way of handling the same separation issues. But supporting the full matrix felt important.

yesod-core/src/Yesod/Core/Dispatch.hs
```diff
@@ -138,33 +200,7 @@ defaultGen = bsToInt <$> getEntropy bytes
...
-            | W.requestMethod env == "GET" = status301
-            | otherwise                    = status307
-
-        dest = joinPath y (resolveApproot y env) segments' []
-        dest' =
-            if S.null (W.rawQueryString env)
-                then dest
-                else dest `mappend`
-                     byteString (W.rawQueryString env)
+toWaiAppYre = toWaiAppYreNested (Proxy :: Proxy (Route site)) ()
```
parsonsmatt, line 202: This exercises that everything is consistent between code paths.

yesod-core/src/Yesod/Core/Handler.hs
```diff
@@ -1032,6 +1035,17 @@ instance RedirectUrl master (Route master) where
...
+instance
+    ( RedirectUrl master (Route master)
+    , ParentSite url ~ master
+    , ToParentRoute url
+    )
+  =>
+    RedirectUrl master (WithParentArgs url)
+  where
+    toTextUrl (WithParentArgs args url) =
+        toTextUrl (toParentRoute args url)
```
parsonsmatt, line 1047: This is the template for `RedirectUrl master url` generated instances that do not capture pieces. `toParentRoute` them and then `toTextUrl` them.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -30,26 +58,134 @@ data MkDispatchSettings b site c = MkDispatchSettings
     , mds405 :: Q Exp
     , mdsGetHandler :: Maybe String -> String -> Q Exp
     , mdsUnwrapper :: Exp -> Q Exp
+    , mdsNestedRouteFallthrough :: !Bool
+    -- ^ When 'True', fall through if no route matches (except in the final
+    -- case). When 'False', return 404 if the current route clause fails to
+    -- match.
+    --
+    -- @since 1.7.0.0
```
parsonsmatt, line 65: Essentially passed through from `RouteOpts`.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -30,26 +58,134 @@ data MkDispatchSettings b site c = MkDispatchSettings
...
+    -- ^ Whether the caller emits the matching per-parent nested-dispatch
+    -- instances in the same splice as this flat dispatch body; see
+    -- 'SameSpliceNestedInstances' for how this affects a parent clause.
+    -- 'mkMDS' defaults to 'NoSameSpliceNestedInstances'.
+    --
+    -- @since 1.7.0.0
     }
 
 data SDC = SDC
-    { clause404 :: Clause
```
parsonsmatt, line 36: `clause404` was unused now; since this isn't exported we can delete.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -174,40 +406,682 @@ mkDispatchClause MkDispatchSettings {..} resources = do
...
+mkDispatchInstance routeOpts master cxt tyargs unwrapper res =
+    -- Branch on the focus target via the accessor rather than a constructor
+    -- pattern, so 'RouteOpts' can stay abstract (its constructor is not
+    -- exported). 'Just target' focuses a single nested route for
+    -- module-splitting; 'Nothing' generates the full top-level instance.
+    case roFocusOnNestedRoute routeOpts of
+        Just target ->
+            mkNestedDispatchInstance routeOpts target master cxt tyargs unwrapper res
+        Nothing ->
+            mkTopLevelDispatchInstance routeOpts master cxt tyargs unwrapper res
```
parsonsmatt, line 516: Here's the split- if we're focusing, we filter and dispatch. Otherwise we carry on as usual.

yesod-core/src/Yesod/Routes/TH/Internal.hs
parsonsmatt: hmm. I would be surprised if `isInstance` found an `instance C (T Int)` if it was looking for `isInstance ''C =<< [t|T a|]`.

parsonsmatt: Claude commented below pretty confidently that I indeed would be surprised: https://github.com/yesodweb/yesod/pull/1887/changes#diff-f8cfe8596f2672887a201a293b33809bbd7e28f0bb58867b2fff6889ef348822R145-R157

Which makes sense, in hindsight. `reifyInstances ''C =<< [t| T a |]` *should* list out the `instance C (T Int)`, and if it's a `isInstance className typeArgs = not . null <$> reifyInstances className typeArgs` then we'll get it here.

yesod-core/test/Hierarchy/Nest2/NestInner.hs
```diff
@@ -0,0 +1,23 @@
+{-# language TemplateHaskell #-}
+{-# language ViewPatterns, OverloadedStrings #-}
+{-# language TypeFamilies #-}
+
+-- | @Hierarchy@ split fragment focused on the deepest nested parent
+-- @NestInner@ (under @NestR@ > @Nest2@). Emits its render\/attrs\/parse
+-- instances via @setFocusOnNestedRoute "NestInner"@ + the @*For@ helpers; the
+-- @ParseRouteNested NestInner@ instance is what "Hierarchy.Nest2" asserts
+-- against at compile time.
+module Hierarchy.Nest2.NestInner where
```
parsonsmatt, line 10: So here starts the tests- by far the largest part of this PR.

I had Claude generate tests that would cover every interaction of feature: parameterized subsites, nested routes, separate compilation, fallthrough, etc.

yesod-core/test/Hierarchy/Nest.hs
```diff
@@ -0,0 +1,22 @@
...
+import Yesod.Routes.TH
+import Hierarchy.ResourceTree
+import Hierarchy.Nest3 ()
+import Hierarchy.Nest2
+import Language.Haskell.TH
+import Hierarchy.Nest2.NestInner
+
+mkRenderRouteInstanceOpts (setFocusOnNestedRoute "NestR" defaultOpts) [] NoTyArgs (ConT ''Hierarchy) hierarchyResourcesWithType
+mkRouteAttrsInstanceFor [] (ConT ''NestR) "NestR" $ hierarchyResourcesWithType
+mkParseRouteInstanceFor "NestR" $ hierarchyResourcesWithType
```
parsonsmatt: This is really not how the public API should be used- it's internal details. `mkYesod` and `mkYesodData/Dispatch` are the better ways to go.

parsonsmatt (commented):
Woof. Huge PR. Sorry about that.

This PR *is* pretty big. But a huge component of that is a pretty good bit of new tests that exercise backwards compatibility and demonstrate behavior.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -174,40 +405,660 @@ mkDispatchClause MkDispatchSettings {..} resources = do
...
+    }
+
+-- | Parse the foundation-type string given to @mkYesod@ into its components:
+-- the type-constructor name, its type arguments, and the @=>@ class-context
+-- groups (each a class name applied to type arguments). Returns 'Left' with a
+-- parse-error message on malformed input.
+--
+-- @since 1.7.0.0
+parseYesodName :: String -> Either String (String, [String], [[String]])
+parseYesodName name = do
```
L0neGamer, line 1024: I don't think we should expose this from yesod if we can help it; if it isn't exposed, great, we can remove the docstring and maybe move it to the one module it's used in.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -30,26 +58,134 @@ data MkDispatchSettings b site c = MkDispatchSettings
...
     }
 
 data SDC = SDC
-    { clause404 :: Clause
-    , extraParams :: [Exp]
+    { extraParams :: [Exp]
     , extraCons :: [Exp]
     , envExp :: Exp
     , reqExp :: Exp
     }
```
L0neGamer, line 88: I know you didn't make it but what is SDC? A comment here would be useful.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -30,26 +58,134 @@ data MkDispatchSettings b site c = MkDispatchSettings
...
-    helperName <- newName $ "helper" ++ show suffix
+-- The function returns the dispatch 'Clause' along with a @['String']@ of the
+-- names of nested route types that require a delegation instance to be
+-- generated. For 'YesodDispatch', those are 'YesodDispatchNested' instances; for
+-- the subsite case, 'YesodSubDispatchNested'.
+--
+-- @since 1.7.0.0 — the leading @[Name]@\/@[Exp]@ parameters were dropped and the
+-- result type changed from @Clause@ to @Q ([String], Clause)@.
+mkDispatchClause :: forall a b site c. TyArgs -> MkDispatchSettings b site c -> [ResourceTree a] -> Q ([String], Clause)
+mkDispatchClause tyargs MkDispatchSettings {..} resources = do
```
L0neGamer, line 194: The previous version did not have [Name] or [Exp] as input as far as I can see.

yesod-core/src/Yesod/Routes/TH/Internal.hs
```diff
@@ -0,0 +1,414 @@
...
+-- | The number of type parameters a type constructor declares (0 for a type
+-- that isn't a data\/newtype\/type-synonym, or that we can't reify).
+typeArity :: Name -> Q Int
+typeArity typeName = do
+    info <- reify typeName
+    pure $ case info of
+        TyConI (DataD _ _ vs _ _ _) -> length vs
+        TyConI (NewtypeD _ _ vs _ _ _) -> length vs
+        TyConI (TySynD _ vs _) -> length vs
+        _ -> 0
```
L0neGamer, line 75: Out of interest, why not `fail` here if it's not an appropriate type constructor?

yesod-core/src/Yesod/Routes/TH/Internal.hs
```diff
@@ -0,0 +1,414 @@
...
+--
+-- @since 1.7.0.0
+assertNestedSubArity :: ArityCallSite -> SubsiteName -> SubsiteArity -> RouteCon -> Q ()
+assertNestedSubArity callSite subName subArity rc =
+    case rcResolved rc of
+        Nothing -> pure ()
+        Just tyname -> do
+            nestedArity <- typeArity tyname
+            maybe (pure ()) (fail . arityMismatchMessage callSite) $
+                checkNestedSubArity subName (RouteName (rcName rc)) subArity (RouteArity nestedArity)
```
L0neGamer, line 290: I don't know the full context of the use of this function but I'm itching to say parse don't validate; are we able to move this sort of assertiong to RouteCon generation somehow?

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
```diff
@@ -1,48 +1,346 @@
...
+-- ('isInstance'), which would pin this to 'Q'; it is instead /hoisted out/ and
+-- passed in as @existingInstances@, leaving only fresh-name generation as an
+-- effect. That makes the generator monad-polymorphic over 'Quote': it runs at
+-- 'Q' in production and at a pure 'Quote' instance (a monotonic counter) in
+-- tests.
+--
+-- @since 1.7.0.0
+generateParseRouteClause
+    :: Quote m
+    => Set.Set String
```
L0neGamer, line 140: I think passing in existing instances as a set is fine.
you could pass in a function `ResourceTree a -> m Bool`, and then pass in `nestedInstanceExists` or similar in  non-pure code and some other basic function in pure code.

I can't say for certain how much of an improvement that would be though.

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
```diff
@@ -1,48 +1,346 @@
...
+    liftQ = Trans.lift
+
+    recordName :: MonadState (Set.Set String) n => String -> n ()
+    recordName name =
+        modify (Set.insert name)
+
+    -- The instance-existence decision is hoisted out (into @existingInstances@,
+    -- computed in 'Q' by the caller), so this stays pure.
+    recordNameIfNotInstance name =
+        when (not (name `Set.member` existingInstances)) $
```
L0neGamer: `unless` might be clearer?

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
```diff
@@ -1,48 +1,346 @@
...
+                    -- 'unsafeQToQuote' because brackets are monomorphic 'Q'
+                    -- on template-haskell < 2.17. It is safe here: the only
+                    -- effect a bracket performs is drawing hygiene names for
+                    -- its binders, and this one has none (the lambda binder
+                    -- comes from a splice).
+                    expr <- unsafeQToQuote [e|
+                        fmap
+                            (\ $(varP subName) -> $(pure wrapSub) )
+                            (parseRoute ( $(varE restName), $(varE queryParamsName) ) )
+                        |]
```
L0neGamer, line 312: instead of using the unsafe, would it make sense to just purely build the expression?

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -59,21 +101,116 @@ defaultOpts = MkRouteOpts
...
+-- Then, in a module for the route specifically, we can generate the route
+-- datatype and instances for later hooking in to the main instance.
+--
+-- @
+-- module App.Routes.NestR where
+--
+-- import App.Routes.Resources
+-- import Yesod.Core
+--
+-- mkYesodOpts (setFocusOnNestedRoute "NestR" defaultOpts) "App" appResources
```
L0neGamer, line 139: so this example would generate a `NestR` type with two constructors, NestIndexR and NestShowR?

These wouldn't have the usual yesod instances but would have Nested instances which the `App` module would use when it makes the full application.

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
```diff
@@ -97,7 +236,23 @@ setCreateResources b rdo = rdo { roCreateResources = b }
...
+-- whose @master@ is determined by the @subsite@ (a @subsite -> master@
+-- functional dependency on the user's class), @UndecidableInstances@.
+--
+-- Monomorphic sites always use nested discovery regardless of this flag; see
+-- 'discoveryMode'.
+--
+-- Default: 'False'.
 --
 -- @since 1.6.28.0
 setParameterizedSubroute :: Bool -> RouteOpts -> RouteOpts
```
L0neGamer, line 253: I'm not sure this docstring is useful, or it might be too useful.

It mentions nesting types a lot but this feature functions regardless of nested routes, at least as I understand it.

yesod-core/src/Yesod/Routes/TH.hs
```diff
@@ -1,11 +1,41 @@
 module Yesod.Routes.TH
     ( module Yesod.Routes.TH.Types
       -- * Functions
-    , module Yesod.Routes.TH.RenderRoute
+    , -- ** RenderRoute
```
L0neGamer, line 4: explicit exports are very good!

yesod-core/src/Yesod/Routes/TH.hs
```diff
@@ -1,11 +1,41 @@
...
+    , roFocusOnNestedRoute
+    , roNestedRouteFallthrough
+    , setCreateResources
+    , setParameterizedSubroute
+    , setNestedRouteFallthrough
+    , nullifyWhenNoParam
+    , DiscoveryMode(..)
+    , discoveryMode
     , module Yesod.Routes.TH.ParseRoute
     , module Yesod.Routes.TH.RouteAttrs
```
L0neGamer, line 24: are these worth being explicit about as well?

L0neGamer (commented):
I've given a look at everything but the tests.

Generally very good; I can see what's happening roughly and why it's happening.

There are a fair amount of refactors here that I'd have liked to see split out to reduce on noise, but with >100 commits I presume that would be very hard.

I'll review the tests tomorrow, but I'd really like it if there were some way to run the non-nested-based tests on the pre-this PR yesod state. If we could do that then we would be able to confidently say that there haven't been any breakage.

To combine those two thoughts: there's a lot of movement that isn't neccessarily strictly nested-dispatch related, which means that the new tests aren't trivially easy to run. Moving the non-feature related changes and tests to earlier in the commit history, or somehow separating them out, would increase my confidence in the lack of changes to functionality.

yesod-core/src/Yesod/Routes/TH/RenderRoute.hs
L0neGamer: This is for my understanding of the general situation.

In normal yesod execution, I'm pretty sure we would make a top level NestR type with those constructors, so I guess the entire point here is that you can split up that specific generation.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
parsonsmatt: Oh god. I knew this at some point. `SingleDispatchClause` maybe? `SomeDataCollection`? 😅

parsonsmatt:
I had Claude extract the backwards compatibility tests to https://github.com/yesodweb/yesod/pull/1919 at least on master, which should give us some confidence that things are working the same.

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
parsonsmatt: We could, but direct constructor use is a common pain point with upgrading TH versions. Quotes are a lot safer.

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
parsonsmatt: yeah the negation could go in a few places. `unless` appeals to my ruby heart but I could also see `when (Set.notMember name existingInstances)` being an improvement.

yesod-core/test/Hierarchy/ResourceTree.hs
```diff
@@ -0,0 +1,57 @@
...
+    /get        Get3      GET
+    /post       Post3         POST
+--    /#Int       Delete3            DELETE
+
+/afterwards AfterR !parent !key=value1:
+  /             After     GET !child !key=value2
+
+-- /trailing-nest TrailingNestR:
+--  /foo TrailingFooR GET
+--  /#Int TrailingIntR GET
```
L0neGamer: commented out routes?

yesod-core/test/Hierarchy/Nest.hs
L0neGamer: If so, it would be good to mention that; the moment this PR merges these are the canonical ways to use yesod

yesod-core/test/Route/SubDispatchAritySpec.hs
```diff
@@ -0,0 +1,66 @@
...
+    it "accepts a parameterized subsite whose nested datatype carries the param" $
+        check "MySub" "NestedR" 1 1 `shouldBe` Nothing
+
+    it "rejects when the nested datatype carries more params than the subsite" $
+        -- Over-arity leaves the instance head partially applied (kind
+        -- @Type -> ...@), so this must be rejected, not silently accepted.
+        check "MySub" "NestedR" 1 2 `shouldSatisfy` isJust
+
+    it "rejects a parameterized subsite with an unparameterized nested datatype" $
+        check "MySub" "NestedR" 1 0 `shouldSatisfy` isJust
```
L0neGamer: do we want to assert the error message?

yesod-core/test/Route/FallthroughSpec.hs
```diff
@@ -0,0 +1,99 @@
...
+        let routeShouldParse path result =
+                parseRoute (path, []) `shouldBe` Just result
+            routeShouldNotParse path =
+                parseRoute (path, []) `shouldBe` (Nothing :: Maybe (Route App))
+        it "can fall through" $ do
+            routeShouldParse ["foo", "blah"] (SecondFooR FooBlahR)
+        it "nested fallthrough works too" $ do
+            routeShouldParse ["foo", "baz", "foo"] (SecondFooR (FooBaz2R FooBaz2FooR))
+        it "can fail" $ do
+            routeShouldNotParse ["asdf"]
```
L0neGamer, line 85: just for completeness would be nice to have a case for `/foo`

yesod-core/test/Route/FocusLeafConsSpec.hs
```diff
@@ -0,0 +1,54 @@
...
+            routes
+    pure [DataD [] (mkName "FocusNestR") [] Nothing cons [DerivClause Nothing [ConT ''Show]]])
+
+spec :: Spec
+spec = describe "mkRouteConsOpts (focus mode)" $
+    it "keeps the focused route's leaf constructors" $ do
+        -- These references only typecheck if the constructors exist on the
+        -- generated datatype; the leaf-drop bug omitted them entirely.
+        show FocusIndexR `shouldBe` "FocusIndexR"
+        show (FocusShowR 7) `shouldBe` "FocusShowR 7"
```
L0neGamer: Is there a way we can assert that these are the only constructors of FocusNestR?

yesod-core/test/Route/InstanceProbeTypes.hs
```diff
@@ -0,0 +1,55 @@
...
+-- type environment at a 'reify' (the same reason "Route.DeepArityTypes" is
+-- separate).
+--
+-- 'Probe' stands in for any of the real nested-discovery classes
+-- (@YesodDispatchNested@, @RenderRouteNested@, …) that 'nestedInstanceExists'
+-- probes: those classes aren't visible to the pure @test-routes@ suite, but
+-- 'nestedInstanceExists' only ever asks @isInstance klass [head]@, so any
+-- single-parameter class with the right instance shape exercises the same
+-- code path.
+module Route.InstanceProbeTypes
```
L0neGamer, line 17: I have more suggested test cases:

```
data HasInst1 a = HasInst1
instance Probe HasInst1

data HasInst2 a  b = HasInst2
instance Probe (HasInst2 a)

data HasInstInt a = HasInstInt
instance Probe (HasInstInt Int)
```

yesod-core/test/YesodCoreTest/FallthroughDispatch/Resources.hs
```diff
@@ -0,0 +1,37 @@
...
+-- uses @parseRoutesNoCheck@ (the compile-time overlap check would otherwise
+-- reject the shared prefix that the fallthrough behaviour exists to handle).
+module YesodCoreTest.FallthroughDispatch.Resources where
+
+import Yesod.Core
+import Yesod.Routes.TH.Types
+
+data App = App
+
+-- TODO: make parseRoutesNoCheck smarter
```
L0neGamer: TODO?

yesod-core/test/YesodCoreTest/FallthroughMatrixRuntime.hs
```diff
@@ -0,0 +1,96 @@
...
+specs :: Spec
+specs = describe "cross-module fallthrough flag mismatch (top True, split child False)" $ do
+    it "matches the split child's index (FooIndexR at /foo)" $
+        testRequestIO 200 ["foo"] (Just "FooIndexR")
+    it "matches a child of the split parent (FooNeatR at /foo/neat)" $
+        testRequestIO 200 ["foo", "neat"] (Just "FooNeatR")
+    it "falls through the split parent to the inline sibling (OtherR at /foo/other)" $
+        -- The headline R10 case. The top module wants fallthrough; FirstFooR
+        -- has no "other" child. Post-R10 this falls through to SecondFooR's
+        -- OtherR (200). Pre-R10, FirstFooR's split instance (flag False) bakes
```
L0neGamer: what does R10 mean?

yesod-core/test/YesodCoreTest/ParamDefaultSplitData.hs
```diff
@@ -0,0 +1,28 @@
...
+{-# LANGUAGE ViewPatterns #-}
+
+-- | Route data for the #1 regression: a *parameterized* top-level site with a
+-- nested route, generated with /default/ opts (i.e. NOT
+-- 'setParameterizedSubroute'), so the nested datatype 'SubParentDR' is
+-- generated at kind 'Type'. The matching dispatch splice lives in the
+-- separately compiled "YesodCoreTest.ParamDefaultSplitRuntime", so the child
+-- datatype is in scope there — the precise shape that made the @go@ delegation
+-- probe build the ill-kinded @SubParentDR a@ and abort the splice.
+module YesodCoreTest.ParamDefaultSplitData where
```
L0neGamer: this breaks the general pattern you set of having the Data module in a subfolder, jsyk.

I think I'd really like a refactor of the tests so that the Runtimes are in the same folder as the Data's; it makes more sense where the tests live in that case.

Might even be worth renaming to Spec, but that can come in a future PR.

yesod-core/test/YesodCoreTest/ParamSubsite/SplitNested.hs
```diff
@@ -0,0 +1,58 @@
...
+        master  = mkName "master"
+        cxt    = [ ConT (mkName "PClass") `AppT` VarT subsite `AppT` VarT master ]
+        tyargs = SomeTyArgs ((VarT subsite, subsite) :| [])
+    mkNestedSubDispatchInstance
+        (setParameterizedSubroute True defaultOpts)
+        "PNestedR"
+        cxt
+        tyargs
+        return
+        (map (fmap (parseType . dropBracket)) resourcesPSub))
```
L0neGamer: why the manual building?

yesod-core/test/YesodCoreTest/RenderRouteSpec.hs
```diff
@@ -0,0 +1,51 @@
...
+
+spec :: Spec
+spec = do
+    describe "renderRoute" $ do
+        it "works on top level" $ do
+            case renderRoute (FirstR 1 BlahR) of
+                (["first", "1", "blah"], []) ->
+                    pure @IO ()
+                wrong ->
+                    fail $ "Expecte renderRoute to work, but got: " <> show wrong
```
L0neGamer: nit: spelling

L0neGamer (commented):
The tests seem good.

It might assist readability and navigation if the Runtime and actual tests lived in the same folders as the setup data; as it stands it's not 100% clear when two modules relate to each other that easily.

If you're happy with waiting I'll do one last read through on monday.

yesod-core/test/Hierarchy/ResourceTree.hs
L0neGamer: These might still be needed to exercise a parse bug

yesod-core/src/Yesod/Routes/TH/ParseRoute.hs
```diff
@@ -1,48 +1,346 @@
...
+                pat = TupP [pathPat, VarP queryParamsName]
+            pure [Clause [pat] (NormalB caseExpr) []]
+  where
+    -- A child clause matches the @(path, queryParams)@ tuple with no @where@
+    -- bindings, so it converts directly to a @case@ alternative. (Assert the
+    -- shape we rely on rather than silently dropping non-empty @where@ decls.)
+    clauseToMatch :: Clause -> Match
+    clauseToMatch (Clause [pat] body []) = Match pat body []
+    clauseToMatch c =
+        error $ "buildInlineParseClauses: unexpected child clause shape: " <> show c
```
L0neGamer: would moving this over to using MonadFail make more sense than using error here? Alternative would be don't match on `Clause`, and instead make `buildInlineParseClauses'` which gives the singular pattern and body that this expects.

yesod-core/src/Yesod/Routes/TH/Dispatch.hs
```diff
@@ -60,86 +196,181 @@ mkDispatchClause MkDispatchSettings {..} resources = do
...
+    -- 'envWrap' to 'Just'. This single 'local' — entered once at the top→nested
+    -- boundary and never undone — is the entire top-vs-nested phase distinction.
+    withChildScope :: [Exp] -> Exp -> DispatchM r -> DispatchM r
+    withChildScope dyns constr = local $ \e ->
+        let sdc = envSdc e
+        in e { envSdc = sdc
+                 { extraParams = extraParams sdc ++ dyns
+                 , extraCons = extraCons sdc ++ [constr]
+                 }
+             , envWrap = justE
```
L0neGamer: Like I mentioned with some of the other things, since envWrap can only either be `id` or `justE`, would it make sense to use an enum/Bool to clearly label that this is either `'Just` or it's nothing at all?

L0neGamer (commented):
Some last extra bits.

yesod-core/test/Route/FocusLeafConsSpec.hs
parsonsmatt: Yeah I think a case expr ought to do it

yesod-core/test/Route/InstanceProbeTypes.hs
parsonsmatt: So `Probe` here is `Type -> Constraint`- `instance Probe HasInst1` is ill kinded.  Interestingly, if we do a polykinded `Probe` class, then it only works if the type is saturated, but then it does pick things up.

yesod-core/test/Hierarchy/Nest.hs
parsonsmatt: I modified the tests to use the more standard approach.

parsonsmatt:
OK I think all review comments are addressed. I'd like to merge tomorrow and release; if you have any blocking comments, feel free to leave 'em!

L0neGamer (approved):
The changes seem good, although I haven't look at the latest stuff in as much detail as previous changes. 

Likely good to go when CI is fixed!

PR #1919, opened 2026-06-11
Pin existing routing behavior with regression tests

parsonsmatt:
Extracted from #1887 (split route compilation), per review: these tests assert behavior yesod-core **already has on master**, so they belong in their own PR where CI demonstrates they pass with no library changes. #1887 will be rebased on top of this branch, removing these files from its diff.

What's pinned:

- **ParameterizedSubDispatchRuntime** — end-to-end dispatch through a parameterized subsite (associated types, ConstraintKinds, dynamic parent pieces), including 404s and method mismatches
- **MultiPieceNestedRuntime** — nested parents whose prefix captures dynamics (`/shop/#Int/admin`), forwarding parent and child captures
- **ZeroPieceShadowRuntime** — a zero-piece nested parent (`/`) shadows later siblings: longstanding, previously unpinned behavior
- **BangSeparatorRuntime** — `/!#Int` dispatches like `/#Int`
- **RuntimeHarness** — the shared WAI round-trip assertion helper the above use

No library code is touched; the `tests` suite goes from 200 to 217 examples.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/yesod-core/test/YesodCoreTest.hs
+++ b/yesod-core/test/YesodCoreTest.hs
@@ -11,6 +11,12 @@ import YesodCoreTest.Links
 import YesodCoreTest.Header
 import YesodCoreTest.NoOverloadedStrings
 import YesodCoreTest.SubSub
+import YesodCoreTest.ParameterizedSubData ()
+import YesodCoreTest.ParameterizedSubDispatch ()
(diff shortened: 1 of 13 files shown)


L0neGamer (approved):
I appreciate splitting this out!

Good to go as far as I can see

PR #1927, opened 2026-07-21
Add runFormPRG for Post/Redirect/Get form handling

jezen:
yesod-form has no support for the Post/Redirect/Get pattern: a handler that redirects after a failed POST loses the user's input and the validation errors, so applications either re-render on the POST (breaking refresh and back-button behaviour) or hand-roll session stashing around the library's private form internals.

runFormPRG is a drop-in sibling of runFormPost, called from both the GET and POST handlers of a route. On a POST it behaves exactly like runFormPost, and additionally stashes the submitted parameters in the user session when the result is FormFailure, so the handler can simply redirect back to the GET route. On a GET it pops any stashed parameters and replays them through the form, re-rendering the user's input and the validation errors; with nothing stashed it is equivalent to generateFormPost.

The replay renders a fresh CSRF token and skips the token check, since it is a re-render rather than a submission. The stash is read through the per-request cache so that several calls in one handler (multiple identifyForm-wrapped forms on a page) all see it, and a FormMissing result on a POST leaves the stash alone so that the non-submitted forms on a page cannot resurrect it after a successful submission has cleared it. Stashes are keyed by request path and serialised with aeson, which is already a dependency. Uploaded files are not stashed, and an undecodable stash is treated as absent.

A new yesod-test integration suite exercises the full cycle: replay after the redirect, one-shot consumption of the stash, clearing on success, CSRF enforcement, two identified forms on one page, resilience to an undecodable stash, and per-path keying. yesod-test's library depends only on yesod-core, so this introduces no dependency cycle. The test application generates a random client-session key at startup, so no key file is written to disk.

### DIFF
--- a/yesod-form/ChangeLog.md
+++ b/yesod-form/ChangeLog.md
@@ -1,5 +1,12 @@
 # ChangeLog for yesod-form
 
+## 1.7.10
+
+* Add `runFormPRG`, a `runFormPost` variant for the
+  Post/Redirect/Get pattern: failed submissions are stashed in the session
+  and replayed (input and validation errors) by the next GET.
+  [#1927](https://github.com/yesodweb/yesod/pull/1927)
+
 ## 1.7.9.3
 
 * Support `yesod-core` 1.7
--- a/yesod-form/Yesod/Form/Functions.hs
+++ b/yesod-form/Yesod/Form/Functions.hs
@@ -32,6 +32,8 @@ module Yesod.Form.Functions
     , runFormPost
     , runFormPostNoToken
     , runFormGet
+      -- ** Post\/Redirect\/Get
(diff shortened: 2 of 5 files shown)


jezen:
@parsonsmatt @MaxGabriel Tagging you both for review since it says (at least for you Matt) that you worked in this area recently, although IIRC you don't rely on yesod-form much 🤔 

In any case, I think this is the right thing and it's been missing from Yesod for too long so I'll merge it, so consider this just for posterity.

jezen:
@snoyberg Would you mind uploading this to Hackage? Alternatively, could you grant me maintainership there?

snoyberg:
@jezen I've added you as an uploader for yesod-form

PR #1929, opened 2026-07-28
Break yesod-form/yesod-test test dependency cycle

### LINKED ISSUE #1928: Stackage: yesod-test dependency cycle detected: yesod-test, yesod-form, yesod-test, sydtest-yesod
Just wanted to report that we hit a circular dependency for in Stackage Nightly after the recent yesod-form-1.7.10, @jezen

Well we can probably workaround it by disabling some testsuite I expect, but reporting this since it is quite unusual.

jezen:
Fixes #1928.

yesod-form-1.7.10 made its test suite depend on yesod-test for the `runFormPRG` integration tests. Cabal accepts this because its solver works per-component, but Stackage checks for cycles at the package level with test suites enabled, and yesod-test's own test suite already depended on yesod-form — hence the cycle reported by the Stackage curator.

This PR:

* Moves the `runFormPRG` integration tests to yesod-test's test suite, which already depended on yesod-form, so no new dependency edge is introduced. A module comment explains why they live there.
* Restores yesod-form's test suite to its pre-1.7.10 dependencies.
* Bumps yesod-form to 1.7.10.1. The fix needs a new Hackage upload, since a metadata revision cannot remove a dependency.
* Bumps yesod-test to 1.7.0.3, which merely carries the relocated tests along and is not urgent to release.

### DIFF
--- a/yesod-form/ChangeLog.md
+++ b/yesod-form/ChangeLog.md
@@ -1,5 +1,13 @@
 # ChangeLog for yesod-form
 
+## 1.7.10.1
+
+* Move the `runFormPRG` integration tests to yesod-test's test suite,
+  removing yesod-form's test-suite dependency on yesod-test. That
+  dependency created a package-level cycle
+  (yesod-test → yesod-form → yesod-test) which broke the Stackage
+  build plan. [#1928](https://github.com/yesodweb/yesod/issues/1928)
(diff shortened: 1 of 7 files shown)


parsonsmatt (approved):
nice!

